PR #7, opened 2020-02-06
CLOUDP-56396: Create list of OSS licenses used by mcli & include it in packaging

gssbzn:
The generated zip/tar file will include something like this

Screenshot 2020-02-06 at 14 36 01


Makefile
```diff
@@ -58,14 +59,22 @@ gen-mocks: ## Generate mocks
...
+.PHONY: gen-notices
+gen-notices: ## Generate 3rd party notices
+	@echo "==> Generating 3rd party notices"
+	@chmod -R 777 ./third_party_notices
+	@rm -Rf third_party_notices
+	go-licenses save "github.com/mongodb/mcli" --save_path=third_party_notices
+
 .PHONY: release
-release:
+release: gen-notices ## Use goreleaser to generate builds and publish
```
colm-quinn: That is a funky little go feature right there! 👏

colm-quinn: [q] Is there anyway to configure it to warn us if there is a potential commercial license that we're about to use that could cause us problems? Would be really cool if we got warning when we tried to pull them in, rather than when we audit the tool with legal later.

gssbzn: Yes, if that's your kind of thing, https://github.com/google/go-licenses#checking-for-forbidden-licenses

colm-quinn: It's a @jamesbroadhead call, but my gut feels like it's something we should do if we can.

gssbzn: @colm-quinn I created CLOUDP-56971 to follow on this just to keep this PR focused

PR #9, opened 2020-02-06
CLOUDP-56969: Fix whitelist not being optional for owner creation

gssbzn:


go.mod
```diff
@@ -11,7 +11,7 @@ require (
 	github.com/mattn/go-colorable v0.1.4 // indirect
 	github.com/mattn/go-isatty v0.0.12 // indirect
 	github.com/mitchellh/go-homedir v1.1.0
-	github.com/mongodb-labs/pcgc v0.0.4-0.20200130102748-3d11fb006354
+	github.com/mongodb-labs/pcgc v0.1.0
```
colm-quinn: 🎉

internal/store/owners.go
```diff
@@ -9,14 +9,18 @@ import (
...
 
 // CreateOwner encapsulate the logic to manage different cloud providers
-func (s *Store) CreateOwner(u *cm.User, IPs string) (*cm.CreateUserResponse, error) {
+func (s *Store) CreateOwner(u *cm.User, IPs []string) (*cm.CreateUserResponse, error) {
 	switch s.service {
 	case config.OpsManagerService:
-		opts := &cm.WhitelistOpts{Whitelist: IPs}
+		var opts *cm.WhitelistOpts
+		if len(IPs) > 0 {
+			opts = &cm.WhitelistOpts{Whitelist: IPs}
```
colm-quinn: [q] Do we have a test that verifies the `CreateFirstUser` functionality? The calls are pretty simple right now, but as we add more parameters it would be good to see any validation that's necessary on the inputs it done correctly.

gssbzn: We do have a test, adding a scenario with and without the param

internal/cli/ops_manager_owner_create_test.go
```diff
@@ -29,25 +29,49 @@ func TestManagerOwnerCreate_Run(t *testing.T) {
...
+		mockStore.
+			EXPECT().
+			CreateOwner(createOpts.newOwner(), createOpts.whitelistIps).Return(expected, nil).
+			Times(1)
+
+		err := createOpts.Run()
+		if err != nil {
+			t.Fatalf("Run() unexpected error: %v", err)
+		}
+	})
```
colm-quinn: Excellent. Thanks for this!

PR #10, opened 2020-02-10
CLOUDP-57093: Hombrew tap

gssbzn:


README.md
```diff
@@ -14,6 +15,11 @@ This works well for shared hosts and other systems where you don't have a privil
 
 Ideally, you should install it somewhere in your PATH for easy use. `/usr/local/bin` is the most probable location.
 
+## Hombrew
+```bash
+brew install puma/puma/puma-dev
```
colm-quinn: [q] Is this just a prereq for running mcli? The PR reads like the instructions you'd add here would be how to install mcli from brew and a prereq would in the section below.

I might be missing something here...

gssbzn: this is how you install mcli, you can (a) get the binary and run it from there, (b) use homebrew, (c) build it from source, that's where you get `Prerequisite`

PR #11, opened 2020-02-10
CLOUDP-57047: set up an E2E test with cloud manager

gssbzn:


.drone.yml
```diff
@@ -37,13 +37,26 @@ steps:
...
     image: golang:1.13
     commands:
-      - make build
+      - make e2e-test
     volumes:
       - name: deps
         path: /go
+    environment:
+      MCLI_PUBLIC_API_KEY:
+        from_secret: mcli_public_api_key
```
colm-quinn: [q] Did this end up targeting cloud dev/QA in the end?

gssbzn: had a chat with @jamesbroadhead today we are now doing against dev

.drone.yml
```diff
@@ -37,13 +37,26 @@ steps:
...
+      MCLI_PRIVATE_API_KEY:
+        from_secret: mcli_private_api_key
+      MCLI_OPS_MANAGER_URL:
+        from_secret: mcli_ops_manager_url
```
colm-quinn: [q] How does it setup/get the ops manager instance?

e2e/atlas_clusters_test.go
```diff
@@ -0,0 +1,110 @@
...
+		t.Fatalf("unexpected error: %v", err)
+	}
+
+	r := rand.New(rand.NewSource(time.Now().UnixNano()))
+
+	atlasEntity := "atlas"
+	clustersEntity := "clusters"
+	clusterName := fmt.Sprintf("e2e-cluster-%v", r.Uint32())
+
+	t.Run("Create", func(t *testing.T) {
```
colm-quinn: [q] Is it possible for us to define command keywords centrally somewhere, so they can be reused here and the cli definitions?

.drone.yml
gssbzn: talked offline

PR #13, opened 2020-02-12
CLOUDP-57241: Delete Project whitelist IP

gssbzn:


e2e/atlas_whitelist_test.go
```diff
@@ -0,0 +1,83 @@
...
+	}
+	_, err = os.Stat(cliPath)
+
+	if err != nil {
+		t.Fatalf("unexpected error: %v", err)
+	}
+
+	atlasEntity := "atlas"
+	whitelistEntity := "whitelist"
+	entry := "77.107.234.162"
```
colm-quinn: [q] Why this IP and not just a testing one? `whois` shows this as some IP in Ireland - not sure if we should have that in the E2E tests if we don't need it.

internal/cli/atlas_whitelist_delete.go
```diff
@@ -0,0 +1,97 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package cli
+
+import (
+	"fmt"
+
+	"github.com/AlecAivazis/survey/v2"
```
colm-quinn: [q] This is coming up as 404 not found. Do we need it? I wonder how's it's working still.

internal/cli/atlas_whitelist_delete.go
```diff
@@ -0,0 +1,97 @@
...
+	return survey.AskOne(prompt, &opts.confirm)
+}
+
+// mcli atlas whitelist delete <entry> --force
+func AtlasWhitelistDeleteBuilder() *cobra.Command {
+	opts := &atlasWhitelistDeleteOpts{
+		globalOpts: newGlobalOpts(),
+	}
+	cmd := &cobra.Command{
+		Use:     "delete [entry]",
```
colm-quinn: [q] Does the --force show up in the help text if we don't put it here?

internal/prompts/prompts.go
```diff
@@ -0,0 +1,15 @@
+package prompts
+
+import (
+	"fmt"
+
+	"github.com/AlecAivazis/survey/v2"
+)
+
+// NewDeleteConfirmation creates a prompt to confirm if the entry should be deleted
+func NewDeleteConfirmation(entry string) *survey.Confirm {
```
colm-quinn: [q] Does this still come up if we do a --force? Should it - I would say no, but don't feel strongly on this.

internal/cli/atlas_whitelist_delete.go
gssbzn: This is go mod versioning system, this is the repo, https://github.com/AlecAivazis/survey, and the v2 just makes go mods to look for a v2.x.x branch, in theory all repos built for go mods should now version like this, some are already doing `v5` and such

internal/cli/atlas_whitelist_delete.go
gssbzn: yes cause that's a flag and all flag related help/autocomplete info comes from the `cmd.Flags()` property, if you don't want it to show then you need to use `MarkHidden`

internal/prompts/prompts.go
gssbzn: It doesn't show, if you look at https://github.com/mongodb/mcli/blob/c5c1fbce3240a80401f688e173d9e8c348fd8b74/internal/cli/atlas_whitelist_delete.go#L63-L65 you'll see that we return before this gets called if the person used `--force`
This method is meant to encapsulate the creation of the survey struct, I plan to reuse it as we keep making the same struct for all our "confirm" operation only changing the thing between quotes but I didn't make those changes on this PR

e2e/atlas_whitelist_test.go
gssbzn: Atlas has validations, since this is for realsy I didn't want it to fail because of the validations

gssbzn: changed to 192.168.0.1, let's see how that goes

internal/prompts/prompts.go
gssbzn: This is also a go thing where you create a `NewSomething` method for encapsulating ` &struct{}` operations, also since go doesn't have polymorphism they recommend making names quite explicit, hence the  full "new Delete Confirmation" tho I could make it just `NewDeleteConfirm` to keep it in line with `survey`

internal/cli/atlas_whitelist_delete.go
colm-quinn: Got it, thanks for the explanation!

e2e/atlas_whitelist_test.go
colm-quinn: Cool. I'd probably have tried google (8.8.8.8) or something in that case - ip's should go away but I have no idea what would happen to the test if it did.

internal/prompts/prompts.go
colm-quinn: Excellent, thanks for the details!

PR #15, opened 2020-02-13
CLOUDP-57303: Improve delete commands structure

gssbzn:
This tries to extract some of the repeating blocks so we can reuse them later

### DIFF
--- a/internal/cli/atlas_clusters_delete.go
+++ b/internal/cli/atlas_clusters_delete.go
@@ -15,9 +15,6 @@
 package cli
 
 import (
-	"fmt"
-
-	"github.com/AlecAivazis/survey/v2"
 	"github.com/mongodb/mcli/internal/flags"
 	"github.com/mongodb/mcli/internal/store"
 	"github.com/mongodb/mcli/internal/usage"
@@ -26,9 +23,8 @@ import (
 
 type atlasClustersDeleteOpts struct {
 	*globalOpts
-	name    string
-	confirm bool
-	store   store.ClusterDeleter
+	*deleteOpts
+	store store.ClusterDeleter
 }
 
 func (opts *atlasClustersDeleteOpts) init() error {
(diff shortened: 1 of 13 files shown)


internal/cli/cloud_manager_clusters_shutdown.go
```diff
@@ -40,17 +40,15 @@ func (opts *cmClustersShutdownOpts) init() error {
...
-
-	opts.store = s
-	return nil
+	var err error
+	opts.store, err = store.New()
+	return err
 }
 
 func (opts *cmClustersShutdownOpts) Run() error {
+	if !opts.confirm {
```
gssbzn: fixing a bug that would ignore the answer to confirm

internal/cli/cloud_manager_clusters_startup.go
```diff
@@ -40,17 +40,15 @@ func (opts *cmClustersStartupOpts) init() error {
...
-
-	opts.store = s
-	return nil
+	var err error
+	opts.store, err = store.New()
+	return err
 }
 
 func (opts *cmClustersStartupOpts) Run() error {
+	if !opts.confirm {
```
gssbzn: fixing a bug that would ignore the answer to confirm

internal/cli/cli.go
```diff
@@ -47,3 +51,65 @@ func (opts *globalOpts) OrgID() string {
 	opts.orgID = config.OrgID()
 	return opts.orgID
 }
+
+// deleteOpts options required when deleting a resource.
+// A command can embed this structure and then safely rely on the methods Confirm, DeleteFromProject or Delete
+// to manage the interactions with the user
+type deleteOpts struct {
```
gssbzn: These are the main change and the rest is just deleting the duplicates to use this struct
We should be able now to embed this and reuse the Confirm, DeleteFromProject or Delete methods

internal/cli/atlas_clusters_delete.go
```diff
@@ -26,53 +23,32 @@ import (
 
 type atlasClustersDeleteOpts struct {
 	*globalOpts
-	name    string
-	confirm bool
-	store   store.ClusterDeleter
+	*deleteOpts
+	store store.ClusterDeleter
```
colm-quinn: 👍

PR #17, opened 2020-02-14
CLOUDP-57043: CLI: Listing Atlas database users

andreaangiolillo:


internal/cli/atlas_dbusers_list.go
```diff
@@ -0,0 +1,88 @@
...
+	}
+
+	s, err := store.New()
+
+	if err != nil {
+		return err
+	}
+
+	opts.store = s
+	return nil
```
gssbzn: [non blocker] This could be simplified to 
```go
var err error
opts.store, err = store.New()
return err
```

internal/store/database_users.go
```diff
@@ -55,3 +60,13 @@ func (s *Store) DeleteDatabaseUser(groupID, username string) error {
 		return fmt.Errorf("unsupported service: %s", s.service)
 	}
 }
+
+func (s *Store) ProjectDatabaseUser(projectID string, opts *atlas.ListOptions) ([]atlas.DatabaseUser, error) {
```
gssbzn: [nit] no need for `Project` here on the name, similar to a previous comment for listing methods let's try to make them plural

internal/store/database_users.go
```diff
@@ -55,3 +60,13 @@ func (s *Store) DeleteDatabaseUser(groupID, username string) error {
 		return fmt.Errorf("unsupported service: %s", s.service)
 	}
 }
+
+func (s *Store) DatabaseUser(projectID string, opts *atlas.ListOptions) ([]atlas.DatabaseUser, error) {
```
gssbzn: maybe it was missed between other changes but let's add the `s` at the end

PR #20, opened 2020-02-17
CLI: Describe Atlas ProjectIP Whitelist

andreaangiolillo:


internal/fixtures/project_ip_whitelists.go
```diff
@@ -19,12 +19,14 @@ import (
...
-		{
-			Comment:   "test",
-			GroupID:   "5def8d5dce4bd936ac99ae9c",
-			CIDRBlock: "37.228.254.100/32",
-			IPAddress: "37.228.254.100",
-		},
+	return []atlas.ProjectIPWhitelist{*IPWhiteList()}
+}
+
+func IPWhiteList() *atlas.ProjectIPWhitelist {
```
colm-quinn: [q] Why can't you use the original `ProjectIPWhitelist []` rather than making the additional helper here?

andreaangiolillo: The reason is that the command mcli `atlas whitelist(s) describe` returns an atlas.ProjectIPWhitelist object rather than an array. This practise seems used also in other helper (see internal/fixtures/clusters.go

PR #21, opened 2020-02-18
CLOUDP-57320: CLI: Listing Atlas Alert Configurations

andreaangiolillo:


internal/cli/atlas_alertconfig.go
```diff
@@ -0,0 +1,32 @@
...
+package cli
+
+import (
+	"github.com/spf13/cobra"
+)
+
+func AtlasAlertConfigBuilder() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:        "alertConfigs",
+		Aliases:    []string{"alerts"},
```
gssbzn: we can't use just `alerts` as we'll have `alerts` for other things
let's make the official command `alert-config` as the spec, feel free to add an alias for the camelCase version

internal/cli/atlas_alertconfig.go
```diff
@@ -0,0 +1,32 @@
...
+		SuggestFor: []string{"replicasets"},
```
gssbzn: [q] copy/paste?

internal/cli/atlas_alertconfig.go
```diff
@@ -0,0 +1,32 @@
...
+		Short:      "Manage Atlas aler configurations for your project.",
```
gssbzn: typo, Atlas alert,

internal/store/alert_configuration.go
```diff
@@ -0,0 +1,42 @@
...
+type AlertConfigurationLister interface {
+	ProjectAlertConfiguration(string, *atlas.ListOptions) ([]atlas.AlertConfiguration, error)
+}
+
+type AlertConfigurationStore interface {
+	AlertConfigurationLister
+}
+
+// ProjectAlertConfiguration encapsulate the logic to manage different cloud providers
+func (s *Store) ProjectAlertConfiguration(projectID string, opts *atlas.ListOptions) ([]atlas.AlertConfiguration, error) {
```
gssbzn: [nit] these configs are always per project, I don't see the need for adding `Project` at the beginning
Also since this is an index and returns multiple values can we make the name plural?

PR #23, opened 2020-02-18
CLOUDP-57322: Connect to atlas cluster

gssbzn:
Sadly there's no easy way for us to test this right now :(


andreaangiolillo:
[q] is there a reason why we are not implementing the unit test and the e2e?

gssbzn:
@andreaangiolillo I don't have an easy way to mock the call to the mongoshell so I'll always hit an error
Don't mind this PR anyway we may not need to do this at all

PR #24, opened 2020-02-18
CLOUDP-57045: CLI: Updating Atlas Database Users

andreaangiolillo:


internal/cli/atlas_dbusers_update.go
```diff
@@ -0,0 +1,110 @@
...
+	if opts.ProjectID() == "" {
+		return errMissingProjectID
+	}
+	var err error
+	opts.store, err = store.New()
+	return err
+}
+
+func (opts *atlasDBUsersUpdateOpts) Run() error {
+	current, err := opts.store.DatabaseUser(opts.orgID, opts.username)
```
gssbzn: @andreaangiolillo this is your error, this should be `opts.ProjectID()`

gssbzn: Also, de we really need to fetch the record first? is there any value to it?

andreaangiolillo: I don’t understand your comment on the pull request. Does the command update an existing record right? so I need to get that record in order to append the new roles to it. what am I missing?

andreaangiolillo: How do we know if the username used in the command is associated to a database user if we don't fetch the record first?

internal/convert/database_user.go
```diff
@@ -0,0 +1,30 @@
...
+
+import (
+	"strings"
+
+	atlas "github.com/mongodb/go-client-mongodb-atlas/mongodbatlas"
+)
+
+const (
+	AdminDB = "admin"
+	RoleSep = "@"
```
gssbzn: could we make this private?

internal/convert/database_user.go
```diff
@@ -0,0 +1,30 @@
...
+)
+
+func BuildRoles(r []string) []atlas.Role {
```
gssbzn: please add docs (`// BuildRoles ...`) to what this does

e2e/atlas_dbusers_test.go
```diff
@@ -81,6 +81,33 @@ func TestAtlasDBUsers(t *testing.T) {
...
+			t.Fatalf("unexpected error: %v, resp: %v", err, string(resp))
+		}
+
+		user := mongodbatlas.DatabaseUser{}
+		err = json.Unmarshal(resp, &user)
+		if err != nil {
+			t.Fatalf("unexpected error: %v", err)
+		}
+
+		if user.Username != username {
```
gssbzn: Thanks for adding an e2e test, could we add a check to see we actually changed the roles? I guess that's the only thing that would change from this API

internal/cli/atlas_dbusers_create.go
```diff
@@ -16,7 +16,8 @@ package cli
 
 import (
 	"errors"
-	"strings"
+
+	"github.com/mongodb/mcli/internal/convert"
```
gssbzn: [nit] sometimes importfmt misses the cases where there's an extra empty line, please delete the line and re run `make fmt`

internal/cli/atlas_dbusers_update.go
```diff
@@ -0,0 +1,95 @@
...
+	"github.com/mongodb/mcli/internal/usage"
+	"github.com/spf13/cobra"
+)
+
+type atlasDBUsersUpdateOpts struct {
+	*globalOpts
+	username string
+	password string
+	roles    []string
+	store    store.DatabaseUserStore
```
gssbzn: could we reduce this surface to just be an `Updater`?

internal/cli/atlas_dbusers_update_test.go
```diff
@@ -0,0 +1,57 @@
...
+// limitations under the License.
+
+package cli
+
+import (
+	"testing"
+
+	atlas "github.com/mongodb/go-client-mongodb-atlas/mongodbatlas"
+	"github.com/mongodb/mcli/internal/fixtures"
+
```
gssbzn: [nit] please delete the extra line

PR #25, opened 2020-02-19
CLOUDP-54304: CLI: Atlas cluster creation via config file

gssbzn:


internal/cli/atlas_clusters_create.go
```diff
@@ -56,31 +62,40 @@ func (opts *atlasClustersCreateOpts) init() error {
...
-		ReplicationSpecs:    []atlas.ReplicationSpec{replicationSpec},
+func (opts *atlasClustersCreateOpts) newCluster() (*atlas.Cluster, error) {
+	cluster := new(atlas.Cluster)
+	if opts.filename != "" {
+		if err := file.Load(opts.fs, opts.filename, cluster); err != nil {
+			return nil, err
+		}
+	} else {
+		cluster.ClusterType = replicaSet
+		opts.applyOpts(cluster)
```
colm-quinn: 👍

internal/cli/atlas_clusters_create.go
```diff
@@ -56,31 +62,40 @@ func (opts *atlasClustersCreateOpts) init() error {
...
 	}
-	return cluster
+	cluster.GroupID = opts.ProjectID()
+	return cluster, nil
+}
+
+func (opts *atlasClustersCreateOpts) applyOpts(out *atlas.Cluster) {
+	replicationSpec := opts.newReplicationSpec()
```
colm-quinn: [q] Have options like `ClusterType` now moved into this option? I'm not seeing a 1:1 mapping with settings from the previous version of this file.

internal/cli/atlas_clusters_create_test.go
```diff
@@ -28,27 +29,83 @@ func TestAtlasClustersCreate_Run(t *testing.T) {
...
+			t.Fatalf("Run() unexpected error: %v", err)
+		}
+	})
+
+	t.Run("file run", func(t *testing.T) {
+		appFS := afero.NewMemMapFs()
+		// create test file
+		fileYML := `
+{
+  "name": "ProjectBar",
```
colm-quinn: [q] Wondering should we put this is a samples or tests folder and load them in a loop - would make it easier to extend in the future.

internal/file/file_test.go
```diff
@@ -0,0 +1,57 @@
+package file
```
colm-quinn: Nice job here!

internal/cli/atlas_clusters_create_test.go
gssbzn: We are using `afero`, afero has a feature to read/write from memory (`afero.NewMemMapFs()`), it does not write to disk and this is ideal for unit tests as they should be faster and guarantee no disk pollution
I did discussed with @andreaangiolillo maybe adding it to the e2e tests... maybe

colm-quinn: Excellent. Sounds good.

internal/file/file_test.go
gssbzn: 95% test coverage 😬

internal/cli/atlas_clusters_create.go
gssbzn: ClusterType is now on line 84, I kind of split this in a way that if tomorrow we want to mix file and flags we just need to take this out of the `else`
This is me thinking if someone may have a file but no mdb version defined and then just do something like `--file file --mdb 4.2` right now is one or the other in sake of keeping it simple but I do see the use case for someone going the other way so this is a small future proof of it

colm-quinn: 👍

gssbzn:
Going to merge this, I created a separate ticket to add some e2e tests

PR #28, opened 2020-02-20
CLOUDP-57563: Implementing the e2e test for the command mcli atlas cluster update

gssbzn:
## Proposed changes
_Jira ticket:_ [CLOUDP-57563](https://jira.mongodb.org/browse/CLOUDP-57563)

## Checklist

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -41,22 +41,25 @@ steps:
     image: golang:1.13
     commands:
       - make e2e-test
+    environment:
+      MCLI_OPS_MANAGER_URL:
+        from_secret: mcli_ops_manager_url
+      MCLI_ORG_ID:
+        from_secret: mcli_org_id
+      MCLI_PRIVATE_API_KEY:
+        from_secret: mcli_private_api_key
+      MCLI_PROJECT_ID:
+        from_secret: mcli_project_id
+      MCLI_PUBLIC_API_KEY:
+        from_secret: mcli_public_api_key
+      MCLI_SERVICE:
+        from_secret: mcli_service
     volumes:
       - name: deps
         path: /go
-    environment:
-      MCLI_PUBLIC_API_KEY:
-        from_secret: mcli_public_api_key
-      MCLI_PRIVATE_API_KEY:
-        from_secret: mcli_private_api_key
-      MCLI_OPS_MANAGER_URL:
-        from_secret: mcli_ops_manager_url
-      MCLI_SERVICE:
-        from_secret: mcli_service
-      MCLI_ORG_ID:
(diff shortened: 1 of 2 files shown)


.drone.yml
```diff
@@ -41,22 +41,25 @@ steps:
...
       MCLI_PROJECT_ID:
         from_secret: mcli_project_id
+      MCLI_PUBLIC_API_KEY:
+        from_secret: mcli_public_api_key
+      MCLI_SERVICE:
+        from_secret: mcli_service
+    volumes:
+      - name: deps
+        path: /go
+    depends_on:
```
gssbzn: this is the main change here, this enables to run test and lint on parallel, the rest is just drone fmt doing its thing

e2e/atlas_clusters_test.go
```diff
@@ -49,27 +49,35 @@ func TestAtlasClusters(t *testing.T) {
 		cmd := exec.Command(cliPath,
 			atlasEntity,
 			clustersEntity,
-			"create", clusterName,
+			"create",
+			clusterName,
 			"--region=US_EAST_1",
 			"--members=3",
-			"--instanceSize=M5",
+			"--instanceSize=M10",
```
gssbzn: we can't update M2 or M5

colm-quinn: [q] Were these tests passing before? - Wondering why we didn't catch this until now.

gssbzn: yes because we weren't trying to update the cluster, that's new, when I try to update the cluster I created it would fail if is anything bellow an m10

e2e/atlas_clusters_test.go
```diff
@@ -49,27 +49,35 @@ func TestAtlasClusters(t *testing.T) {
...
 
 		if cluster.Name != clusterName {
-			t.Errorf("got=%#v\nwant=%#v\n", cluster.Name, clusterName)
+			t.Errorf("Name, got=%#v\nwant=%#v\n", cluster.Name, clusterName)
+		}
+		if cluster.MongoDBMajorVersion != "4.0" {
+			t.Errorf("MongoDBMajorVersion, got=%#v\nwant=%#v\n", cluster.MongoDBMajorVersion, "4.0")
+		}
+		if *cluster.DiskSizeGB != 10 {
+			t.Errorf("MongoDBMajorVersion, got=%#v\nwant=%#v\n", cluster.DiskSizeGB, 10)
```
colm-quinn: [np] define this in a reused const of some form - it's used here and above on the Run command, so would be better to reuse a single param.

PR #36, opened 2020-03-05
CLOUDP-58568: Deleting alert configs is not asking for confirmation

gssbzn:
## Proposed changes

_Jira ticket:_ [CLOUDP-58568](https://jira.mongodb.org/browse/CLOUDP-58568)

## Checklist

## Further comments

We need to start testing stdio

### DIFF
--- a/internal/cli/atlas_alert_config_delete.go
+++ b/internal/cli/atlas_alert_config_delete.go
@@ -46,18 +46,21 @@ func AtlasAlertConfigDeleteBuilder() *cobra.Command {
 	opts := &atlasAlertConfigDeleteOpts{
 		globalOpts: newGlobalOpts(),
 		deleteOpts: &deleteOpts{
-			successMessage: "Alert Config '%s' deleted\n",
-			failMessage:    "Alert Config not deleted",
+			successMessage: "Alert config '%s' deleted\n",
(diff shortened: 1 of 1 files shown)


internal/cli/atlas_alert_config_delete.go
```diff
@@ -46,18 +46,21 @@ func AtlasAlertConfigDeleteBuilder() *cobra.Command {
...
+		Short:   "Delete an alert config.",
 		Aliases: []string{"rm", "Delete", "Remove"},
 		Args:    cobra.ExactArgs(1),
 		PreRunE: func(cmd *cobra.Command, args []string) error {
+			if err := opts.init(); err != nil {
+				return err
+			}
 			opts.entry = args[0]
-			return opts.init()
+			return opts.Confirm()
```
andreaangiolillo: [qa] Should not we also rename the init function to Confirm? What am I missing?

andreaangiolillo: Never mind, it is a global function 🙄

gssbzn: till there, different line, `Confirm` is not global comes from `deleteOpts` save us from defining the same function for all delete commands

PR #45, opened 2020-03-12
CLOUDP-56480: Clusters list command should return all clusters

andreaangiolillo:
_Jira ticket:_ [CLOUDP-56480](https://jira.mongodb.org/browse/CLOUDP-56480)


internal/cli/atlas_clusters_list.go
```diff
@@ -27,22 +27,25 @@ type atlasClustersListOpts struct {
 	*globalOpts
 	pageNum      int
 	itemsPerPage int
-	store        store.ClusterLister
+	storeAT      store.ClusterLister
```
gssbzn: there's no need to create a new store, just create a new interface that defines the two methods you need

internal/cli/atlas_clusters_list.go
```diff
@@ -27,22 +27,25 @@ type atlasClustersListOpts struct {
...
+	storeOM      store.ListAllClusters
 }
 
 func (opts *atlasClustersListOpts) init() error {
+	var err error
+
 	if opts.ProjectID() == "" {
-		return errMissingProjectID
+		opts.storeOM, err = store.New()
```
gssbzn: store.New() implements all the methods of available this if is big code smell that you probably just need a new contract `atlasClustersListOpts` and not multiple fields, see my previous comment

internal/cli/atlas_clusters_list.go
```diff
@@ -27,22 +27,25 @@ type atlasClustersListOpts struct {
...
+		opts.storeAT, err = store.New()
 	}
 
-	var err error
-	opts.store, err = store.New()
 	return err
 }
 
-func (opts *atlasClustersListOpts) Run() error {
+func (opts *atlasClustersListOpts) RunAT() error {
```
gssbzn: this all comes from the same thing, multiple properties instead of a more flexible contract, there's no need to have multiple `Run` methods

internal/cli/atlas_clusters_list_test.go
```diff
@@ -41,8 +41,35 @@ func TestAtlasClustersList_Run(t *testing.T) {
...
 		Times(1)
 
-	err := listOpts.Run()
+	err := listOpts.RunAT()
 	if err != nil {
 		t.Fatalf("Run() unexpected error: %v", err)
 	}
 }
+
+func TestAtlasAllClustersProjectsList_Run(t *testing.T) {
```
gssbzn: you don't need a new test function, you should just define a new scenario
this block should be able to live in the already existing function
```go
t.Run("List all clusters projects", func(t *testing.T) {...}
```

internal/cli/cloud_manager_clusters_list.go
```diff
@@ -23,23 +23,26 @@ import (
 	"github.com/spf13/cobra"
 )
 
-type cmClustersListOpts struct {
+type clustersListOpts struct {
```
andreaangiolillo: I decided to change the name since this is also used by OM

internal/cli/cloud_manager_clusters_list_test.go
```diff
@@ -41,8 +41,35 @@ func TestCloudManagerClustersList_Run(t *testing.T) {
...
 		Times(1)
 
-	err := listOpts.Run()
+	err := listOpts.RunCM()
 	if err != nil {
 		t.Fatalf("Run() unexpected error: %v", err)
 	}
 }
+
+func TestCloudManagerAllClustersProjectsList_Run(t *testing.T) {
```
andreaangiolillo: Should we change the name since this is also used by only OM?

internal/store/all_clusters_projects.go
```diff
@@ -0,0 +1,24 @@
...
+
+import (
+	"context"
+	"fmt"
+
+	om "github.com/mongodb/go-client-mongodb-ops-manager/opsmngr"
+	"github.com/mongodb/mongocli/internal/config"
+)
+
+type ListAllClusters interface {
```
gssbzn: it's a go convention to name single method interfaces as `er` adjectives, this should be `AllClustersLister`

internal/store/all_clusters_projects.go
```diff
@@ -0,0 +1,24 @@
...
+	ListAllClustersProjects() (*om.AllClustersProjects, error)
+}
+
+// ListAllClustersProjects encapsulate the logic to manage different cloud providers
+func (s *Store) ListAllClustersProjects() (*om.AllClustersProjects, error) {
+	switch s.service {
+	case config.OpsManagerService:
```
gssbzn: you should support `CloudManagerService` as well as is the same API

internal/cli/cloud_manager_clusters_list.go
gssbzn: if you want to rename this do it to `cloudManagerClustersListOpts` I'm not that worried about it being shared as they are both the same, at some point `ops-manager` was an alias for `cloud-manager` but we were asked to give it its own space
We may eventually rename all to `opsManager` and files along or like we discussed find a new way to organize these commands (folders/packages?) so feel free of start thinking on those

internal/cli/cloud_manager_clusters_list.go
```diff
@@ -23,36 +25,41 @@ import (
...
+func (opts *cloudManagerClustersListOpts) Run() error {
+
+	var result interface{}
+	var err error
+
+	if opts.projectID == "" && config.Service() == config.OpsManagerService {
+		result, err = opts.store.ListAllClustersProjects()
+
+	} else {
+		var clusterConfigs *om.AutomationConfig
```
gssbzn: [np] this could be a function but feel free to ignore unless you want to do other changes

internal/cli/cloud_manager_clusters_list.go
```diff
@@ -23,36 +25,41 @@ import (
...
 	opts.store, err = store.New()
 	return err
 }
 
-func (opts *cmClustersListOpts) Run() error {
-	result, err := opts.store.GetAutomationConfig(opts.ProjectID())
+func (opts *cloudManagerClustersListOpts) Run() error {
+
+	var result interface{}
+	var err error
```
gssbzn: good use of a programatic error 👍

internal/cli/cloud_manager_clusters_list_test.go
```diff
@@ -17,32 +17,59 @@ package cli
...
 
 func TestCloudManagerClustersList_Run(t *testing.T) {
 	ctrl := gomock.NewController(t)
-	mockStore := mocks.NewMockAutomationGetter(ctrl)
+	mockStore := mocks.NewMockCloudManagerClustersLister(ctrl)
 
 	defer ctrl.Finish()
 
-	expected := fixtures.AutomationConfig()
+	t.Run("ProjectID is given", func(t *testing.T) {
```
gssbzn: brilliant 💯

internal/cli/cloud_manager_clusters_list_test.go
```diff
@@ -17,32 +17,59 @@ package cli
 import (
 	"testing"
 
+	"github.com/mongodb/mongocli/internal/config"
+
```
gssbzn: small thing, extra empty line, let's remove it and fmtimport this file

PR #48, opened 2020-03-15
CLOUDP-57863: List available checkpoints, atlas/cm/om

andreaangiolillo:
_Jira ticket:_ [CLOUDP-57863](https://jira.mongodb.org/browse/CLOUDP-57863)

## Checklist

- [x] I have tested it by using the AWS test environment


.idea/.gitignore
```diff
@@ -0,0 +1,2 @@
+# Default ignored files
```
gssbzn: can you please rollback this file?
also please consider setting up global ignore files https://gist.github.com/subfuzion/db7f57fff2fb6998a16c

internal/cli/atlas_backups_checkpoints_list.go
```diff
@@ -0,0 +1,92 @@
...
+	itemsPerPage int
+	store        store.CheckpointsLister
+}
+
+func (opts *atlasBackupsCheckpointsListOpts) init() error {
+	if opts.ProjectID() == "" {
+		return errMissingProjectID
+	}
+
+	if opts.clusterName == "" {
```
gssbzn: you don't need this, please do 
```go
_ = cmd.MarkFlagRequired(flags. ClusterName)
```
in line 90, the only reason we do this check for the project id is that a flag is not the only way to set it

internal/cli/atlas_backups_checkpoints_list.go
```diff
@@ -0,0 +1,92 @@
...
+
+// mongocli atlas backup(s) checkpoint(s) list
+func AtlasBackupsCheckpointsListBuilder() *cobra.Command {
+	opts := &atlasBackupsCheckpointsListOpts{
+		globalOpts: newGlobalOpts(),
+	}
+	cmd := &cobra.Command{
+		Use:     "list",
+		Aliases: []string{"ls"},
+		Short:   "List backups checkpoints.",
```
gssbzn:
```suggestion
		Short:   "List continuous backup checkpoints.",
```

internal/cli/atlas_backups_checkpoints_list.go
```diff
@@ -0,0 +1,92 @@
...
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return opts.Run()
+		},
+	}
+
+	cmd.Flags().IntVar(&opts.pageNum, flags.Page, 0, usage.Page)
+	cmd.Flags().IntVar(&opts.itemsPerPage, flags.Limit, 0, usage.Limit)
+
+	cmd.Flags().StringVar(&opts.projectID, flags.ProjectID, "", usage.ProjectID)
+	cmd.Flags().StringVar(&opts.clusterName, flags.ClusterName, "", usage.ClusterNameCheckpoints)
```
gssbzn: the spec describe this as being an argument of the command and not a flag, any reason for this change?
it should be 
```
mongocli atlas backup(s) checkpoint(s) list <clusterId|clusterName>
```
Instead of 
```
mongocli atlas backup(s) checkpoint(s) list --clusterId <clusterId|clusterName> | --clusterName <clusterName>
```
Remember that OM uses ids while atlas uses names

andreaangiolillo: no reason. my mistake 🤕 
Only one thing that may need an update, the spec says 

`mcli atlas|cloud-manager|ops-manager|cm|om backup checkpoints clusterID list|ls [--projectId projectId]`

where clusterID is between **checkpoints** and **list** . That should be an error I guess.. also the definition of clusterID seems taken from the snapshot.

I can update the spec as well if that is okay with you.

gssbzn: I'll fix the spec, let's try to finish this so we can tag a release later today

internal/cli/atlas_backups_checkpoints_list.go
```diff
@@ -0,0 +1,88 @@
...
+// mongocli atlas backup(s) checkpoint(s) list clusterName [--projectId projectId]
+func AtlasBackupsCheckpointsListBuilder() *cobra.Command {
+	opts := &atlasBackupsCheckpointsListOpts{
+		globalOpts: newGlobalOpts(),
+	}
+	cmd := &cobra.Command{
+		Use:     "list",
+		Aliases: []string{"ls"},
+		Short:   "List continuous backup checkpoints.",
+		Args:    cobra.NoArgs,
```
gssbzn: you may need to change this or we won't be able to use args for the cluster name/id

internal/cli/atlas_backups_restores_start.go
```diff
@@ -220,7 +220,7 @@ func AtlasBackupsRestoresStartBuilder() *cobra.Command {
 
 	cmd.Flags().StringVar(&opts.snapshotID, flags.SnapshotID, "", usage.SnapshotID)
 	// Atlas uses cluster name
-	cmd.Flags().StringVar(&opts.clusterName, flags.ClusterName, "", usage.ClusterName)
+	cmd.Flags().StringVar(&opts.clusterName, flags.ClusterName, "", usage.ClusterNameSnapshots)
```
gssbzn: can you please rollback this change as well as there's no need for it now?

internal/store/checkpoints.go
```diff
@@ -0,0 +1,65 @@
...
+import (
+	"context"
+	"fmt"
+
+	atlas "github.com/mongodb/go-client-mongodb-atlas/mongodbatlas"
+	om "github.com/mongodb/go-client-mongodb-ops-manager/opsmngr"
+	"github.com/mongodb/mongocli/internal/config"
+)
+
+type CheckpointsGetter interface {
```
gssbzn: can we delete this as we don't need it right now? same with the implementation and the store?
They can be implemented later when needed

andreaangiolillo:
Question about your comment " we are still missing on adding the command for om and cm": the command is inside the **AtlasBackupsCheckpointsBuilder**, which is inside the **AtlasBackupsBuilder** which is inside OM, Atlas and CM. So the command should be already inside OM and CM. Am I missing something?

gssbzn:
@andreaangiolillo my bad with the altas/om/cm stuff, we need to rethink how we organize the commands, this is getting hard

internal/store/checkpoints.go
```diff
@@ -0,0 +1,60 @@
...
+type CheckpointsLister interface {
+	List(string, string, *atlas.ListOptions) (*atlas.Checkpoints, error)
+}
+
+type CheckpointsStore interface {
+	CheckpointsLister
+}
+
+// Get encapsulate the logic to manage different cloud providers
+func (s *Store) Get(projectID, clusterID, checkpointID string) (*atlas.Checkpoint, error) {
```
gssbzn: we could delete this as well but I won't block this pr for this

PR #49, opened 2020-03-16
CLOUDP-57854: List automated servers

andreaangiolillo:
_Jira ticket:_ [CLOUDP-57854](https://jira.mongodb.org/browse/CLOUDP-57854)

- [x] **I tested it using the AWS testing environment**


internal/cli/ops_manager_agents_list.go
```diff
@@ -0,0 +1,71 @@
+// Copyright 2020 MongoDB Inc
```
gssbzn: [nit] this file should be named `internal/cli/ops_manager_servers_list.go` we (so far) always match the command in our naming

internal/cli/ops_manager_agents_list.go
```diff
@@ -0,0 +1,71 @@
...
+	"github.com/mongodb/mongocli/internal/store"
+	"github.com/mongodb/mongocli/internal/usage"
+	"github.com/spf13/cobra"
+)
+
+const (
+	AgentType = "AUTOMATION"
+)
+
+type OpsManagerAgentsListOpts struct {
```
gssbzn: Same as with file name this should be 
```suggestion
type OpsManagerServersListOpts struct {
```
CLI naming over API names

internal/cli/ops_manager_agents_list.go
```diff
@@ -0,0 +1,71 @@
...
+}
+
+func (opts *OpsManagerAgentsListOpts) init() error {
+	var err error
+	opts.store, err = store.New()
+	return err
+}
+
+func (opts *OpsManagerAgentsListOpts) Run() error {
+	var projects interface{}
```
gssbzn: [q] copy/paste? also you can avoid this here

internal/cli/ops_manager_agents_list.go
```diff
@@ -0,0 +1,71 @@
...
+	var err error
+	projects, err = opts.store.Agents(opts.projectID, AgentType)
```
gssbzn: and by avoid I mean doing
```suggestion
	servers, err := opts.store.Agents(opts.projectID, AgentType)
```

internal/cli/ops_manager_agents_list.go
```diff
@@ -0,0 +1,71 @@
...
+import (
+	"github.com/mongodb/mongocli/internal/flags"
+	"github.com/mongodb/mongocli/internal/json"
+	"github.com/mongodb/mongocli/internal/store"
+	"github.com/mongodb/mongocli/internal/usage"
+	"github.com/spf13/cobra"
+)
+
+const (
+	AgentType = "AUTOMATION"
```
gssbzn: [q] does this need to be public?

internal/cli/ops_manager_agents_list.go
```diff
@@ -0,0 +1,71 @@
...
+// mongocli om server(s) list [--projectId projectId]
+func OpsManagerAgentsListBuilder() *cobra.Command {
+	opts := &OpsManagerAgentsListOpts{
+		globalOpts: newGlobalOpts(),
+	}
+	cmd := &cobra.Command{
+		Use:     "list",
+		Aliases: []string{"ls"},
+		Args:    cobra.NoArgs,
+		Short:   "List all available servers which are running an automation server, they may or may not be running mongod/s processes.",
```
gssbzn:
```suggestion
		Short:   "List all available servers running an automation agent for the given project",
```

internal/cli/ops_manager_agents_list_test.go
```diff
@@ -0,0 +1,56 @@
...
+	expected := fixtures.Agents()
+
+	listOpts := OpsManagerAgentsListOpts{
+		globalOpts: &globalOpts{
+			projectID: "1",
+		},
+		store: mockStore,
+	}
+
+	t.Run("ProjectID is given", func(t *testing.T) {
```
gssbzn: [nit] feel free to ignore this one but you don't need an extra scenario here as you only have one

internal/store/agents.go
```diff
@@ -0,0 +1,41 @@
...
+
+	om "github.com/mongodb/go-client-mongodb-ops-manager/opsmngr"
+	"github.com/mongodb/mongocli/internal/config"
+)
+
+type AgentLister interface {
+	Agents(string, string) (*om.Agents, error)
+}
+
+type AgentStore interface {
```
gssbzn: we can do without this interface since is the same as `AgentLister`, don't forget to regenerate the mocks

internal/cli/ops_manager_agents_list.go
andreaangiolillo: I am using it also in the test

internal/store/agents.go
andreaangiolillo: Yes but we may decide to add also the other agent api call in the future 😄

internal/cli/ops_manager_agents_list.go
gssbzn: it's the same package so is visible in your test even if private

internal/store/agents.go
andreaangiolillo: okay, we will add the interface when we will add the other call 🙄

gssbzn: then that person can add the store in their PR, anyway this is nothing major so I won't block based on this just a goos practice to avoid [Yagni](https://martinfowler.com/bliki/Yagni.html)

internal/cli/ops_manager_servers_list.go
```diff
@@ -22,41 +22,40 @@ import (
...
+func (opts *OpsManagerServersListOpts) init() error {
 	var err error
 	opts.store, err = store.New()
 	return err
 }
 
-func (opts *OpsManagerAgentsListOpts) Run() error {
-	var projects interface{}
+func (opts *OpsManagerServersListOpts) Run() error {
 	var err error
```
gssbzn: you can delete this line

PR #53, opened 2020-03-19
CLOUDP-58790: Take database name from a flag when handling dbusers, Atlas

andreaangiolillo:
_Jira ticket:_ [CLOUDP-58790](https://jira.mongodb.org/browse/CLOUDP-58790)


internal/cli/cli.go
```diff
@@ -60,11 +60,15 @@ type deleteOpts struct {
 	confirm        bool
 	successMessage string
 	failMessage    string
+	authDB         string
 }
 
 // DeleterFromProject a function to delete from the store.
 type DeleterFromProject func(projectID string, entry string) error
```
andreaangiolillo: I am not sure about why we are doing all of this before calling the delete function in the store

internal/cli/cli.go
```diff
@@ -83,6 +87,24 @@ func (opts *deleteOpts) DeleteFromProject(d DeleterFromProject, projectID string
 	return nil
 }
 
+// DeleterFromProjectAuthDB deletes a resource from a project, it expects a callback
```
andreaangiolillo: I do not why after this change every time I run a test regarding any delete I get `=== RUN   TestAtlasDBUsersDelete_Run
%!(EXTRA string=test)--- PASS: TestAtlasDBUsersDelete_Run (0.00s)
PASS`. 

There is this extra string **(EXTRA string=test)**, and I do not know where it came from. I also tried to revert all of my changes but I still get the same result 🤕

internal/usage/usage.go
```diff
@@ -23,6 +23,7 @@ const (
 	DiskSizeGB                      = "Capacity, in gigabytes, of the host’s root volume."
 	Backup                          = "If true, uses Atlas Continuous Backups to back up cluster data."
 	MDBVersion                      = "MongoDB version of the cluster to deploy."
+	AuthDB                          = "Database name."
```
andreaangiolillo: I am not sure about the purpose of this new flag, this is why I put this poor description

gssbzn: let's use "Authentication database name.

internal/cli/cli.go
andreaangiolillo: I fixed the issue by adding the successMessage to the test.  The problem was caused by `fmt.Printf(opts.successMessage, opts.entry)` inside the function **DeleterFromProject** which was executed without the successMessage

internal/cli/atlas_dbusers_delete.go
```diff
@@ -70,6 +70,7 @@ func AtlasDBUsersDeleteBuilder() *cobra.Command {
 	cmd.Flags().BoolVar(&opts.confirm, flags.Force, false, usage.Force)
 
 	cmd.Flags().StringVar(&opts.projectID, flags.ProjectID, "", usage.ProjectID)
+	cmd.Flags().StringVar(&opts.authDB, flags.AuthDB, "", usage.AuthDB)
```
gssbzn: let's make this retro-compatible, this means we don't require it and we default to "admin"
```suggestion
	cmd.Flags().StringVar(&opts.authDB, flags.AuthDB, "admin", usage.AuthDB)
```

internal/cli/cli.go
gssbzn: can we improve on the `deleteOpts` api then? let's do this in another PR but this was probably an overlook from my side, we could improve this by having a fallback message and check the value we are using, like a said separate PR to not complicate this too much

internal/cli/cli.go
```diff
@@ -60,11 +60,15 @@ type deleteOpts struct {
 	confirm        bool
 	successMessage string
 	failMessage    string
+	authDB         string
```
gssbzn: this should not be here, not all deleters should have an `authDB`, this should work the same as a project id that is passed as a parameter of the function

internal/cli/cli.go
```diff
@@ -83,6 +87,24 @@ func (opts *deleteOpts) DeleteFromProject(d DeleterFromProject, projectID string
...
 }
 
+// DeleterFromProjectAuthDB deletes a resource from a project, it expects a callback
+// that should perform the deletion from the store.
+func (opts *deleteOpts) DeleterFromProjectAuthDB(d DeleterFromProjectAuthDB, projectID string) error {
+	if !opts.confirm {
+		fmt.Println(opts.failMessage)
+		return nil
+	}
+	err := d(opts.authDB, projectID, opts.entry)
```
gssbzn: like I mentioned before we should pass `authDB` as a parameter and not as a member of `opts`

PR #57, opened 2020-03-20
CLOUDP-57841: Delete dbusers C/OM

gssbzn:
## Proposed changes

_Jira ticket:_ [CLOUDP-57841](https://jira.mongodb.org/browse/CLOUDP-57841)

## Checklist

### DIFF
--- a/internal/cli/atlas_dbusers_create.go
+++ b/internal/cli/atlas_dbusers_create.go
@@ -31,6 +31,7 @@ type atlasDBUsersCreateOpts struct {
 	*globalOpts
 	username string
 	password string
+	authDB   string
 	roles    []string
 	store    store.DatabaseUserCreator
 }
@@ -58,7 +59,7 @@ func (opts *atlasDBUsersCreateOpts) Run() error {
 
 func (opts *atlasDBUsersCreateOpts) newDatabaseUser() *atlas.DatabaseUser {
 	return &atlas.DatabaseUser{
-		DatabaseName: convert.AdminDB,
+		DatabaseName: opts.authDB,
 		Roles:        convert.BuildAtlasRoles(opts.roles),
 		GroupID:      opts.ProjectID(),
 		Username:     opts.username,
@@ -105,6 +106,7 @@ func AtlasDBUsersCreateBuilder() *cobra.Command {
 	cmd.Flags().StringVar(&opts.username, flags.Username, "", usage.Username)
 	cmd.Flags().StringVar(&opts.password, flags.Password, "", usage.Password)
 	cmd.Flags().StringSliceVar(&opts.roles, flags.Role, []string{}, usage.Roles)
+	cmd.Flags().StringVar(&opts.authDB, flags.AuthDB, convert.AdminDB, usage.AuthDB)
 
 	cmd.Flags().StringVar(&opts.projectID, flags.ProjectID, "", usage.ProjectID)
 
--- a/internal/cli/cli.go
+++ b/internal/cli/cli.go
@@ -77,7 +77,6 @@ func (opts *deleteOpts) Delete(d interface{}, a ...string) error {
 	}
 
 	var err error
-	fmt.Printf("%#v", d)
 	switch f := d.(type) {
 	case func(string) error:
 		err = f(opts.entry)
--- a/internal/cli/cli.go
+++ b/internal/cli/cli.go
@@ -108,7 +108,7 @@ func (opts *deleteOpts) Confirm() error {
 	return survey.AskOne(prompt, &opts.confirm)
 }
 
-// SuccessMessage prints a success message
+// SuccessMessage gets the set success message or the default value
 func (opts *deleteOpts) SuccessMessage() string {
 	if opts.successMessage != "" {
 		return opts.successMessage
@@ -116,7 +116,7 @@ func (opts *deleteOpts) SuccessMessage() string {
 	return fallbackSuccessMessage
 }
 
-// FailMessage prints a fail message
+// FailMessage gets the set fail message or the default value
 func (opts *deleteOpts) FailMessage() string {
 	if opts.failMessage != "" {
 		return opts.failMessage
--- a/internal/cli/cli.go
+++ b/internal/cli/cli.go
@@ -15,6 +15,7 @@
 package cli
 
 import (
+	"errors"
 	"fmt"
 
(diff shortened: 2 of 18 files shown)


internal/cli/atlas_dbusers_create.go
```diff
@@ -58,7 +59,7 @@ func (opts *atlasDBUsersCreateOpts) Run() error {
 
 func (opts *atlasDBUsersCreateOpts) newDatabaseUser() *atlas.DatabaseUser {
 	return &atlas.DatabaseUser{
-		DatabaseName: convert.AdminDB,
+		DatabaseName: opts.authDB,
```
gssbzn: we missed this in the last PR

internal/cli/cli.go
```diff
@@ -24,7 +25,7 @@ import (
 
 const (
 	fallbackSuccessMessage = "'%s' deleted\n"
-	fallbackFailMessage    = "'%s' not deleted\n"
+	fallbackFailMessage    = "entry not deleted"
```
gssbzn: other fail messages don't interpolate so making the fall back compatible

internal/cli/cli.go
```diff
@@ -67,83 +68,32 @@ type deleteOpts struct {
...
-type DeleterFromProject func(projectID string, entry string) error
-
-// DeleterFromProjectAuthDB a function to delete from the store.
-type DeleterFromProjectAuthDB func(authDB string, projectID string, entry string) error
-
-// DeleteFromProject deletes a resource from a project, it expects a callback
+// Delete deletes a resource not associated to a project, it expects a callback
 // that should perform the deletion from the store.
-func (opts *deleteOpts) DeleteFromProject(d DeleterFromProject, projectID string) error {
+func (opts *deleteOpts) Delete(d interface{}, a ...string) error {
```
gssbzn: this function can do now all three functions

internal/cli/cli.go
```diff
@@ -67,83 +68,32 @@ type deleteOpts struct {
...
-// DeleterFromProjectAuthDB deletes a resource from a project, it expects a callback
-// that should perform the deletion from the store.
-func (opts *deleteOpts) DeleterFromProjectAuthDB(d DeleterFromProjectAuthDB, authDB, projectID string) error {
-	if !opts.confirm {
-		opts.printFailMessage()
-		return nil
+	var err error
+	fmt.Printf("%#v", d)
+	switch f := d.(type) {
+	case func(string) error:
```
gssbzn: for some reason I could not use types here so getting rid of them

internal/cli/cli.go
```diff
@@ -157,3 +107,19 @@ func (opts *deleteOpts) Confirm() error {
...
+	return fallbackSuccessMessage
+}
+
+// FailMessage gets the set fail message or the default value
+func (opts *deleteOpts) FailMessage() string {
+	if opts.failMessage != "" {
+		return opts.failMessage
+	}
+	return fallbackFailMessage
+}
```
gssbzn: making this just getters and leaving the printing out

internal/cli/cli.go
```diff
@@ -67,83 +68,32 @@ type deleteOpts struct {
...
-}
 
-// DeleterFromProjectAuthDB deletes a resource from a project, it expects a callback
-// that should perform the deletion from the store.
-func (opts *deleteOpts) DeleterFromProjectAuthDB(d DeleterFromProjectAuthDB, authDB, projectID string) error {
-	if !opts.confirm {
-		opts.printFailMessage()
-		return nil
+	var err error
+	fmt.Printf("%#v", d)
```
andreaangiolillo: [qa] why do we have a print here?

internal/convert/database_user_test.go
```diff
@@ -0,0 +1,99 @@
+package convert
```
gssbzn: we were missing tests for this

internal/cli/cli.go
```diff
@@ -67,83 +68,32 @@ type deleteOpts struct {
...
+	switch f := d.(type) {
```
andreaangiolillo: wow! 👏 😮

internal/cli/cli.go
gssbzn: ha, good catch, I think this is a left over from em doing some testing on why typing wouldn't work

internal/cli/ops_manager_dbusers_delete.go
```diff
@@ -0,0 +1,97 @@
...
+	cmd := &cobra.Command{
+		Use:     "delete [username]",
+		Short:   "Delete a database user for a project.",
+		Aliases: []string{"rm"},
+		Args:    cobra.ExactArgs(1),
+		PreRunE: func(cmd *cobra.Command, args []string) error {
+			if err := opts.init(); err != nil {
+				return err
+			}
+			opts.entry = args[0]
```
andreaangiolillo: [qa] Do you think it is better to set the arg inside PreRun or Run?. I am quite sure that we are setting that arg inside Run in other commands. We should decide one place and change all the other commands in accordance with that decision. I honestly think that the preRun would be a better place 👍

internal/cli/ops_manager_dbusers_delete.go
```diff
@@ -0,0 +1,97 @@
...
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return opts.Run()
+		},
+	}
+
+	cmd.Flags().StringVar(&opts.authDB, flags.AuthDB, convert.AdminDB, usage.AuthDB)
+
+	cmd.Flags().StringVar(&opts.projectID, flags.ProjectID, "", usage.ProjectID)
+
+	_ = cmd.MarkFlagRequired(flags.Username)
```
andreaangiolillo: [qa] why are we using this? username is not a flag right?

gssbzn: oops copy/paste leftover

internal/cli/ops_manager_dbusers_delete.go
gssbzn: Yeah we definitely need to agree on this, I've moved it around based on need, the prompt makes sense to run on the pre run so we need the entry set before prompting, but some validations don't run before preRun, I know flags don't do it but I can't remember if the number of args validation does, if we can confirm it does then we should move them all to a preRun

gssbzn:
@andreaangiolillo RFAL

PR #70, opened 2020-03-26
CLOUDP-57842:  Update alert configs, Atlas

andreaangiolillo:
## Proposed changes

_Jira ticket:_ [CLOUDP-57842](https://jira.mongodb.org/browse/CLOUDP-57842)

## Checklist

- [x] **I have tested it by creating/running the e2e test on the cloud-dev env**

### DIFF
--- a/internal/cli/atlas_alerts_config_update.go
+++ b/internal/cli/atlas_alerts_config_update.go
@@ -23,6 +23,7 @@ import (
 )
 
 type atlasAlertsConfigUpdateOpts struct {
+	*globalOpts
 	*atlasAlertsConfigOpts
 	store   store.AlertConfigurationUpdater
 	alertID string
@@ -56,9 +57,8 @@ func (opts *atlasAlertsConfigUpdateOpts) Run() error {
 // [--projectId projectId]
 func AtlasAlertsConfigUpdateBuilder() *cobra.Command {
 	opts := &atlasAlertsConfigUpdateOpts{
-		atlasAlertsConfigOpts: &atlasAlertsConfigOpts{
-			globalOpts: newGlobalOpts(),
-		},
+		globalOpts:            newGlobalOpts(),
+		atlasAlertsConfigOpts: newAtlasAlertsConfigOpts(),
 	}
 	cmd := &cobra.Command{
 		Use:     "update",
--- a/internal/cli/atlas_alerts_config_update.go
+++ b/internal/cli/atlas_alerts_config_update.go
@@ -23,7 +23,7 @@ import (
 )
 
 type atlasAlertsConfigUpdateOpts struct {
-	*atlasAlertsConfigCreateOpts
+	*AtlasAlertsConfig
 	store   store.AlertConfigurationUpdater
 	alertID string
 }
@@ -56,10 +56,9 @@ func (opts *atlasAlertsConfigUpdateOpts) Run() error {
 // [--projectId projectId]
 func AtlasAlertsConfigUpdateBuilder() *cobra.Command {
 	opts := &atlasAlertsConfigUpdateOpts{
-		atlasAlertsConfigCreateOpts: &atlasAlertsConfigCreateOpts{
+		AtlasAlertsConfig: &AtlasAlertsConfig{
 			globalOpts: newGlobalOpts(),
 		},
-		store: nil,
 	}
 	cmd := &cobra.Command{
 		Use:     "update",
--- a/internal/cli/atlas_alerts_config_update.go
+++ b/internal/cli/atlas_alerts_config_update.go
@@ -1,0 +1,109 @@
+// Copyright 2020 MongoDB Inc
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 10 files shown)


internal/cli/atlas_alerts_config_update.go
```diff
@@ -0,0 +1,109 @@
...
+import (
+	"github.com/mongodb/mongocli/internal/flags"
+	"github.com/mongodb/mongocli/internal/json"
+	"github.com/mongodb/mongocli/internal/store"
+	"github.com/mongodb/mongocli/internal/usage"
+	"github.com/spf13/cobra"
+)
+
+type atlasAlertsConfigUpdateOpts struct {
+	*atlasAlertsConfigCreateOpts
```
gssbzn: [nit] I'm not sure if I like this, we now have `atlasAlertsConfigUpdateOpts` implementing two `init`s and two `Run` methods, If the point is to reuse `buildAlertConfiguration` let's create a third `*Opts` struct and embed that in both 
`atlasAlertsConfigUpdateOpts` and `atlasAlertsConfigUpdateOpts`

gssbzn: you can place that struct in the cli.go file which is kind of our dumping ground for these helpers

internal/cli/cli.go
```diff
@@ -122,3 +124,125 @@ func (opts *deleteOpts) FailMessage() string {
 	}
 	return fallbackFailMessage
 }
+
+type AtlasAlertsConfig struct {
```
gssbzn: [nit] let's follow our naming convention for this structs
```suggestion
type atlasAlertsConfigOpts struct {
```
Lets please also add documentation on what to expect when using this

internal/cli/cli.go
```diff
@@ -122,3 +124,125 @@ func (opts *deleteOpts) FailMessage() string {
...
+
+	case pager:
+		out.ServiceKey = opts.notificationServiceKey
+
+	}
+
+	return out
+}
+
+func newMetricThreshold(opts *AtlasAlertsConfig) *atlas.MetricThreshold {
```
gssbzn: You can make this a function of `AtlasAlertsConfig` (also remember to rename it as above
```suggestion
func (opts *AtlasAlertsConfig) newMetricThreshold *atlas.MetricThreshold {
```

internal/cli/cli.go
```diff
@@ -122,3 +124,125 @@ func (opts *deleteOpts) FailMessage() string {
...
+	enabled                         bool
+	notificationEmailEnabled        bool
+	metricThresholdThreshold        float64
+}
+
+func (opts *AtlasAlertsConfig) buildAlertConfiguration() *atlas.AlertConfiguration {
+
+	out := new(atlas.AlertConfiguration)
+
+	out.GroupID = opts.ProjectID()
```
gssbzn: [nit ]I'd rather pass projectID as a parma of the function instead of depending on `globalOpts` but I don't feel to strongly about it

internal/cli/cli.go
```diff
@@ -122,3 +124,125 @@ func (opts *deleteOpts) FailMessage() string {
...
+	notificationVictorOpsRoutingKey string
+	notificationDelayMin            int
+	notificationIntervalMin         int
+	notificationSmsEnabled          bool
+	enabled                         bool
+	notificationEmailEnabled        bool
+	metricThresholdThreshold        float64
+}
+
+func (opts *AtlasAlertsConfig) buildAlertConfiguration() *atlas.AlertConfiguration {
```
gssbzn: [nit] small ranaming here
```suggestion
func (opts *AtlasAlertsConfig) newAlertConfiguration() *atlas.AlertConfiguration {
```

internal/cli/cli.go
```diff
@@ -122,3 +124,125 @@ func (opts *deleteOpts) FailMessage() string {
...
+	if opts.metricThresholdMetricName != "" {
+		out.MetricThreshold = newMetricThreshold(opts)
+	}
+
+	out.Notifications = []atlas.Notification{*newNotification(opts)}
+
+	return out
+}
+
+func newNotification(opts *AtlasAlertsConfig) *atlas.Notification {
```
gssbzn:
```suggestion
func (opts *AtlasAlertsConfig) newNotification *atlas.Notification {
```

internal/cli/cli.go
```diff
@@ -122,3 +124,125 @@ func (opts *deleteOpts) FailMessage() string {
...
+	return &atlas.MetricThreshold{
+		MetricName: strings.ToUpper(opts.metricThresholdMetricName),
+		Operator:   strings.ToUpper(opts.metricThresholdOperator),
+		Threshold:  opts.metricThresholdThreshold,
+		Units:      strings.ToUpper(opts.metricThresholdUnits),
+		Mode:       strings.ToUpper(opts.metricThresholdMode),
+	}
+}
+
+func newMatcher(opts *AtlasAlertsConfig) *atlas.Matcher {
```
gssbzn: same this could be a function of `AtlasAlertsConfig`

internal/cli/cli.go
```diff
@@ -122,3 +124,126 @@ func (opts *deleteOpts) FailMessage() string {
 	}
 	return fallbackFailMessage
 }
+
+// atlasAlertsConfigOpts contains all the information and functions to manage an alert configuration
+type atlasAlertsConfigOpts struct {
+	*globalOpts
```
gssbzn: if you pass the project id, you no longer need this and can move it back, this is just to facilitate composability of the elements,
if you don't need `globalOpts` don't make others depend on it, this helps to have lighter APIs for the structs a clear use cases (single responsibility) and let's us mix and match in the future (in case we want to include `atlasAlertsConfigOpts ` but not `globalOpts`)

andreaangiolillo: I do not why with my last commit the e2e tests stop working 🤕

PR #75, opened 2020-03-27
CLOUDP-59682: Enable security on C/OM

gssbzn:
## Proposed changes

_Jira ticket:_ [CLOUDP-59682](https://jira.mongodb.org/browse/CLOUDP-59682)

## Checklist

### DIFF
--- a/.githooks/pre-commit
+++ b/.githooks/pre-commit
@@ -1,4 +1,11 @@
 $!/usr/bin/env bash
 
-echo "==> Running pre-commit git hook..."
-make fmt check
+STAGED_GO_FILES=$(git diff --cached --name-only | grep ".go$")
+
+for FILE in $STAGED_GO_FILES
+do
+    gofmt -w -s $FILE
+    goimports -w $FILE
+done
+
+make check
--- a/internal/cli/ops_manager_security_enable.go
+++ b/internal/cli/ops_manager_security_enable.go
@@ -1,0 +1,86 @@
+// Copyright 2020 MongoDB Inc
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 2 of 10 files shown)


.githooks/pre-commit
```diff
@@ -1,4 +1,11 @@
 $!/usr/bin/env bash
 
-echo "==> Running pre-commit git hook..."
-make fmt check
+STAGED_GO_FILES=$(git diff --cached --name-only | grep ".go$")
+
+for FILE in $STAGED_GO_FILES
```
andreaangiolillo: [qa] Does this piece of code format your files before the commit?

internal/cli/ops_manager_security_enable.go
```diff
@@ -0,0 +1,86 @@
...
+	return nil
+}
+
+// mongocli ops-manager security enable[MONGODB-CR|SCRAM-SHA-256]  [--projectId projectId]
+func OpsManagerSecurityEnableBuilder() *cobra.Command {
+	opts := &opsManagerSecurityEnableOpts{
+		globalOpts: newGlobalOpts(),
+	}
+	cmd := &cobra.Command{
+		Use:       "enable [MONGODB-CR|SCRAM-SHA-256]",
```
andreaangiolillo: [qa] why aren't we passing the authentication mechanism as an argument?

.githooks/pre-commit
gssbzn: yes, if you install it, it was taking a lot of time in my machine so I changed to only do it for staged files, should improve speed a little

internal/cli/ops_manager_security_enable.go
gssbzn: what do you mean here exactly?

andreaangiolillo: Could you explain to me what this `enable[MONGODB-CR|SCRAM-SHA-256]` mean? Is `enable` using an argument which can be set as **MONGODB-CR** or **SCRAM-SHA-256** (I do not think so because you are not setting an argument),  or, is the command `enable` always used in the form `enable[MONGODB-CR|SCRAM-SHA-256]`?

gssbzn: the values can be, MONGODB-CR, SCRAM-SHA-256 or both at the same time, check line 78

andreaangiolillo: Thank you 👍

gssbzn:
@andreaangiolillo did a small change here so RFAL

PR #81, opened 2020-04-01
CLOUDP-57849:List events, Atlas

andreaangiolillo:
## Proposed changes

_Jira ticket:_ [CLOUDP-57849](https://jira.mongodb.org/browse/[CLOUDP-57849])

## Checklist


internal/cli/atlas_events_list.go
```diff
@@ -0,0 +1,114 @@
...
+func (opts *atlasEventsListOpts) Run() error {
+	listOpts := opts.newEventListOptions()
+
+	var result *atlas.EventResponse
+	var err error
+
+	source := strings.ToLower(opts.source)
+	switch source {
+	case "organization":
+		result, err = opts.store.OrganizationEvents(opts.projectID, listOpts)
```
andreaangiolillo, line 59: we are using `--projectID `also for calling the organization event which could be confusing. It may be better if we define another flag for it such as `--organizationID `

andreaangiolillo:
Waiting for https://github.com/mongodb/go-client-mongodb-atlas/pull/72

internal/cli/atlas_events_list.go
gssbzn: you're totally right and probably and overlook from us on the spec,
What do you think instead of `--source` we do `--projectId` or `--orgId` (I think we already have one defined), if `--projectId` is set then search projects if `--orgId` is set then search by org, if none, or both are set then error?

andreaangiolillo: Sounds good to me 👍

PR #88, opened 2020-04-03
Modified .goreleaser.yml to create a snap

mdg-2018:
More detials on snaps are available here: https://snapcraft.io/

To automatically upload the snap to Canonical's snap store, add the attribte "publish:true", more detials at goreleaser snapcraft documentation https://goreleaser.com/snapcraft/

## Proposed changes
I'm proposing setting up goreleaser to output a snap package for mongocli. Snap packages work across a variety of Linux distributions  but are most common on Ubuntu. Snap packages can be uploaded to Canonical's snap store so that ubuntu users can easily install mongocli with the command <code>~~snap install mongocli --classic~~</code>
edit: command is now <code>snap install mongocli</code>

Closes #[issue number]

## Checklist

- [N/A] I have added tests that prove my fix is effective or that my feature works
- [N/A] I have run `make fmt` and formatted my code

## Further comments
To install/test the snap package:
- Use Ubuntu 16.04 or later or another distribution with snapd installed
- after running <code>make package</code>, run <code>sudo snap install ./dist/SnapPackageName.snap --classic --dangerous</code>
- run <code>mongocli</code> in the terminal


gssbzn:
@mdg-2018 can you please update with master before we merge?

.goreleaser.yml
```diff
@@ -77,3 +77,17 @@ changelog:
...
     - '^test:'
+    
+snapcrafts:
+  -
+    replacements:
+      386: i386
+      amd64: x86_64
+    summary: The MongoDB Command Line Interface
+    description: The MongoDB Command Line Interface (mongocli) is a tool for managing your MongoDB cloud services, like MongoDB Atlas, MongoDB Cloud Manager, and MongoDB Ops Manager.
+    grade: stable
```
gssbzn, line 88: can you please change this to `devel` as we are not GA yet?

.goreleaser.yml
```diff
@@ -77,3 +77,17 @@ changelog:
...
+    confinement: devmode
```
gssbzn, line 89: could we change this to strict and specify the permissions we need?

.goreleaser.yml
```diff
@@ -77,3 +77,17 @@ changelog:
...
+    license: Apache-2.0
+    base: core18
+    apps:
+      mongocli:
```
gssbzn, line 93: we try to keep files ending on a new line, let's add an enter here

.goreleaser.yml
mdg-2018: done!

.goreleaser.yml
mdg-2018: switched confinement to strict and added network plug

.goreleaser.yml
mdg-2018: done!

gssbzn:
@mdg-2018 sadly this change would break our automated packaging as our CI/CD does not have support for snapcraft and this change would force us to run it there
I was considering to close this PR as adding snapcraft support to evergreen may be too complicated and hard to maintain for us atm but we could do something in between if you're willing to do some extra work to get this merged

Basically what we would need is to split our go releaser in two, one for humans one for evergreen (you can check [.evergreen.yml](https://github.com/mongodb/mongocli/blob/master/.evergreen.yml) on how we do it there)

For humans I think can stay the same, for evergreen we only want binaries and debs/rpms, so we can remove brew and snaps from it

It would be a good idea as well to add some `REALEASING.md` indicating this split and the dependency on snapcraft for the people using `make release` as this would be a new dependency for us

I understand this may be more effort than you were planning for this PR so we'll understand if you decide to close this PR

mdg-2018:
Closing PR since snap support would add unnecessary complexity. Thanks for looking at this!

PR #110, opened 2020-04-17
CLOUDP-60441: Create a rolling index, C/OM

andreaangiolillo:
## Proposed changes

_Jira ticket:_ [CLOUDP-60441](https://jira.mongodb.org/browse/CLOUDP-60441)

## Checklist

- [x] I have tested my code `make with my local ops manager
## Further comments

```
./bin/mongocli om clusters indexes create --collectionName test --db test --rsName myReplicaSet --key location:1  --projectId 5e9d600f338fa44c13bc6cbf --profile OPS`
Changes are being applied, please check http://127.0.0.1:8080/v2/5e9d600f338fa44c13bc6cbf#deployment/topology for status
```


internal/cli/ops_manager_clusters_indexes_create_test.go
```diff
@@ -0,0 +1,114 @@
...
+		store:           mockStore,
+	}
+
+	indexMap, err := createOpts.newIndexMap()
+
+	if err != nil {
+		t.Fatalf("Run() unexpected error: %v", err)
+	}
+
+	if diff := deep.Equal(indexMap, expected); diff[0] != "slice[0].map[keys]: []interface {} != []map[string]interface {}" {
```
andreaangiolillo, line 110: this piece of code converts the `[]map[string]interface{}` of the variable `keys`  in `[]interface{}`
```
indexRec, _ := json.Marshal(index) 
err = json.Unmarshal(indexRec, &inInterface)
```

internal/cli/ops_manager_clusters_indexes_create.go
```diff
@@ -0,0 +1,206 @@
...
+
+func (opts *opsManagerClustersIndexesCreateOpts) newIndexMap() ([]*map[string]interface{}, error) {
+
+	index, err := opts.newIndex()
+
+	if err != nil {
+		return nil, err
+	}
+
+	var inInterface *map[string]interface{}
```
andreaangiolillo, line 104: I haven't found another way to get a  `[]map[string]interface{}` out of a struct.

internal/cli/ops_manager_clusters_indexes_create.go
```diff
@@ -0,0 +1,206 @@
...
+	indexRec, _ := json.Marshal(index)
+	err = json.Unmarshal(indexRec, &inInterface)
+	if err != nil {
```
andreaangiolillo, line 107: I haven't found another way to get a  `[]map[string]interface{}` out of a struct.

internal/fixtures/automation_configs.go
```diff
@@ -258,3 +260,35 @@ func EmptyAutomationConfig() *opsmngr.AutomationConfig {
...
 }
+
+func Indexes() []*map[string]interface{} {
+
+	return []*map[string]interface{}{
+		{
+			"dbName":         "dbname",
+			"collectionName": "collection",
+			"rsName":         "name",
+			"keys": []map[string]interface{}{
```
andreaangiolillo, line 271: the marshal and unmarshal of a `[]map[string]interface{} `creates a `[]interface{}` which has inside a `[]map[string]interface{}`. I do not know how to change this to include ` []interface{}`

internal/cli/ops_manager_clusters_indexes_create_test.go
```diff
@@ -0,0 +1,113 @@
...
+		Times(1)
+
+	err := createOpts.Run()
+
+	if err != nil {
+		t.Fatalf("Run() unexpected error: %v", err)
+	}
+}
+
+func TestOpsManagerClustersIndexesCreate_newIndexMap(t *testing.T) {
```
andreaangiolillo, line 73: It may be better checking newIndexMap inside `TestOpsManagerClustersIndexesCreate_Run`

internal/cli/cli.go
```diff
@@ -269,3 +269,17 @@ func getHostNameAndPort(hostInfo string) (string, int, error) {
 
 	return host[0], port, nil
 }
+
+// indexKeys returns a map of properties starting from a []string which contains properties in the format field:type
```
gssbzn, line 273: this ocmment doesn't make much sense, a more helpful message would be
```suggestion
// indexKeys  takes a slice of values formatted as key:vale and returns an array of [key]:value
```

internal/cli/ops_manager_clusters_indexes_create.go
```diff
@@ -0,0 +1,206 @@
...
+	return err
+}
+
+func (opts *opsManagerClustersIndexesCreateOpts) Run() error {
+	current, err := opts.store.GetAutomationConfig(opts.ProjectID())
+	if err != nil {
+		return err
+	}
+
+	clusterConfigs := convert.FromAutomationConfig(current)
```
gssbzn, line 72: I guess this is for debugging?

internal/cli/ops_manager_clusters_indexes_create.go
gssbzn: Why don't you do a PR in the go client and type it?

internal/cli/ops_manager_clusters_indexes_create.go
andreaangiolillo: mh... I may misunderstand this method. Here I am checking that the replica set name is inside the automationConfig

internal/cli/ops_manager_clusters_indexes_create.go
```diff
@@ -0,0 +1,189 @@
...
+	if err != nil {
+		return err
+	}
+
+	index, err := opts.newIndex()
+	if err != nil {
+		return err
+	}
+
+	current.IndexConfigs = index
```
gssbzn, line 75: this is always empty right? we may need to consider that another operation is in progress and we may need to add and not do a replace

internal/cli/ops_manager_clusters_indexes_create.go
```diff
@@ -0,0 +1,189 @@
...
+}
+
+// mongocli cloud-manager cluster(s) index(es) create [name]  --rsName rsName --dbName dbName [--key field:type] --projectId projectId
+// --locale locale --caseFirst caseFirst --alternate alternate --maxVariable maxVariable --strength strength --caseLevel caseLevel --numericOrdering numericOrdering
+// --normalization normalization --backwards backwards --unique unique --sparse sparse --background background
+func OpsManagerClustersIndexesCreateBuilder() *cobra.Command {
+	opts := &opsManagerClustersIndexesCreateOpts{}
+	cmd := &cobra.Command{
+		Use:   "create",
+		Short: description.CreateCluster,
```
gssbzn, line 151: copy/paste?

andreaangiolillo: I got that description from `AtlasClustersIndexesCreateBuilder` 🤔

gssbzn: then we both have a copy paste error :P

internal/cli/ops_manager_clusters_indexes_create.go
andreaangiolillo: mh.. actually here I am not sure. I think that is not empty when we have already defined an index and we try to add a new one. 
In my ops manager I have this behavior: after creating a new index I see it in the `indexConfig` but I do not see the others that I have already created even if they are still under the index section. When I reload the automation config after a few minutes the `indexConfig`  is empty again but all my indexes are still inside the index section.

andreaangiolillo: is this behavior normal? It does not make sense to me...

gssbzn: It's kind of normal, the way it works is that it only keeps the new index to create, OM will not try to keep this list consistent, as in, if you create it here, and then go and delete it, OM will not try to recreate it (hence why it goes away)

I'm more concern on multiple index creations happening in parallel and being overwritten but it may be an edge case this is why for now this may work but we should try to make this a bit on the safer side later

internal/description/description.go
```diff
@@ -37,8 +37,10 @@ const (
 	ListSnapshots          = "List continuous snapshots for a project."
 	Logs                   = "Download host logs."
 	Clusters               = "Manage clusters for your project."
+	ClustersIndexes        = "Manage cluster indexes for your project."
 	ClustersLong           = "The clusters command provides access to your cluster configurations. You can create, edit, and delete clusters."
 	CreateCluster          = "Create a MongoDB cluster."
+	CreateIndex            = "Create a MongoDB index."
```
gssbzn, line 43: [nit]
```suggestion
	CreateIndex            = "Create a rolling index for your MongoDB cluster."
```

andreaangiolillo:
Waiting for https://github.com/mongodb/go-client-mongodb-ops-manager/pull/20

internal/cli/ops_manager_clusters_indexes_create.go
```diff
@@ -0,0 +1,189 @@
...
+		CaseLevel:       opts.caseLevel,
+		CaseFirst:       opts.caseFirst,
+		Strength:        opts.strength,
+		NumericOrdering: opts.numericOrdering,
+		Alternate:       opts.alternate,
+		MaxVariable:     opts.maxVariable,
+		Normalization:   opts.normalization,
+		Backwards:       opts.backwards,
+	}
+}
```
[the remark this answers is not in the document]
andreaangiolillo: I do not see any advantages this time 🙈

PR #111, opened 2020-04-17
CLOUDP-56474: Validate orgIDs and projectIDs are valid values

gssbzn:
## Proposed changes

_Jira ticket:_ [CLOUDP-56474](https://jira.mongodb.org/browse/CLOUDP-56474)

## Checklist

## Further comments

This is a small change but we had the project validation duplicated everywhere so also centralising this and using now an init function with callbacks we can reuse and extend in a better way

### DIFF
--- a/internal/cli/atlas_events_list.go
+++ b/internal/cli/atlas_events_list.go
@@ -26,15 +26,16 @@ import (
 )
 
 type atlasEventsListOpts struct {
-	globalOpts
 	listOpts
+	orgID     string
+	projectID string
 	eventType string
 	minDate   string
 	maxDate   string
 	store     store.EventsStore
 }
 
-func (opts *atlasEventsListOpts) init() error {
+func (opts *atlasEventsListOpts) initStore() error {
 	var err error
 	opts.store, err = store.New()
 	return err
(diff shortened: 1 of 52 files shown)


internal/cli/cli.go
```diff
@@ -46,6 +55,26 @@ func (opts *globalOpts) ProjectID() string {
 	return opts.projectID
 }
 
+type cmdOpt func() error
+
+// PreRunE is a function to call before running the command,
+// this will validate the project ID and call any additional function pass as a callback
+func (opts *globalOpts) PreRunE(cbs ...cmdOpt) error {
```
gssbzn, line 62: This is our main change, we now have the validation here, and we can pass callbacks with any additional init function that the command may need

internal/cli/cli.go
```diff
@@ -36,6 +37,14 @@ type globalOpts struct {
 	projectID string
 }
 
+func validateObjectID(s string) error {
+	b, err := hex.DecodeString(s)
+	if err != nil || len(b) != 12 {
```
gssbzn, line 42: got this from the go mongodb client

internal/cli/atlas_events_list.go
```diff
@@ -26,15 +26,16 @@ import (
 )
 
 type atlasEventsListOpts struct {
-	globalOpts
```
gssbzn, line 29: this command wasn't using any of `globalOpts` functions so no need to compose it here and I'll just use the properties

PR #113, opened 2020-04-20
CLOUDP-61081: Cleanup fixtures

gssbzn:
## Proposed changes

_Jira ticket:_ [CLOUDP-61081](https://jira.mongodb.org/browse/CLOUDP-61081)

## Checklist

## Further comments

We don't actually need a full working fixture for our test, only a match in types so I'm doing some clean up in preparation to renaming the package to singular


internal/cli/ops_manager_clusters_list.go
```diff
@@ -37,30 +36,22 @@ func (opts *opsManagerClustersListOpts) init() error {
...
-	result, err := opts.cloudManagerClustersListRun()
+	if opts.ProjectID() == "" {
+		result, err := opts.store.ListAllProjectClusters()
+		if err != nil {
+			return err
+		}
+		return json.PrettyPrint(result)
+	}
 
+	clusterConfigs, err := opts.store.GetAutomationConfig(opts.ProjectID())
```
gssbzn, line 47: This is a small drive by fix/clean up, we weren't checking for the error of the get and that was giving me then a nil pointer exception

PR #123, opened 2020-04-24
CLOUDP-60954: mongocli om logs collect

andreaangiolillo:
## Proposed changes

_Jira ticket:_ [CLOUDP-60954](https://jira.mongodb.org/browse/CLOUDP-60954)

## Checklist

## Further comments

```shell
./bin/mongocli om logs collect replicaset  myReplicaSet --sizeRequestedPerFileBytes 64 --logTypes AUTOMATION_AGENT --profile OPS

{
	"id": "5ea2c9dd2f3b443e64fb727d"
}
```


internal/cli/ops_manager_logs_collect.go
```diff
@@ -0,0 +1,84 @@
...
+		ResourceType:              opts.resourceType,
+		ResourceName:              opts.resourceName,
+		Redacted:                  &opts.redacted,
+		SizeRequestedPerFileBytes: opts.sizeRequestedPerFileBytes,
+		LogTypes:                  opts.logTypes,
+	}
+}
+
+// mongocli om logs collect --resourceType type --resourceName resource --sizeRequestedPerFileBytes size --logTypes type --redacted redacted [--projectId projectId]
+func opsManagerLogsCollectOptsBuilder() *cobra.Command {
```
gssbzn, line 62: This should be public
```suggestion
func OpsManagerLogsCollectOptsBuilder() *cobra.Command {
```

internal/cli/ops_manager_logs_collect.go
```diff
@@ -0,0 +1,84 @@
...
+		PreRunE: func(cmd *cobra.Command, args []string) error {
+			return opts.PreRunE(opts.initStore)
+		},
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return opts.Run()
+		},
+	}
+
+	cmd.Flags().StringVar(&opts.resourceName, flags.ResourceName, "", usage.ResourceName)
+	cmd.Flags().StringVar(&opts.resourceType, flags.ResourceType, "", usage.ResourceType)
```
gssbzn, line 76: [q] are `resourceName` and `resourceType` mandatory?
if so two options:
1. mark the flag as mandatory

2. we could move them as args 
`mongocli om logs [resourceType] [resourceName] [...flags]`

What I like of the second one is we can validate the `resourceType` to be in a list of valid values, you can see an example of this pattern here, https://github.com/mongodb/mongocli/blob/744455b48db4d56f1229adbc63d5b4fb612c04bf/internal/cli/config_set.go#L46-L55

internal/cli/ops_manager_logs_collect.go
```diff
@@ -0,0 +1,84 @@
...
+	cmd.Flags().StringArrayVar(&opts.logTypes, flags.LogTypes, nil, usage.LogTypes)
```
gssbzn, line 77: [q] same question can this be empty? should we mark it as mandatory?

internal/store/logs.go
```diff
@@ -20,13 +20,28 @@ import (
...
 	atlas "github.com/mongodb/go-client-mongodb-atlas/mongodbatlas"
+	om "github.com/mongodb/go-client-mongodb-ops-manager/opsmngr"
 	"github.com/mongodb/mongocli/internal/config"
 )
 
 type LogsDownloader interface {
 	DownloadLog(string, string, string, io.Writer, *atlas.DateRangetOptions) error
 }
 
+type Logs interface {
```
gssbzn, line 31: [nit] one line interfaces should be named as what they do, like
```suggestion
type LogCollector interface {
```

internal/usage/usage.go
```diff
@@ -81,6 +81,11 @@ const (
 	RSName                          = "The replica set that the index is built on."
 	Key                             = "Index keys. Should be formatted as field:type."
 	Unique                          = "Create a unique key index."
+	ResourceName                    = "Name of the resource from which to collect logs."
+	ResourceType                    = "Type of resource from which to collect logs."
```
gssbzn, line 85: if you decide to keep these flags let's make this more specific as the content includes logs
```suggestion
	LogResourceName                    = "Name of the resource from which to collect logs."
	LogResourceType                    = "Type of resource from which to collect logs."
```

internal/usage/usage.go
```diff
@@ -81,6 +81,11 @@ const (
...
+	LogTypes                        = "Array of strings specifying the types of logs to collect."
+	SizeRequestedPerFileBytes       = "Size for each log file in bytes."
+	Redacted                        = "If set to true, emails, hostnames, IP addresses, and namespaces in API responses involving this job are replaced with random string values."
```
gssbzn, line 88: same as above
```suggestion
	LogRedacted                        = "If set to true, emails, hostnames, IP addresses, and namespaces in API responses involving this job are replaced with random string values."
```

Or `RedactedLog`

internal/description/description.go
```diff
@@ -91,6 +91,8 @@ A user’s roles apply to all the clusters in the project.`
 	Automation           = "Manage Ops Manager automation config."
 	ShowAutomationStatus = "Show the current status of the automation config."
 	Global               = "Manage Ops Manager global properties."
+	OMLogs               = "Manage Ops Manager log collection job."
```
gssbzn, line 94: [nit] this message is showing for cloud manager so this should probably be something like
```suggestion
	LogCollection               = "Manage log collection jobs."
```

internal/store/logs.go
```diff
@@ -20,13 +20,28 @@ import (
...
+	Collect(string, *om.LogCollectionJob) (*om.LogCollectionJob, error)
```
andreaangiolillo, line 32: Happy to change its name. Maybe to `CreateLogJob`

internal/description/description.go
```diff
@@ -91,6 +91,8 @@ A user’s roles apply to all the clusters in the project.`
...
+	CollectOMLogs        = "Start a job to collect logs"
```
gssbzn, line 95: same, I think this could just be 
```suggestion
	StartLogCollectionJob        = "Start a job to collect logs"
```

internal/cli/ops_manager_logs_collect.go
andreaangiolillo: All the flags are required except for redacted

internal/usage/usage.go
andreaangiolillo: Do you want me to rename also `SizeRequestedPerFileBytes`?

gssbzn: Up to you, I think that one is quite unique for this command but redacted may show up in other commands and we may use it without realising it has log copy in it and not something generic

internal/flags/flags.go
```diff
@@ -115,4 +115,7 @@ const (
 	Normalization                   = "normalization"                   // Normalization flag
 	Backwards                       = "backwards"                       // Backwards flag
 	Strength                        = "strength"                        // Strength flag
+	LogTypes                        = "logTypes"                        // LogTypes flag
+	SizeRequestedPerFileBytes       = "sizeRequestedPerFileBytes"       //SizeRequestedPerFileBytes flag
+	LogRedacted                     = "redacted"                        // LogRedacted flag
```
gssbzn, line 120: This could have stayed `Redacted`, there's no log in the value so anyone should be free to use it

internal/flags/flags.go
```diff
@@ -115,4 +115,7 @@ const (
 	Normalization                   = "normalization"                   // Normalization flag
 	Backwards                       = "backwards"                       // Backwards flag
 	Strength                        = "strength"                        // Strength flag
+	LogTypes                        = "logTypes"                        // LogTypes flag
```
gssbzn, line 118: there's already a `Type` const in this file, let's use that one.
we don't need the log part in the flag since are already in a log command

PR #124, opened 2020-04-24
CLOUDP-61290: Validate base url

gssbzn:
## Proposed changes

_Jira ticket:_ [CLOUDP-61290](https://jira.mongodb.org/browse/CLOUDP-61290)

## Checklist

## Further comments

This is a breaking change, users will need to update their base url to include a trailing slash


internal/cli/cli.go
```diff
@@ -37,14 +37,6 @@ type globalOpts struct {
 	projectID string
 }
 
-func validateObjectID(s string) error {
```
gssbzn, line 40: moved to its own package

andreaangiolillo:
something seems broken in the e2e test. Do you think it is related to this changeset?

gssbzn:
@andreaangiolillo yes I need to change a setting in drone but since you have a PR I'm waiting for that as the change would break you instead

gssbzn:
@andreaangiolillo fix

PR #129, opened 2020-04-25
Clarify that project uses GitHub issues

p-mongo:


gssbzn:
@p-mongo thanks for the improvement

PR #132, opened 2020-04-27
CLOUDP-61529: Pause and Start an atlas cluster

gssbzn:
## Proposed changes

[CLOUDP-61529](https://jira.mongodb.org/browse/CLOUDP-61529)

## Checklist

### DIFF
--- a/internal/cli/atlas_clusters_update.go
+++ b/internal/cli/atlas_clusters_update.go
@@ -50,11 +50,9 @@ func (opts *atlasClustersUpdateOpts) Run() error {
 	}
 	if opts.filename == "" {
 		opts.patchOpts(cluster)
-	} else {
-		cluster.GroupID = opts.ProjectID()
 	}
 
-	result, err := opts.store.UpdateCluster(cluster)
+	result, err := opts.store.UpdateCluster(opts.ProjectID(), opts.name, cluster)
 
 	if err != nil {
 		return err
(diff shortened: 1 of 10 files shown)


internal/store/clusters.go
```diff
@@ -62,10 +62,10 @@ func (s *Store) CreateCluster(cluster *atlas.Cluster) (*atlas.Cluster, error) {
 }
 
 // UpdateCluster encapsulate the logic to manage different cloud providers
-func (s *Store) UpdateCluster(cluster *atlas.Cluster) (*atlas.Cluster, error) {
+func (s *Store) UpdateCluster(projectID, name string, cluster *atlas.Cluster) (*atlas.Cluster, error) {
```
gssbzn, line 65: when pausing/starting a cluster we can only submit, the `Paused` property or they API will error with a "can't update and pause at the same time" even if the values are not changing so we needed to reconfigure this

internal/cli/atlas_clusters_update.go
```diff
@@ -69,8 +67,8 @@ func (opts *atlasClustersUpdateOpts) cluster() (*atlas.Cluster, error) {
 	if opts.filename != "" {
 		cluster = new(atlas.Cluster)
 		err = file.Load(opts.fs, opts.filename, cluster)
-		if opts.name != "" {
-			cluster.Name = opts.name
+		if opts.name == "" {
```
gssbzn, line 70: flipping this so we always use the command property and not the cluster value due to the change on the store signature

PR #149, opened 2020-05-05
Validate the given project ID is valid

gssbzn:
## Proposed changes

Validate the given value is a valid ObjectID

## Checklist

### DIFF
--- a/internal/validate/validate.go
+++ b/internal/validate/validate.go
@@ -24,23 +24,49 @@ import (
 	"github.com/mongodb/mongocli/internal/config"
 )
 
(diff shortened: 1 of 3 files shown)


internal/validate/validate.go
```diff
@@ -49,6 +75,7 @@ func ObjectID(s string) error {
 	return nil
 }
 
+// Credentials
```
andreaangiolillo, line 78: I do not see the value of this comment
```suggestion
```

gssbzn: ups, committed to soon, updated

PR #152, opened 2020-05-05
CLOUDP-57865: Download tar.gz / .zip diagnose archive

andreaangiolillo:
## Proposed changes

_Jira ticket:_ [CLOUDP-57865](https://jira.mongodb.org/browse/CLOUDP-57865)

## Checklist

- [x] I have tested my code using my local OM


internal/cli/ops_manager_diagnose_archive.go
```diff
@@ -0,0 +1,31 @@
...
+package cli
+
+import (
+	"github.com/mongodb/mongocli/internal/description"
+	"github.com/spf13/cobra"
+)
+
+func OpsManagerDiagnoseArchive() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "diagnose-archive",
```
andreaangiolillo, line 24: Should we create the command `diagnose` and `archive` or leave it as `diagnose-archive`?

internal/cli/ops_manager_diagnose_archive_download.go
```diff
@@ -0,0 +1,99 @@
...
+		},
+	}
+
+	cmd.Flags().StringVarP(&opts.out, flags.Out, flags.OutShort, "", usage.LogOut)
+	cmd.Flags().Int64Var(&opts.limit, flags.Limit, 0, usage.ArchiveLimit)
+	cmd.Flags().Int64Var(&opts.minutes, flags.Minutes, 0, usage.ArchiveMinutes)
+
+	cmd.Flags().StringVar(&opts.projectID, flags.ProjectID, "", usage.ProjectID)
+
+	_ = cmd.MarkFlagRequired(flags.Out)
```
gssbzn, line 96: Let's not make it mandatory but use diagnose-archive.tar.gz` as the default value of the flag

internal/store/diagnose_archive.go
```diff
@@ -0,0 +1,40 @@
...
+
+package store
+
+import (
+	"context"
+	"fmt"
+	"io"
+
+	"go.mongodb.org/ops-manager/opsmngr"
+
```
gssbzn, line 23: extra new line

internal/cli/ops_manager_diagnose_archive_download.go
```diff
@@ -0,0 +1,99 @@
...
+// limitations under the License.
+
+package cli
+
+import (
+	"io"
+	"os"
+
+	"go.mongodb.org/ops-manager/opsmngr"
+
```
gssbzn, line 22: extra line

internal/cli/ops_manager_diagnose_archive.go
gssbzn: let's leave it like this this is not going to be a main feature I belive

internal/cli/ops_manager_diagnose_archive_download.go
andreaangiolillo: If `out` were mandatory I would not know which name I should use to open the file in `newWriteCloser`
```
func (opts *opsManagerDiagnoseArchiveDownloadOpts) newWriteCloser() (io.WriteCloser, error) {
	// Create file only if is not there already (don't overwrite)
	ff := os.O_CREATE | os.O_TRUNC | os.O_WRONLY | os.O_EXCL
	f, err := opts.fs.OpenFile(opts.out, ff, 0777)
	return f, err
}
```

gssbzn: Flags can have default values
`cmd.Flags().StringVarP(&opts.out, flags.Out, flags.OutShort, "diagnose-archive.tar.gz", usage.LogOut)`, sane defaults over a complicated API

internal/cli/ops_manager_diagnose_archive_download.go
```diff
@@ -0,0 +1,96 @@
...
+		Short:   description.DownloadDiagnoseArchive,
+		PreRunE: func(cmd *cobra.Command, args []string) error {
+			return opts.PreRunE(opts.initStore)
+		},
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return opts.Run()
+		},
+	}
+
+	cmd.Flags().StringVarP(&opts.out, flags.Out, flags.OutShort, "diagnose-archive.tar.gz", usage.LogOut)
```
gssbzn, line 89: just caught this, `LogOut` has log specific terms, "Optional output filename, if none given will use the log name."
let's create DiagnoseOut, and make it something like "Optional output filename, if none given will use diagnose-archive.tar.gz"

internal/cli/ops_manager_diagnose_archive_download.go
```diff
@@ -0,0 +1,96 @@
...
+	opts.store, err = store.New()
+	return err
+}
+
+func (opts *opsManagerDiagnoseArchiveDownloadOpts) Run() error {
+	out, err := opts.newWriteCloser()
+	if err != nil {
+		return err
+	}
+
```
gssbzn, line 50: we are missing something here
```suggestion
	defer out.Close()
```
I saw and I think we are missing this for the OM logs command as well, can you do a small fly by fix there?

PR #160, opened 2020-05-07
CLOUDP-62053: It is not possible to unack an alert

andreaangiolillo:
## Proposed changes

_Jira ticket:_ [CLOUDP-62053](https://jira.mongodb.org/browse/CLOUDP-62053)

## Checklist

- [x] I have tested it using my local OM
## Further comments

**ACK:**
```
./bin/mongocli om alerts ack 5eac0cb0323b0e4445a7b477  -P OPS --until 2021-10-01T00:00:00-0400
{
	"id": "5eac0cb0323b0e4445a7b477",
	"groupId": "5ea6a50fec050149b3105d1e",
	"alertConfigId": "5ea6a50fec050149b3105d1f",
	"eventTypeName": "MONITORING_AGENT_DOWN",
	"created": "2020-05-01T11:49:04Z",
	"updated": "2020-05-07T10:51:18Z",
	"status": "OPEN",
	"lastNotified": "2020-05-07T08:20:29Z",
	"acknowledgedUntil": "2021-10-01T03:59:18Z",
	"acknowledgingUsername": "crzrxjqz"
}
```
**ACK FOREVER:**
```
./bin/mongocli om alerts ack 5eac0cb0323b0e4445a7b477  -P OPS --forever
{
	"id": "5eac0cb0323b0e4445a7b477",
	"groupId": "5ea6a50fec050149b3105d1e",
	"alertConfigId": "5ea6a50fec050149b3105d1f",
	"eventTypeName": "MONITORING_AGENT_DOWN",
	"created": "2020-05-01T11:49:04Z",
	"updated": "2020-05-07T10:50:27Z",
	"status": "OPEN",
	"lastNotified": "2020-05-07T08:20:29Z",
	"acknowledgedUntil": "2120-06-08T11:49:27Z",
	"acknowledgingUsername": "crzrxjqz"
}
```
**UNACK:**
```
./bin/mongocli om alerts ack 5eac0cb0323b0e4445a7b477  -P OPS
{
	"id": "5eac0cb0323b0e4445a7b477",
	"groupId": "5ea6a50fec050149b3105d1e",
	"alertConfigId": "5ea6a50fec050149b3105d1f",
	"eventTypeName": "MONITORING_AGENT_DOWN",
	"created": "2020-05-01T11:49:04Z",
	"updated": "2020-05-07T10:51:26Z",
	"status": "OPEN",
	"lastNotified": "2020-05-07T08:20:29Z"
}
```


andreaangiolillo:
Getting this error when running make check (it might be related to https://github.com/mongodb/mongocli/pull/159):

```
==> Linting all packages...
golangci-lint run ./...
ERRO Running error: no such linter "nolintlint"
make: *** [lint] Error 3
```

internal/cli/atlas_alerts_acknowledge.go
```diff
@@ -52,20 +53,18 @@ func (opts *atlasAlertsAcknowledgeOpts) Run() error {
...
 	// To acknowledge an alert “forever”, set the field value to 100 years in the future.
-	if until == "" {
-		until = time.Now().AddDate(100, 1, 1).Format(time.RFC3339)
+	if opts.forever {
+		opts.until = time.Now().AddDate(100, 1, 1).Format(time.RFC3339)
 	}
 
 	return &atlas.AcknowledgeRequest{
-		AcknowledgedUntil:      until,
+		AcknowledgedUntil:      opts.until,
```
gssbzn, line 62: [q] do we need to update the client? this field has an omit empty so I'm wondering how is it working with blank?

andreaangiolillo: I don't think we have to update the atlas client. The field is set to `nil` when the `until` flag is black which is what we want here.

gssbzn: `AcknowledgedUntil` is not a pointer so it can't be nil, only empty ("")
when we serialize it it would only be something like an empty object `{}` so maybe the API is assuming that since there's no key is null?

andreaangiolillo: I thought that the serialization of an empty string with `omitempty`  was `nil` since I got `null` when I printed the value of until in the server.

gssbzn: `omitempty` means that it won't serialize/deserialize that key not about its content, the content defines the rule for `omitempty`, which in the case of string is the empty string ""

andreaangiolillo: Yes, I noticed it, I was reading this article https://www.sohamkamani.com/golang/2018-07-19-golang-omitempty/ that explains it very well. Regarding your question, as you previously said, the only explanation seems to be that the API set the object to null when there is no key (which makes sense).

gssbzn: yes, omitempty is confusing, I know, is one of the things took me the most to get on go, and we can trip a lot with it
if you are sure this works as expected we can merge but we need to be aware that a change on the API can break this command

andreaangiolillo: I am quite confident that it is working as expected but we can make `AcknowledgedUntil` in the atlas client a pointer just to be sure that this behaviour is not going to change in the future. Let me know if you want me to open a PR in the altlas go client.

gssbzn: @andreaangiolillo sorry, was busy and forgot about this, up to you, I'm ok to merge this as it is and we can improve the client later

PR #167, opened 2020-05-11
Experimental msi packaging

gssbzn:
## Proposed changes

This let's us package an msi installer with evergreen

## Checklist

### DIFF
--- a/.evergreen.yml
+++ b/.evergreen.yml
@@ -9,11 +9,8 @@ ignore:
   - "*.txt"
 
 variables:
-  - &go_bin
-      "/opt/golang/go1.14/bin"
   - &go_options
     GO111MODULE: "on"
-    GOROOT: "/opt/golang/go1.14"
 functions:
   "clone":
     - command: git.get_project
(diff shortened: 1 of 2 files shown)


.evergreen.yml
```diff
@@ -57,27 +82,67 @@ functions:
...
+          go-msi check-env
+
+          VERSION=$(git describe | cut -d "v" -f 2)
+
+          env GOOS=windows GOARCH=amd64 go build -ldflags "-s -w -X github.com/mongodb/mongocli/internal/version.Version=$VERSION" -o mongocli.exe
+
+          mkdir {build,dist}
+
+          go-msi make --msi "dist/mongocli_$VERSION.msi" --version $VERSION
+
```
colm-quinn, line 145: Wow! Having had to do some msi work in the past, that's a _lot_ easier in go than other configurations.

wix.json
```diff
@@ -0,0 +1,27 @@
...
+  "license": "LICENSE",
+  "upgrade-code": "a995cda5-9731-4158-910c-c8d15371dbda",
+  "files": {
+    "guid": "c7b0df57-c3c1-4591-becd-f0f73a912eb8",
+    "items": [
+      "mongocli.exe"
+    ]
+  },
+  "env": {
+    "guid": "94faac3d-4478-431c-8497-fba55dcfb249",
```
colm-quinn, line 13: [q] Does this need to change for each version, so should be generated? My memories of msi is each installer has a new guide and the upgrade-code is how you track that it's the same product.

.evergreen.yml
gssbzn: well the hard part was making it run in evg

wix.json
gssbzn: according to go-msi this should be done just once `go-msi set-guid` so hopefully this won't change again(?)

colm-quinn: Cool, fingers crossed. I remember in the past changing the product guid each time, but needing to keep the same upgrade guid to link releases. This will come out in the wash if there's a problem.

PR #169, opened 2020-05-13
CLOUDP-62567: Enable security fails if only scramsha-256 is given

gssbzn:
## Proposed changes

_Jira ticket:_ [CLOUDP-62567](https://jira.mongodb.org/browse/CLOUDP-62567)

## Checklist

### DIFF
--- a/internal/convert/automation_config.go
+++ b/internal/convert/automation_config.go
@@ -17,9 +17,8 @@ package convert
 import (
 	"fmt"
 
-	"go.mongodb.org/ops-manager/atmcfg"
+	"github.com/mongodb/mongocli/internal/search"
 	"go.mongodb.org/ops-manager/opsmngr"
-	"go.mongodb.org/ops-manager/search"
 )
 
 const (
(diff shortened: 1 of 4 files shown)


internal/convert/automation_config.go
```diff
@@ -56,20 +55,34 @@ func FromAutomationConfig(in *opsmngr.AutomationConfig) (out []ClusterConfig) {
...
+	cr        = "MONGODB-CR"
+	sha256    = "SCRAM-SHA-256"
+)
 
 func EnableMechanism(out *opsmngr.AutomationConfig, m []string) error {
-	out.Auth.DeploymentAuthMechanisms = append(out.Auth.DeploymentAuthMechanisms, m...)
-	out.Auth.AutoAuthMechanisms = append(out.Auth.AutoAuthMechanisms, m...)
 	out.Auth.Disabled = false
+	for _, v := range m {
+		if v != cr && v != sha256 {
```
gssbzn, line 67: we only want to support this for now

internal/convert/automation_config.go
```diff
@@ -56,20 +55,34 @@ func FromAutomationConfig(in *opsmngr.AutomationConfig) (out []ClusterConfig) {
...
+			return fmt.Errorf("unsupported mechanism %s", v)
+		}
+		if v == sha256 {
```
gssbzn, line 70: this was the problem, this would default to cr and then fo into an infinite loop in trying to resolve the config

internal/convert/automation_config.go
```diff
@@ -56,20 +55,34 @@ func FromAutomationConfig(in *opsmngr.AutomationConfig) (out []ClusterConfig) {
...
+			out.Auth.AutoAuthMechanism = v
+		}
+		if !search.StringInSlice(out.Auth.DeploymentAuthMechanisms, v) {
```
gssbzn, line 73: avoid adding a mechanism that already exists

internal/convert/automation_config.go
```diff
@@ -56,20 +55,34 @@ func FromAutomationConfig(in *opsmngr.AutomationConfig) (out []ClusterConfig) {
...
+		}
+	}
 
 	if out.Auth.AutoUser == "" {
 		if err := setAutoUser(out); err != nil {
 			return err
 		}
 	}
-	addMonitoringUser(out)
-	addBackupUser(out)
```
gssbzn, line 72: these are not needed anymore but we may need to do something different

PR #173, opened 2020-05-18
CLOUDP-62721: Move Atlas E2E to an evergreen task

andreaangiolillo:
## Proposed changes

_Jira ticket:_ [CLOUDP-62721](https://jira.mongodb.org/browse/CLOUDP-62721)

## Checklist


.evergreen.yml
```diff
@@ -41,6 +41,30 @@ functions:
           GOROOT: ${go_root}
           <<: *go_options
         command: make test
+  "atlas e2e":
+    - command: subprocess.exec
+        type: test
+        params:
+          add_to_path:
+            - ${go_bin}
+          working_dir: src/github.com/mongodb/mongocli/e2e/atlas
```
gssbzn: if your working dir is `src/github.com/mongodb/mongocli/e2e/atlas` you won't have access to the `male file` which is a couple of levels before

.evergreen.yml
```diff
@@ -41,6 +41,30 @@ functions:
...
+          env:
+            WORKDIR: ${workdir}
+            GOROOT: ${go_root}
+            <<: *go_options
+          command: make e2e-test-atlas
+  "iam e2e":
```
gssbzn, line 56: [q] any reason to split this here? you are increasing the challenges of the PR by just also changing how tests run which was working fine before

.evergreen.yml
```diff
@@ -292,6 +316,26 @@ tasks:
     commands:
       - func: "clone"
       - func: "test"
+  - name: atlas_e2e
+    commands:
+      - func: "atlas e2e"
+        params:
```
gssbzn, line 322: these can't be `params` but `env` if you want to be able to use them as environment variables
(`env` is only available in `command: subprocess.exec` type of commands)

Makefile
```diff
@@ -90,6 +90,18 @@ build: ## Generate a binary in ./bin
 	@echo "==> Building binary"
 	go build -ldflags "${LINKER_FLAGS}" -o ${DESTINATION}
 
+.PHONY: e2e-test-atlas
```
gssbzn, line 93: this split was not part of the ticket and like I mention before now adds to the complexity of the PR
also with the changes you're making does `make e2e-test` still works?

.evergreen.yml
andreaangiolillo: I have changed all the e2e file to reflect that change. I have also tested that those two new commands are working as expected

Makefile
andreaangiolillo: yes, it is working. I decided to create an atlas folder inside the e2e because we will have to run ops_manager on e2e as well and those e2e test files must be inside different folders I guess. I have created iam just because I thought will be better to keep them separated. I will revert that change if you wish.

.evergreen.yml
andreaangiolillo: I thought you meant that the make e2e-test-atlas would not work when running` make e2e-test-atlas` inside the terminal (that is what I tested). Here you mean that the evergreen task cannot access the make file I guess. I suppose there is a way out.

andreaangiolillo: I am going to try `../../make e2e-test-atlas`

gssbzn: the correct solution here is to change your working dir to be `src/github.com/mongodb/mongocli`

andreaangiolillo: mh... The reason why I did that change is that I thought that working_dir must contain the directory of the e2e test (I don't know why, I guess because I associated working directory to "where the job must be done" instead of "where is the file that run my test". I am going to update working_dir with src/github.com/mongodb/mongocli

gssbzn: `working dir` is where you need to run your commands evergreen has not context on what your test are or how they work, since your plan is to run `make`, and you need a makefile for that then you should work from the folder where that makefile is, the make file already has the logic on how to run those test and you should not [leak](https://en.wikipedia.org/wiki/Leaky_abstraction) that up to evergreen

.evergreen.yml
```diff
@@ -41,6 +41,40 @@ functions:
...
+        working_dir: src/github.com/mongodb/mongocli
+        env:
+          WORKDIR: ${workdir}
+          GOROOT: ${go_root}
+          <<: *go_options
+          MCLI_OPS_MANAGER_URL: ${mcli_ops_manager_url}
+          MCLI_ORG_ID: ${mcli_org_id}
+          MCLI_PRIVATE_API_KEY: ${mcli_private_api_key}
+          MCLI_PROJECT_ID: ${mcli_project_id}
+          MCLI_PUBLIC_API_KEY: ${mcli_public_api_key}
```
gssbzn, line 59: small nit, let leave the append (`<<`)  to the end
```suggestion
          WORKDIR: ${workdir}
          GOROOT: ${go_root}
          MCLI_OPS_MANAGER_URL: ${mcli_ops_manager_url}
          MCLI_ORG_ID: ${mcli_org_id}
          MCLI_PRIVATE_API_KEY: ${mcli_private_api_key}
          MCLI_PROJECT_ID: ${mcli_project_id}
          MCLI_PUBLIC_API_KEY: ${mcli_public_api_key}
          <<: *go_options
```

.evergreen.yml
```diff
@@ -41,6 +41,40 @@ functions:
...
+    - command: subprocess.exec
+      type: test
+      params:
+        add_to_path:
+          - ${go_bin}
+        working_dir: src/github.com/mongodb/mongocli
+        env:
+          WORKDIR: ${workdir}
+          GOROOT: ${go_root}
+          <<: *go_options
```
gssbzn, line 71: same comment as above, let's move this line to the end of the block

Makefile
gssbzn: the main reason of my comment is that we are adding additional changes that make the PR a bit harder to review, not only now we need to review evergreen works but like my question signals, does this break expected behaviour? no need to revert this but keep that in mind for future PRs, if there are things to improve is always a good idea to just open a new PR for it

.evergreen.yml
```diff
@@ -41,6 +41,40 @@ functions:
...
+          MCLI_OPS_MANAGER_URL: ${mcli_ops_manager_url}
+          MCLI_ORG_ID: ${mcli_org_id}
```
gssbzn, line 73: non blocking and can be done later, we may need to reconsider this name `mcli_org_id` as we are going to add cloud manager later and the id may be different, so I would do something like 
` MCLI_ORG_ID: ${atlas_org_id}` and later add a ` MCLI_ORG_ID: ${cloud_manager_org_id}` for cloud manager

We can address this in another PR but wanted to let you know

Makefile
```diff
@@ -90,6 +90,18 @@ build: ## Generate a binary in ./bin
...
+e2e-test-atlas: build ## Run Atlas E2E tests
+	@echo "==> Running E2E tests..."
+	# the target assumes the MCLI-* environment variables are exported
+	go test -v -p 1 -parallel 1 -tags=e2e ./e2e/atlas...
```
gssbzn, line 97: This is merely informative and you don't need to change anything here but another thing you could've done here is to change the build tags to something like`-tags=atlas_e2e` 
and then in the test file change line 14 to `+build atlas_e2e`

This is a golang thing to specify build targets
By default go will run all your test files, and that tag is how we prevent e2e from running when you do a simple `go test .` command

.evergreen.yml
andreaangiolillo: I am going to do this change in the next PR

PR #174, opened 2020-05-18
CLOUDP-62722: Set up and e2e scenario/task for cloud manager

andreaangiolillo:
## Proposed changes

_Jira ticket:_ [CLOUDP-62722](https://jira.mongodb.org/browse/CLOUDP-62722)

## Checklist


e2e/cloud_manager/list_project_test.go
```diff
@@ -0,0 +1,48 @@
...
+package cloud_manager
+
+import (
+	"os"
+	"os/exec"
+	"path/filepath"
+	"testing"
+)
+
+func TestAtlasIAMProjects(t *testing.T) {
```
andreaangiolillo, line 25: This file is only for testing that evergreen is working as expected. I will remove this file once we start to implement e2e test for cloud manager

gssbzn: what does this mean? do you plan to remove this later on? from a quick review of evergreen I think you could use the same `make e2e-test-iam` given that everything is controlled by environment variables so those changing should make your test "just work(?)"

andreaangiolillo: yes, I was thinking to remove this later on.

andreaangiolillo:
@gssbzn ready for review

.evergreen.yml
```diff
@@ -69,12 +70,31 @@ functions:
...
+          WORKDIR: ${workdir}
+          GOROOT: ${go_root}
+          MCLI_OPS_MANAGER_URL: ${mcli_ops_manager_url}
+          MCLI_ORG_ID: ${cloud_manager_org_id}
+          MCLI_PROJECT_ID: ${cloud_manager_project_id}
+          MCLI_PRIVATE_API_KEY: ${cloud_manager_private_api_key}
+          MCLI_PUBLIC_API_KEY: ${cloud_manager_public_api_key}
+          MCLI_SERVICE: ${cloud_manager_service}
+          <<: *go_options
+        command: make e2e-test-cloud
```
gssbzn, line 97: [q] any reason not to just do `make e2e-test-iam`? probably we can rename the old iam to atlas iam, and this to cloud manager iam

andreaangiolillo: My idea was that the command` make e2e-test-cloud` would contain also the e2e/cloud manager folder such as

```
.PHONY: e2e-test-cloud
e2e-test-cloud: build ## Run IAM E2E tests
	@echo "==> Running Cloud Manger E2E tests..."
	# the target assumes the MCLI-* environment variables are exported
	go test -v -p 1 -parallel 1 -tags=e2e ./e2e/iam...
	go test -v -p 1 -parallel 1 -tags=e2e ./e2e/cloud_manager...
```
I think it is handy to have an e2e command for cloud manager which contains also iam. But I think that we can live without it.

PR #177, opened 2020-05-18
CLOUDP-62814: add DBUsers e2e test - cloud manager

andreaangiolillo:
## Proposed changes

_Jira ticket:_ [CLOUDP-62814](https://jira.mongodb.org/browse/CLOUDP-62814)

## Checklist


e2e/cloud_manager/dbusers_test.go
```diff
@@ -0,0 +1,102 @@
...
+	t.Run("Delete", func(t *testing.T) {
+		cmd := exec.Command(cliPath, entity, dbusersEntity, "delete", username, "--force", "--authDB", "admin")
+		cmd.Env = os.Environ()
+		resp, err := cmd.CombinedOutput()
+
+		if err != nil {
+			t.Fatalf("unexpected error: %v, resp: %v", err, string(resp))
+		}
+
+		expected := "Changes are being applied, please check https://cloud-dev.mongodb.com/v2/5ec2839e74c5aa25f02ff8ee#deployment/topology for status\n"
```
gssbzn, line 96: couple fo nits here
first, great on adding this check but can we add it to the create as well?
second, I'd change this to use something like https://golang.org/pkg/strings/#Contains as the id and url may change (remember those are controlled by env variables) or use those env variables instead but my preference right now is to go with contains and just "Changes are being applied"

PR #179, opened 2020-05-20
CLOUDP-62907: Deploying an on-demand host in evergreen

andreaangiolillo:
## Proposed changes

_Jira ticket:_ [CLOUDP-62907](https://jira.mongodb.org/browse/CLOUDP-62907)

## Checklist


andreaangiolillo:
@gssbzn I got this error
`[2020/05/20 13:37:28.675] Command failed: problem getting hosts list: 0 hosts of 1 are up, waiting
[2020/05/20 13:37:28.675] Task completed - FAILURE.
[2020/05/20 13:37:28.678] Running post-task commands.
 `

.evergreen.yml
```diff
@@ -76,6 +77,79 @@ functions:
...
+          with open(sys.argv[1]) as hostsfile:
+              hosts = json.load(hostsfile)
+              for host in hosts:
+                  print(host["dns_name"])
+          EOF
+          )
+          for host in $hosts; do
+              set +e
+              echo "installing the automation agent on $host"
+              ssh -i "$keyfile" -o ConnectTimeout=10 -o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no -tt "$user@$host" \  < scripts/automation_agent.sh
```
andreaangiolillo, line 137: Here I am calling the script that should install the automation agent

.evergreen.yml
```diff
@@ -76,6 +77,79 @@ functions:
...
+          set -ex
+
+          while getopts 'i:h:t:u:' opt; do
+              case $opt in
+                  i) keyfile="$OPTARG";;      # SSH identity file
+                  u) user="$OPTARG";;         # Username on the remote host
+                  h) hostsfile="$OPTARG";;    # Output of Evergreen host.list
+                  *) exit 1;;
+              esac
+          done
```
gssbzn, line 123: this is not going to work here

scripts/automation_agent.sh
```diff
@@ -0,0 +1,42 @@
...
+     cyrus-sasl-plain krb5-libs libcurl \
+     lm_sensors-libs net-snmp net-snmp-agent-libs \
+     openldap openssl tcp_wrappers-libs xz-libs
+
+echo "Download and extract the automation agent"
+curl -OL https://cloud-dev.mongodb.com/download/agent/automation/mongodb-mms-automation-agent-manager-10.15.0.6396-1.x86_64.rhel7.rpm
+sudo rpm -U mongodb-mms-automation-agent-manager-10.15.0.6396-1.x86_64.rhel7.rpm
+
+echo "Replacing mmsGroupId and mmsApiKey properties"
+replace_property_in_file "/etc/mongodb-mms/automation-agent.config" "mmsGroupId" "5ec2839e74c5aa25f02ff8ee"
```
gssbzn, line 32: hardcoded id

scripts/automation_agent.sh
```diff
@@ -0,0 +1,42 @@
...
+replace_property_in_file "/etc/mongodb-mms/automation-agent.config" "mmsApiKey" "5ec51babe380d40ec2710bd95cf2f39c01fb6ad8647a82da5228dde8"
```
gssbzn, line 33: hardcoded key, please disable this key ASAP

build/package/automation_agent.sh
```diff
@@ -0,0 +1,44 @@
+#!/bin/bash
+
+set -ex
+
+PATH="$PATH":/bin:/usr/bin:/usr/local/bin
+
```
andreaangiolillo, line 6: I don't know why I don't see the spawn host in the cloud-manager UI. Could you have a look at this script and tell me if you find something wrong? I tested in my local machine the function `replace_property_in_file` and it works. @gssbzn

andreaangiolillo:
@gssbzn Ready for review 😄

build/package/install_agent_spawn_host.sh
```diff
@@ -0,0 +1,53 @@
...
+    hosts = json.load(hostsfile)
+    for host in hosts:
+        print(host["dns_name"])
+EOF
+)
+for host in $hosts; do
+    set +e
+    echo "Installing dependeces on $host"
+    ssh -i "$keyfile" -o ConnectTimeout=50 -o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no -tt "$user@$host" 'bash -s' <<'ENDSSH'
+        free -m
```
gssbzn, line 29: [q] can we remove this?

build/package/install_agent_spawn_host.sh
```diff
@@ -0,0 +1,53 @@
...
+        sudo apt-get install -y --no-install-recommends ca-certificates curl logrotate openssl snmp && exit
+ENDSSH
+    echo "Installing the automation agent on $host"
+    ssh -i "$keyfile" -o ConnectTimeout=50 -o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no -tt "$user@$host" ARG1="$groupid" ARG2="$apiKey" 'bash -s' <<'ENDSSH'
+        free -m
```
gssbzn, line 34: same question, is there any point on knowing available memory?

build/package/install_agent_spawn_host.sh
```diff
@@ -0,0 +1,53 @@
...
+        sudo sed -i "s/\(mmsGroupId *= *\).*/\1$ARG1/" /etc/mongodb-mms/automation-agent.config
+        sudo sed -i "s/\(mmsApiKey *= *\).*/\1$ARG2/" /etc/mongodb-mms/automation-agent.config
+
+        echo "Preparing the /data directory to store your MongoDB data"
+        sudo mkdir -p /data
+        sudo chown mongod:mongod /data
+
+        echo "Starting the agent"
+        sudo systemctl start mongodb-mms-automation-agent.service
+        exit
```
gssbzn, line 49: you can create another bash file you can reference here, this would make this more maintainable
like, having a `agent-config.sh` file with this and then here do something like 
`ssh ... 'bash ~/agent-config.sh`

build/package/install_agent_spawn_host.sh
```diff
@@ -0,0 +1,53 @@
...
+)
+for host in $hosts; do
+    set +e
+    echo "Installing dependeces on $host"
+    ssh -i "$keyfile" -o ConnectTimeout=50 -o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no -tt "$user@$host" 'bash -s' <<'ENDSSH'
+        free -m
+        sudo apt-get install -y --no-install-recommends ca-certificates curl logrotate openssl snmp && exit
+ENDSSH
+    echo "Installing the automation agent on $host"
+    ssh -i "$keyfile" -o ConnectTimeout=50 -o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no -tt "$user@$host" ARG1="$groupid" ARG2="$apiKey" 'bash -s' <<'ENDSSH'
```
gssbzn, line 33: can we apply a similar thing to what ego does and centralise these args given you use them twice

build/package/install_agent_spawn_host.sh
```diff
@@ -0,0 +1,53 @@
...
+        i) keyfile="$OPTARG";;      # SSH identity file
+        u) user="$OPTARG";;         # Username on the remote host
+        h) hostsfile="$OPTARG";;    # Output of Evergreen host.list
+        t) groupid="$OPTARG";;      # APIGroupId
+        v) apiKey="$OPTARG";;       # APIKey
+        *) exit 1;;
+    esac
+done
+
+hosts=$(cat << EOF | python - "$hostsfile"
```
gssbzn, line 16: [non blocking ]I'd rather not use python but we can look at this later on

build/package/install_agent_spawn_host.sh
```diff
@@ -0,0 +1,53 @@
...
+        sudo mkdir -p /data
+        sudo chown mongod:mongod /data
+
+        echo "Starting the agent"
+        sudo systemctl start mongodb-mms-automation-agent.service
+        exit
+ENDSSH
+    echo "Storing $host in src/github.com/mongodb/mongocli/e2e/cloud_manager/e2e.env"
+    sudo sed -i "s/\(hostname *= *\).*/\1$host/" src/github.com/mongodb/mongocli/e2e/cloud_manager/e2e.env
+done
```
gssbzn, line 53: [nit] missing new line (you can usually set your IDE/editor to add one for you on save)

build/package/install_agent_spawn_host.sh
```diff
@@ -0,0 +1,53 @@
+#!/bin/bash
+
+set -ex
```
gssbzn, line 3: [nit] this one is a bit more common on our scripts

```suggestion
set -Eeou pipefail
```

build/package/install_agent_spawn_host.sh
```diff
@@ -0,0 +1,53 @@
...
+    echo "Installing dependeces on $host"
+    ssh -i "$keyfile" -o ConnectTimeout=50 -o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no -tt "$user@$host" 'bash -s' <<'ENDSSH'
+        free -m
+        sudo apt-get install -y --no-install-recommends ca-certificates curl logrotate openssl snmp && exit
+ENDSSH
+    echo "Installing the automation agent on $host"
+    ssh -i "$keyfile" -o ConnectTimeout=50 -o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no -tt "$user@$host" ARG1="$groupid" ARG2="$apiKey" 'bash -s' <<'ENDSSH'
+        free -m
+        echo "Downloadind and extracting the automation agent"
+        curl -OL https://cloud-dev.mongodb.com/download/agent/automation/mongodb-mms-automation-agent-manager_10.15.0.6409-1_amd64.ubuntu1604.deb
```
gssbzn, line 36: [nit] two things here, let's get the agent from cloud (https://cloud.mongodb.com) and can we use latest here? as in 
`mongodb-mms-automation-agent-manager_latest`? I think 
`https://cloud.mongodb.com/download/agent/automation/mongodb-mms-automation-agent-manager_latest_amd64.ubuntu1604.deb` works
The main reason to use latest is os we don't have to go updating this test and the reason to use cloud is to use an stable release

build/package/install_agent_spawn_host.sh
andreaangiolillo: no, it is a leftover from previous attempts

e2e/cloud_manager/e2e.env
```diff
@@ -0,0 +1,7 @@
+# HOSTNAME contains the name of the host that is running the automation agent.
```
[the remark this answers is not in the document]
andreaangiolillo: I forgot to update the gitignore

build/package/install_agent_spawn_host.sh
andreaangiolillo: I tried to do it but I got  `No such file or directory` and it seems that it is looking for that file inside the ubuntu 16.04 spawn host.

build/package/install_agent_spawn_host.sh
andreaangiolillo: After making this change one cloud manager e2e test is failing and the agent is no longer display in the UI
```
Invalid config: You need to upgrade the automation agent before publishing other changes
[2020/05/28 17:52:37.759]     --- FAIL: TestCloudManagerDBUsers/Create (0.20s)
```

andreaangiolillo: I found the issue: since we are downloading the agent from **cloud.mongodb.com** instead of **cloud-dev.mongodb.com**, the property `mmsBaseUrl` was pointing to **cloud.mongodb.com**. Unfortunately, the following piece of code does not work if we have special characters.

```
sudo sed -i "s/\(mmsGroupId *= *\).*/\1$ARG1/" /etc/mongodb-mms/automation-agent.config
sudo sed -i "s/\(mmsApiKey *= *\).*/\1$ARG2/" /etc/mongodb-mms/automation-agent.config
```


**I fixed this issue by downloading the latest from cloud-dev**. I don't think that makes a difference here since the main point was to download always the latest version of the automation agent. Let me know if this solution is okay with you @gssbzn

build/package/install_agent_spawn_host.sh
andreaangiolillo: Could you explain why? I don't see any downside in using python.

gssbzn: a couple of reasons
1) we are introducing a dependency on python for a go project, I prefer to stay in language for tools
2) since this going to run mostly on evergreen and we aren't setting anything to have a python available this may break if evergreen decides to change python, which takes me back to point one of introducing a new dependency we need to maintain

This is working now so don't touch it but we may change it later to either be all bash or add some `go` to do it

CONTRIBUTING.md
```diff
@@ -40,8 +40,11 @@ The following is a short list of commands that can be run in the root of the pro
 - Run `make` see a list of available targets.
 - Run `make test` to run all unit tests.
 - Run `make lint` to validate against our linting rules.
-- Run `make e2e` will run end to end tests against an Atlas instance,
-  please make sure too have set `MCLI_*` variables pointing to that instance.
+- Run `export E2E_TAGS=e2e,atlas && make e2e-test` will run end to end tests against an Atlas instance,
```
gssbzn, line 43: [nit] this works and is shorter
```suggestion
- Run `E2E_TAGS=e2e,atlas make e2e-test` will run end to end tests against an Atlas instance,
```

CONTRIBUTING.md
```diff
@@ -40,8 +40,11 @@ The following is a short list of commands that can be run in the root of the pro
...
+  please make sure to have set `MCLI_*` variables pointing to that instance.
+- Run `export E2E_TAGS=e2e,cloudmanager && make e2e-test` will run end to end tests against an Cloud Manager instance.<br />
```
gssbzn, line 45: [nit] same
```suggestion
- Run `E2E_TAGS=e2e,cloudmanager make e2e-test` will run end to end tests against an Cloud Manager instance.<br />
```

CONTRIBUTING.md
```diff
@@ -40,8 +40,11 @@ The following is a short list of commands that can be run in the root of the pro
...
+  Please remember to update `the e2e/cloud_manager/e2e.env` with the name of the host that is running your automation agent and 
+  to set `MCLI_*` variables to point to your Cloud Manager instance.
```
gssbzn, line 47: not for this pr but for when you start using the `e2e.env` file, we should probably start using something like https://github.com/joho/godotenv end put all those env there instead of forcing users to export them

build/package/install-agent-spawn-host.sh
```diff
@@ -0,0 +1,70 @@
...
+for host in $hosts; do
+  set +e
+  echo "Installing dependeces on $host"
+  ssh "${ssh_opts[@]}" -tt "$user@$host" 'bash -s' <<'ENDSSH'
+        sudo apt-get install -y --no-install-recommends ca-certificates curl logrotate openssl snmp && exit
+ENDSSH
+  echo "Installing the automation agent on $host"
+  ssh "${ssh_opts[@]}" -tt "$user@$host" ARG1="$groupid" ARG2="$apiKey" 'bash -s' <<'ENDSSH'
+        echo "Downloadind and extracting the automation agent"
+        curl -OL https://cloud-dev.mongodb.com/download/agent/automation/mongodb-mms-automation-agent-manager_latest_amd64.ubuntu1604.deb
```
gssbzn, line 52: ok to use ` https://cloud-dev.mongodb.com` but we have an env variable for the base url don't we? could we use that one here as well so they match?

build/package/install-agent-spawn-host.sh
```diff
@@ -0,0 +1,70 @@
+#!/bin/bash
+
+set -ex
```
gssbzn, line 3:
```suggestion
set -euo pipefail
```

build/package/ssh-ready.sh
```diff
@@ -0,0 +1,46 @@
+#!/bin/bash
+set -ex
```
gssbzn, line 2:
```suggestion
set -euo pipefail
```

e2e/cloud_manager/e2e.env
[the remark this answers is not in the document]
gssbzn: since you committed the file you may need to delete it after adding it to the gitignore

build/package/install_agent_spawn_host.sh
andreaangiolillo: I googled this and It seems that running a local script in the ssh host is not so easy when there are arguments to pass.

build/package/install-agent-spawn-host.sh
andreaangiolillo: As I said here: https://github.com/mongodb/mongocli/pull/179#discussion_r432089778, the piece of code that changes properties inside the `mongodb-mms/automation-agent.config` does not work if the string contains special characters (in this case it would be https://.....).
I have tried only a few attempts to make that piece of code working also with special characters. Is there any difference in downloading the latest agent from cloud-dev instead of cloud? It might not be worthed the try if is the same

andreaangiolillo: Sorry for the previous comment. I misunderstood it.

CONTRIBUTING.md
andreaangiolillo: I have created this task for that [CLOUDP-64365](https://jira.mongodb.org/browse/CLOUDP-64365)

e2e/cloud_manager/e2e.env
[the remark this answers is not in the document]
gssbzn: this file is still here

andreaangiolillo:
I have renamed the e2e.env to e2e.env.example.

PR #182, opened 2020-05-25
CLOUDP-62837: Output test results in junit format for evergreen

gssbzn:
## Proposed changes

_Jira ticket:_ [CLOUDP-62837](https://jira.mongodb.org/browse/CLOUDP-62837)

## Checklist

## Further comments

We had a function to run golang ci but never enable that so that's a bonus thing here

### DIFF
--- a/.evergreen.yml
+++ b/.evergreen.yml
@@ -362,10 +362,12 @@ post:
 
 tasks:
   - name: compile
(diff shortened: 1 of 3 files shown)


gssbzn:
changed here to add tags and configure evg to run based on tags instead of name of task (less configuration)

PR #183, opened 2020-05-25
Add junit reports to e2e

gssbzn:
## Checklist

### DIFF
--- a/.evergreen.yml
+++ b/.evergreen.yml
@@ -964,9 +964,7 @@ buildvariants:
       go_root: "/opt/golang/go1.14"
       go_bin: "/opt/golang/go1.14/bin"
     tasks:
-      - name: compile
-      - name: unit_test
-      - name: lint
+      - name: .code_health
   - name: e2e
     display_name: "E2E"
     run_on:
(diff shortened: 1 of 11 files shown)


.evergreen.yml
```diff
@@ -382,15 +386,23 @@ tasks:
...
       - func: "atlas e2e"
   - name: cloud_manager_e2e
     tags: ["e2e"]
+    depends_on:
+      - name: compile
+        variant: "code_health"
+        patch_optional: true
     commands:
       - func: "clone"
-      - func: "build"
```
gssbzn, line 393: `make e2e` internally builds so no need to duplicate that

.evergreen.yml
```diff
@@ -382,15 +386,23 @@ tasks:
       - func: "lint"
   - name: atlas_e2e
     tags: ["e2e"]
+    depends_on:
```
gssbzn, line 389: since this depends on mcli building let's not trigger this task until we are sure we can build

.evergreen.yml
```diff
@@ -79,8 +79,10 @@ functions:
           MCLI_PRIVATE_API_KEY: ${atlas_private_api_key}
           MCLI_PUBLIC_API_KEY: ${atlas_public_api_key}
           MCLI_SERVICE: cloud
+          E2E_CMD: gotestsum --junitfile unit-tests.xml --
+          E2E_TAGS: e2e,atlas
           <<: *go_options
-        command: make e2e-test-atlas
+        command: make e2e-test
   "cloud manager e2e":
```
gssbzn, line 86: this can probably be just one function with different arguments but I'll change that later

PR #187, opened 2020-05-25
CLOUDP-62054: replace mock generation in make with go generate

gssbzn:
## Proposed changes

_Jira ticket:_ [CLOUDP-62054](https://jira.mongodb.org/browse/CLOUDP-62054)

Closes #[issue number]

## Checklist


CONTRIBUTING.md
```diff
@@ -48,8 +48,8 @@ We provide a git pre-commit hook to format and check the code, to install it run
 
 ### Generating Mocks
 
-We use [mockgen](https://github.com/golang/mock) to handle mocking in our unit tests
-If you need a new mock please add a reference on the [Make](Makefile) file and run `make gen-mocks`
+We use [mockgen](https://github.com/golang/mock) to handle mocking in our unit tests.
+If you need a new mock please update or add the `//go:generate` instruction to the appropriate file
```
andreaangiolillo, line 52: 💯

internal/store/deployments.go
```diff
@@ -30,6 +33,36 @@ type HostDescriber interface {
 	Host(string, string) (*opsmngr.Host, error)
 }
 
+type HostDatabaseLister interface {
```
andreaangiolillo, line 36: I don't think this should be part of this PR. You don't have to change it, it is just my opinion since this doesn't seem related to this scope of the PR.

internal/store/measurements.go
```diff
@@ -23,10 +23,26 @@ import (
 	"go.mongodb.org/ops-manager/opsmngr"
 )
 
+//go:generate mockgen -destination=../mocks/mock_measurements.go -package=mocks github.com/mongodb/mongocli/internal/store HostMeasurementLister,HostDiskMeasurementsLister
+
+type HostMeasurementLister interface {
```
andreaangiolillo, line 28: Same as above

internal/store/deployments.go
gssbzn: the reason I did it was to group the related methods and thus reducing the files to tag with go generate, if you check they all  use the same `opsmngr. Deployments` module, this changed on the client but we never updated the file grouping so here I'm doing that, but to your point yes, they could've been a separate PR

PR #193, opened 2020-05-29
CLOUDP-64120: implementing e2e test for logs - atlas

andreaangiolillo:
## Proposed changes

_Jira ticket:_ [CLOUDP-64120](https://jira.mongodb.org/browse/CLOUDP-64120)

## Checklist


e2e/atlas/logs_test.go
```diff
@@ -0,0 +1,188 @@
...
+	if err != nil {
+		t.Fatalf("unexpected error: %v", err)
+	}
+
+	atlasEntity := "atlas"
+	logsEntity := "logs"
+	clusterName := ""
+
+	// Deploy a cluster only if there is not one available
+	if !atlas.ExistCluster(cliPath) {
```
andreaangiolillo, line 47: I thought this was a good idea: deploying a new cluster only if there are no available but now metric is failing probably because of it. It does not create a new cluster because one cluster is running but then the test fails because the cluster has been deleted.
Can we just deploy one cluster in atlas that we use for testing metrics, logs etc and we do not remove it? This would simplify the code and reduce the time to run the test. @gssbzn

gssbzn: > Can we just deploy one cluster in atlas that we use for testing metrics
short answer, no, they get removed after 24h
This goes back to to a point I made before, probably you're better off running all this tests depending on the cluster create test
Or both logs and metrics will need to create their own cluster to watch for
if you decide to go for logs and metrics depending on their own cluster I'd advice to start separating these tests in different evg tasks so they can run on parallel

e2e/atlas/helper.go
```diff
@@ -0,0 +1,117 @@
...
+		return "", fmt.Errorf("got=%#v\nwant=%#v", 0, "len(processes) > 0")
+	}
+
+	// The first element may not be the created cluster but that is fine since
+	// we just need one cluster up and running
+	return processes[0].Hostname + ":" + strconv.Itoa(processes[0].Port), nil
+}
+
+// ExistCluster returns true if there is at least a cluster is deployed, false otherwise
+func ExistCluster(cliPath string) bool {
```
gssbzn, line 49: [nit] since you don't search for a particular cluster
```suggestion
func AnyCluster(cliPath string) bool {
```

e2e/atlas/helper.go
```diff
@@ -0,0 +1,117 @@
...
+
+	if err != nil {
+		return false
+	}
+
+	if len(clusters) > 0 {
+		return true
+	}
+
+	return false
```
gssbzn, line 72: [nit]
```suggestion
	return len(clusters) > 0
```

e2e/atlas/logs_test.go
```diff
@@ -0,0 +1,188 @@
...
```
gssbzn, line 47: if you plan to run on parallel you need each test to create their own cluster or that cluster can be deleted in the middle of the test by the other test that created it

e2e/atlas/helper.go
```diff
@@ -0,0 +1,117 @@
+package atlas
```
gssbzn, line 1: I like the idea of this file, one general comment I would make `cliPath` a global var of the module so you don't have to pass it to each method

gssbzn: another general thing here let's make this package `atlas_test`
so we can keep the methods private since I don't expect us to use them outside these tests

build/ci/evergreen.yml
```diff
@@ -78,6 +78,24 @@ functions:
           E2E_TAGS: e2e,atlas
           <<: *go_options
         command: make e2e-test
+  "atlas logs e2e":
```
gssbzn, line 81: [q] do you want to do the same for metrics and clusters?
we'll end up with 4 e2e tasks for atlas on evg, one general, one for clusters, one for metrics and one for logs all run on parallel

e2e/atlas/helper.go
andreaangiolillo: It seems that in order to change the package name I have to move the helper inside the folder `atlas_test`.  Also after doing that and making the methods private I am getting a lint error which says that my methods are not being used. Should be private method visibility within the package? @gssbzn

gssbzn: If you see the other tests in this folder are in `atlas_test` package, you only need to change the name of this file to be `helper_test` as well

build/ci/evergreen.yml
```diff
@@ -78,6 +78,78 @@ functions:
...
+          MCLI_ORG_ID: ${atlas_org_id}
+          MCLI_PROJECT_ID: ${atlas_project_id}
+          MCLI_PRIVATE_API_KEY: ${atlas_private_api_key}
+          MCLI_PUBLIC_API_KEY: ${atlas_public_api_key}
+          MCLI_SERVICE: cloud
+          TEST_CMD: gotestsum --junitfile e2e-tests.xml --
+          E2E_TAGS: e2e,atlas,clusters
+          <<: *go_options
+        command: make e2e-test
+  "iam e2e":
```
gssbzn, line 135: looks like we are no longer testing this for cloud manager, I'm ok to split it out to a separate task but let's add one for Cloud manager as well

e2e/atlas/clusters_test.go
```diff
@@ -30,11 +29,7 @@ import (
 )
 
 func TestAtlasClusters(t *testing.T) {
-	cliPath, err := filepath.Abs("../../bin/mongocli")
-	if err != nil {
-		t.Fatalf("unexpected error: %v", err)
-	}
-	_, err = os.Stat(cliPath)
+	_, err := os.Stat(cliPath)
```
gssbzn, line 32: [nit] this is a test to make sure the binary exists in the path, you could move this to the same `init` function where you do many of the cli path discovery

PR #201, opened 2020-05-30
Clarify build instructions

p-mongo:
This is an attempt to clarify the situation for https://github.com/mongodb/mongocli/issues/195 by separately listing build and install commands, and specifying where the output of each is placed.


gssbzn:
Thanks for adding the clarifications

PR #220, opened 2020-06-24
CLOUDP-65942: Add auto complete to brew installer

gssbzn:
## Proposed changes

_Jira ticket:_ [CLOUDP-65942](https://jira.mongodb.org/browse/CLOUDP-65942)

## Checklist

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -50,11 +50,14 @@ brews:
     owner: mongodb
     name: homebrew-brew
   folder: Formula
-  skip_upload: false
+  skip_upload: auto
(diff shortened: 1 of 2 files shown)


.goreleaser.yml
```diff
@@ -50,11 +50,14 @@ brews:
     owner: mongodb
     name: homebrew-brew
   folder: Formula
-  skip_upload: false
+  skip_upload: auto
```
colm-quinn, line 53: Just checking, so this means it will skip uploading rc's but will upload the tags?

gssbzn: Rcs, betas, alpha, and such, we disabled this when we were doing our rc, so going back to not publishing unstable

PR #221, opened 2020-06-25
CLOUDP-64975: Add a message with link to get keys at the top of the config command

gssbzn:
## Proposed changes

## Checklist

## Further comments

Screenshot 2020-06-25 at 11 36 26

### DIFF
--- a/internal/cli/config/config.go
+++ b/internal/cli/config/config.go
@@ -18,7 +18,6 @@ import (
 	"fmt"
 
 	"github.com/AlecAivazis/survey/v2"
-	"github.com/mongodb/mongocli/internal/cli"
 	"github.com/mongodb/mongocli/internal/config"
 	"github.com/mongodb/mongocli/internal/description"
 	"github.com/mongodb/mongocli/internal/flag"
@@ -27,13 +26,21 @@ import (
 	"github.com/spf13/cobra"
 )
 
+const (
+	atlasAPIHelp  = "Please provide your API keys, to get new keys follow: https://docs.atlas.mongodb.com/configure-api-access/"
+	omAPIHelp     = "Please provide your API keys, to get new keys follow: https://docs.opsmanager.mongodb.com/current/tutorial/configure-public-api-access/"
+	omBaseURLHelp = "FQDN and port number of the Ops Manager Application."
+	projectHelp   = "ID of an existing project, your API keys should have access to it. Leaving this blank means using --projectId for every command that requires it."
+	orgHelp       = "ID of an existing organization, your API keys should have access to it. Leaving this blank means using --orgId for every command that requires it."
+)
+
 type configOpts struct {
-	cli.GlobalOpts
 	Service       string
 	PublicAPIKey  string
 	PrivateAPIKey string
 	OpsManagerURL string
 	ProjectID     string
+	OrgID         string
 }
 
 func (opts *configOpts) IsCloud() bool {
@@ -67,13 +74,47 @@ func (opts *configOpts) Save() error {
 }
 
 func (opts *configOpts) Run() error {
-	helpLink := "https://docs.atlas.mongodb.com/configure-api-access/"
+	fmt.Printf(`You are setting up %s
 
-	if opts.IsOpsManager() {
-		helpLink = "https://docs.opsmanager.mongodb.com/current/tutorial/configure-public-api-access/"
+All values are optional and you can use environment variables (MCLI_*) instead.
+
+Enter [?] on any option to get help.
+
+`, config.ToolName)
+
+	q := opts.accessQuestions()
+	if err := survey.Ask(q, opts); err != nil {
+		return err
 	}
 
-	defaultQuestions := []*survey.Question{
+	q = opts.defaultQuestions()
+	if err := survey.Ask(q, opts); err != nil {
+		return err
+	}
+
+	if err := opts.Save(); err != nil {
+		return err
+	}
+
+	fmt.Printf("\n%s is all set now\n", config.ToolName)
+	fmt.Printf("You can use [%s config set] to change more settings. \n", config.ToolName)
+	if config.Name() != config.DefaultProfile {
+		fmt.Printf("To use this profile remember to set the flag [-%s %s]\n", flag.ProfileShort, config.Name())
+	}
+	return nil
+}
+
+func (opts *configOpts) apiKeyHelp() string {
+	if opts.IsOpsManager() {
+		return omAPIHelp
+	}
+	return atlasAPIHelp
+}
+
+func (opts *configOpts) accessQuestions() []*survey.Question {
+	helpLink := opts.apiKeyHelp()
+
+	q := []*survey.Question{
 		{
 			Name: "publicAPIKey",
 			Prompt: &survey.Input{
@@ -89,17 +130,7 @@ func (opts *configOpts) Run() error {
 				Help:    helpLink,
 			},
 		},
-		{
-			Name: "projectId",
-			Prompt: &survey.Input{
-				Message: "Default Project ID [optional]:",
-				Help:    "This is the ID of an existing project your API keys have access to, you can leave this blank and specify it on every command with --projectId",
-				Default: config.ProjectID(),
-			},
-			Validate: validate.OptionalObjectID,
-		},
 	}
-
 	if opts.IsOpsManager() {
 		opsManagerQuestions := []*survey.Question{
 			{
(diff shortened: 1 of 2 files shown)


viviandiancheng:
Can we change the [? for help] to [enter ? for help], which feels more like an action guide to me and more consistent with the language above.
Also, what is the behavior after users entered "?" and then complete the command?
I thought the blue help notice would disappear like the public API key. But in private API key, after I entered the correct key, there still one line of blue help showing on the screen, which is unnecessary.  
image

internal/cli/config/config.go
```diff
@@ -27,13 +26,21 @@ import (
 	"github.com/spf13/cobra"
 )
 
+const (
+	atlasAPIHelp  = "Please provide your API keys, to get new keys follow: https://docs.atlas.mongodb.com/configure-api-access/"
```
melissamahoney-mongodb, line 30:
```suggestion
	atlasAPIHelp  = "Please provide your API keys. To create new keys, see the documentation: https://docs.atlas.mongodb.com/configure-api-access/"
```

internal/cli/config/config.go
```diff
@@ -27,13 +26,21 @@ import (
...
+	omAPIHelp     = "Please provide your API keys, to get new keys follow: https://docs.opsmanager.mongodb.com/current/tutorial/configure-public-api-access/"
```
melissamahoney-mongodb, line 31:
```suggestion
	omAPIHelp     = "Please provide your API keys. To create new keys, see the documentation: https://docs.opsmanager.mongodb.com/current/tutorial/configure-public-api-access/"
```

internal/cli/config/config.go
```diff
@@ -27,13 +26,21 @@ import (
...
+	omBaseURLHelp = "FQDN and port number of the Ops Manager Application."
+	projectHelp   = "ID of an existing project, your API keys should have access to it. Leaving this blank means using --projectId for every command that requires it."
```
melissamahoney-mongodb, line 33:
```suggestion
	projectHelp   = "ID of an existing project that your API keys have access to. If you don't enter an ID, you must use --projectId for every command that requires it."
```

internal/cli/config/config.go
```diff
@@ -27,13 +26,21 @@ import (
...
+	orgHelp       = "ID of an existing organization, your API keys should have access to it. Leaving this blank means using --orgId for every command that requires it."
```
melissamahoney-mongodb, line 34:
```suggestion
	orgHelp       = "ID of an existing organization that your API keys have access to. If you don't enter an ID, you must use --orgId for every command that requires it."
```

internal/cli/config/config.go
```diff
@@ -67,13 +74,47 @@ func (opts *configOpts) Save() error {
 }
 
 func (opts *configOpts) Run() error {
-	helpLink := "https://docs.atlas.mongodb.com/configure-api-access/"
+	fmt.Printf(`You are setting up %s
```
melissamahoney-mongodb, line 77: Should we make this more obvious that we are setting up a profile?
```suggestion
	fmt.Printf(`You are configuring a profile for %s.
```

internal/cli/config/config.go
```diff
@@ -67,13 +74,47 @@ func (opts *configOpts) Save() error {
...
+	q = opts.defaultQuestions()
+	if err := survey.Ask(q, opts); err != nil {
+		return err
+	}
+
+	if err := opts.Save(); err != nil {
+		return err
+	}
+
+	fmt.Printf("\n%s is all set now\n", config.ToolName)
```
melissamahoney-mongodb, line 99:
```suggestion
	fmt.Printf("\nYour profile is now configured.\n", config.ToolName)
```

internal/cli/config/config.go
```diff
@@ -67,13 +74,47 @@ func (opts *configOpts) Save() error {
...
+	fmt.Printf("You can use [%s config set] to change more settings. \n", config.ToolName)
```
melissamahoney-mongodb, line 100:
```suggestion
	fmt.Printf("You can use [%s config set] to change these settings at a later time.\n", config.ToolName)
```

internal/cli/config/config.go
```diff
@@ -67,13 +74,47 @@ func (opts *configOpts) Save() error {
...
+	if config.Name() != config.DefaultProfile {
+		fmt.Printf("To use this profile remember to set the flag [-%s %s]\n", flag.ProfileShort, config.Name())
+	}
+	return nil
+}
```
melissamahoney-mongodb, line 105:
```suggestion
	if config.Name() != config.DefaultProfile {
		fmt.Printf("To use this profile, you must set the flag [-%s %s] for every command.\n", flag.ProfileShort, config.Name())
	}
	return nil
}
```

Also I think we should move this up under the first sentence, after line 99. So that the order makes a little more sense:
(To use this profile, you must set the flag [-%s %s] for every command.)
You can use [%s config set] to change these settings at a later time.

internal/cli/config/config.go
```diff
@@ -67,13 +74,47 @@ func (opts *configOpts) Save() error {
...
+	}
+
+	if err := opts.Save(); err != nil {
+		return err
+	}
+
+	fmt.Printf("\n%s is all set now\n", config.ToolName)
+	fmt.Printf("You can use [%s config set] to change more settings. \n", config.ToolName)
+	if config.Name() != config.DefaultProfile {
+		fmt.Printf("To use this profile remember to set the flag [-%s %s]\n", flag.ProfileShort, config.Name())
```
melissamahoney-mongodb, line 102: Providing them with the flag is a really nice touch! :+1:

internal/description/description.go
```diff
@@ -75,8 +75,8 @@ const (
 	CloudManager             = "Cloud Manager operations."
 	ShutdownCluster          = "Shutdown a cluster."
 	StartUpCluster           = "Start up a cluster."
-	ConfigDescription        = "Configure a profile. This let you store access settings to your cloud."
-	ConfigSetDescription     = "Configure specific properties of the profile."
+	ConfigDescription        = "Configure a profile. This let you store default settings."
```
melissamahoney-mongodb, line 78: Are we sure we want to use "default settings" here? Will that cause confusion between settings stored in a default profile vs other profile? Maybe something like this:
```suggestion
	ConfigDescription        = "Configure a profile to store access settings for your MongoDB deployment."
```

internal/description/description.go
```diff
@@ -118,8 +118,12 @@ The name of the log file must be one of: mongodb.gz, mongos.gz, mongodb-audit-lo
 	DBUsersLong = `The dbusers command retrieves, creates and modifies the MongoDB database users in your cluster.
 Each user has a set of roles that provide access to the project’s databases. 
 A user’s roles apply to all the clusters in the project.`
-	ConfigLongDescription = `Setting API keys is optional and env variables can be used instead.
-Leaving any option empty won't override existing stored values.`
+	ConfigLongDescription = `Configure default values in a user profile.
```
melissamahoney-mongodb, line 121: Same question around "default" causing confusion? Maybe:
```suggestion
	ConfigLongDescription = `Configure settings in a user profile.
```

internal/description/description.go
```diff
@@ -118,8 +118,12 @@ The name of the log file must be one of: mongodb.gz, mongos.gz, mongodb-audit-lo
...
+All settings are optional, and can be set individually, to find out how run:
```
melissamahoney-mongodb, line 122:
```suggestion
All settings are optional. You can specify settings individually by running: 
```

internal/description/description.go
```diff
@@ -118,8 +118,12 @@ The name of the log file must be one of: mongodb.gz, mongos.gz, mongodb-audit-lo
...
 Each user has a set of roles that provide access to the project’s databases. 
 A user’s roles apply to all the clusters in the project.`
-	ConfigLongDescription = `Setting API keys is optional and env variables can be used instead.
-Leaving any option empty won't override existing stored values.`
+	ConfigLongDescription = `Configure default values in a user profile.
+All settings are optional, and can be set individually, to find out how run: 
+  $ mongocli config set --help 
+
+You can always use environment variables (MCLI_*) when running the tool,
+To find out more check, https://docs.mongodb.com/mongocli/stable/configure/environment-variables/.`
```
melissamahoney-mongodb, line 126:
```suggestion
You can also use environment variables (MCLI_*) when running the tool.
To find out more, see the documentation: https://docs.mongodb.com/mongocli/stable/configure/environment-variables/.`
```

gssbzn:
@viviandiancheng the library we use for those prompts looks like doesn't support changing the text to ask for help, also it seems to be a bug with the password type where they won't clean after you input something, probably to avoid removing the `****` but I'm looking at their code to see if I can find a fix a submit it to them

gssbzn:
@melissamahoney-mongodb I think I addressed all you comments thank you for all of them

PR #223, opened 2020-06-29
CLOUDP-64978: List projects by name on config

### LINKED ISSUE #200: List available projects by name instead of asking for an ID
I am following getting started instructions [here](https://docs.mongodb.com/mongocli/master/quick-start/#set-the-id-of-the-project-in-your-profile). The third step is:

> Set the ID of the project in your profile.

> When the command prompts you, enter your project ID.

I do not get this prompt - I am only prompted for public and private API keys.

I tried with and without `~/.config/mongocli.toml               ` existing, same behavior.

gssbzn:
## Proposed changes

_Jira ticket:_ [CLOUDP-64978](https://jira.mongodb.org/browse/CLOUDP-64978)
_Jira ticket:_ [CLOUDP-65614](https://jira.mongodb.org/browse/CLOUDP-65614)

Closes #200

## Checklist

## Further comments

If there's an error we fallback to the old ways

Screenshot 2020-07-07 at 14 03 39

New select
Screenshot 2020-07-07 at 14 04 05

### DIFF
--- a/internal/cli/config/config.go
+++ b/internal/cli/config/config.go
@@ -21,18 +21,22 @@ import (
 	"github.com/mongodb/mongocli/internal/config"
 	"github.com/mongodb/mongocli/internal/description"
 	"github.com/mongodb/mongocli/internal/flag"
+	"github.com/mongodb/mongocli/internal/store"
 	"github.com/mongodb/mongocli/internal/usage"
 	"github.com/mongodb/mongocli/internal/validate"
 	"github.com/spf13/cobra"
+	atlas "go.mongodb.org/atlas/mongodbatlas"
+	"go.mongodb.org/ops-manager/opsmngr"
 )
 
 const (
-	atlasAPIHelp  = "Please provide your API keys. To create new keys, see the documentation: https://docs.atlas.mongodb.com/configure-api-access/"
-	omAPIHelp     = "Please provide your API keys. To create new keys, see the documentation: https://docs.opsmanager.mongodb.com/current/tutorial/configure-public-api-access/"
-	omBaseURLHelp = "FQDN and port number of the Ops Manager Application."
-	projectHelp   = "ID of an existing project that your API keys have access to. If you don't enter an ID, you must use --projectId for every command that requires it."
-	orgHelp       = "ID of an existing organization that your API keys have access to. If you don't enter an ID, you must use --orgId for every command that requires it."
+	nameIDFormat = "%s (%s)"
 )
+
+type ProjectOrgsLister interface {
+	Projects(*atlas.ListOptions) (interface{}, error)
+	Organizations(*atlas.ListOptions) (*atlas.Organizations, error)
+}
 
 type configOpts struct {
 	Service       string
@@ -41,6 +45,13 @@ type configOpts struct {
 	OpsManagerURL string
 	ProjectID     string
 	OrgID         string
+	store         ProjectOrgsLister
+}
+
+func (opts *configOpts) initStore() error {
+	var err error
+	opts.store, err = store.New(config.Default())
+	return err
 }
 
 func (opts *configOpts) IsCloud() bool {
@@ -51,11 +62,7 @@ func (opts *configOpts) IsOpsManager() bool {
 	return opts.Service == config.OpsManagerService
 }
 
-func (opts *configOpts) IsCloudManager() bool {
-	return opts.Service == config.CloudManagerService
-}
-
-func (opts *configOpts) Save() error {
+func (opts *configOpts) SetUpAccess() {
 	config.SetService(opts.Service)
 	if opts.PublicAPIKey != "" {
 		config.SetPublicAPIKey(opts.PublicAPIKey)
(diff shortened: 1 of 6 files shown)


internal/prompt/prompt.go
```diff
@@ -21,17 +21,17 @@ import (
 )
 
 // NewDeleteConfirm creates a prompt to confirm if the entry should be deleted
-func NewDeleteConfirm(entry string) *survey.Confirm {
+func NewDeleteConfirm(entry string) survey.Prompt {
```
gssbzn, line 24: let's use the interface instead

internal/cli/config/config.go
```diff
@@ -51,11 +62,7 @@ func (opts *configOpts) IsOpsManager() bool {
 	return opts.Service == config.OpsManagerService
 }
 
-func (opts *configOpts) IsCloudManager() bool {
-	return opts.Service == config.CloudManagerService
-}
```
colm-quinn, line 56: [q] Was this just not being used or now exists somewhere else?

gssbzn: not being used

internal/cli/config/config.go
```diff
@@ -107,72 +140,82 @@ Enter [?] on any option to get help.
...
-					Message: "URL to Access Ops Manager:",
-					Default: config.OpsManagerURL(),
-					Help:    omBaseURLHelp,
-				},
-				Validate: validate.URL,
-			},
+	return pMap, pSlice, nil
+}
+
+// askOrg will try to to construct a select based on fetched organizations.
```
colm-quinn, line 181: [np] `try to to` -> `try to`

internal/cli/config/config.go
```diff
@@ -204,3 +247,27 @@ func Builder() *cobra.Command {
...
+func omProjects(projects []*opsmngr.Project) (pMap map[string]string, pSlice []string) {
+	pMap = make(map[string]string, len(projects))
+	pSlice = make([]string, len(projects))
+	for i, p := range projects {
+		d := fmt.Sprintf(nameIDFormat, p.Name, p.ID)
+		pMap[d] = p.ID
+		pSlice[i] = d
+	}
+	return pMap, pSlice
+}
```
colm-quinn, line 273: [q] Is it possible to make a common helper here? The code looks almost identical except for `*atlas.Project` vs `*opsmngr.Project` in the signature.

gssbzn: sadly I don't know how, golang does not have great support for generic slices, so the for would need to be duplicated given that I would not be able to do `p.ID` for two different types even when they gave the same properties (supposedly [they are working on this](https://blog.golang.org/why-generics))

`pMap = make(map[string]string, len(projects))` and such are also troublesome  since I can't do len of the interface unless I cast it, which just makes me extract this from this method but duplicated somewhere else 

so is just a question of where do we duplicate :(

colm-quinn: Fair enough, no language is perfect. Thanks for the response for my own learning.

internal/cli/config/config.go
```diff
@@ -107,72 +140,82 @@ Enter [?] on any option to get help.
...
+// If it fails or there are no projects to show we fallback to ask for project by ID
+func (opts *configOpts) askProject() (string, error) {
+	pMap, pSlice, err := opts.projects()
+	var projectID string
+	if err != nil || len(pSlice) == 0 {
+		prompt := newProjectIDInput()
+		if err := survey.AskOne(prompt, &projectID, survey.WithValidator(validate.OptionalObjectID)); err != nil {
+			return projectID, err
+		}
+		return projectID, nil
```
robcarlan-mlab, line 153: [np] if is unnecessary here

internal/cli/config/config.go
```diff
@@ -107,72 +140,82 @@ Enter [?] on any option to get help.
...
+func (opts *configOpts) askOrg() (string, error) {
+	oMap, oSlice, err := opts.orgs()
+	var orgID string
+	if err != nil || len(oSlice) == 0 {
+		prompt := newOrgIDInput()
+		if err := survey.AskOne(prompt, &orgID, survey.WithValidator(validate.OptionalObjectID)); err != nil {
+			return orgID, err
 		}
-		q = append(opsManagerQuestions, q...)
+		return orgID, nil
```
robcarlan-mlab, line 191: [np] and here

internal/cli/config/prompt.go
```diff
@@ -0,0 +1,105 @@
...
+
+func newOrgIDInput() survey.Prompt {
+	return &survey.Input{
+		Message: "Default Org ID:",
+		Help:    orgHelp,
+		Default: config.OrgID(),
+	}
+}
+
+func newOrgSelect(a []string) survey.Prompt {
```
robcarlan-mlab, line 33: [np] `a` -> `orgs` or `options`?

internal/cli/config/config.go
```diff
@@ -85,17 +96,39 @@ Enter [?] on any option to get help.
...
-	if err := survey.Ask(q, opts); err != nil {
+	if err := opts.initStore(); err != nil {
 		return err
 	}
 
-	if err := opts.Save(); err != nil {
+	if config.IsAccessSet() {
+		var err error
+
+		opts.OrgID, err = opts.askOrg()
```
robcarlan-mlab, line 112: [np] if we are setting a property of `opts` should we just do this in the function call to begin with?

gssbzn: fair point, let me check how this looks

internal/cli/config/prompt.go
gssbzn: I'm a lazy bastard who embraced the go philosophy of single letter variables, usually slices are just `a` or `s` but I'll change this to options

robcarlan-mlab: :D

gssbzn:
Ignoring drone since evergreen passed

PR #224, opened 2020-06-30
CLOUDP-66130: add config describe command

robcarlan-mlab:
## Proposed changes

_Jira ticket:_ [CLOUDP-66130](https://jira.mongodb.org/browse/CLOUDP-66130)

## Checklist

## Further comments

```
> ./bin/mongocli config describe om
public_api_key = redacted
service = ops-manager
ops_manager_url = http://om.com/
private_api_key = redacted
project_id = 5efa65463e8e86592c4ca608
```

```
> ./bin/mongocli config describe om123
There is no config with name 'om123'
```


internal/cli/config/config_describe.go
```diff
@@ -0,0 +1,59 @@
...
+	"fmt"
+
+	"github.com/mongodb/mongocli/internal/config"
+	"github.com/mongodb/mongocli/internal/description"
+	"github.com/spf13/cobra"
+)
+
+type ListOpts struct {
+	name  string
+	store config.Config
```
gssbzn, line 27: this seems to be unused, you could move to use this and add tests to your command but if that gets to complex feel free to just remove this

internal/config/profile.go
```diff
@@ -225,18 +225,48 @@ func (p *profile) SetOrgID(v string) {
 	p.Set(orgID, v)
 }
 
+// GetConfigDescription returns a map describing the configuration
+func GetConfigDescription() (map[string]string, error) {
+	if err := p.Load(false); err != nil {
+		return nil, err
+	}
```
gssbzn, line 232: you can avoid reloading the config (loading of the config happens automatically on [cli run](https://github.com/mongodb/mongocli/blob/e9b2e2b338b5577cfbe7231a08a7bd96ef3e2fb1/cmd/mongocli/mongocli.go#L104-L108))

and do `viper.GetStringMapString(profileName)` this will also ignore env variables as it just reads from the file
to ignore api keys then you can do something like
```go
if _, ok := settings[privateAPIKey]; ok {
    settings[privateAPIKey] = "redacted"
}

File: internal/cli/config/config_describe.go [hunk 3]
```diff
@@ -0,0 +1,59 @@
...
+	for k, v := range configDescription {
+		fmt.Printf("%v = %v\n", k, v)
+	}
+
+	return nil
+}
+
+func DescribeBuilder() *cobra.Command {
+	opts := &ListOpts{
+		store: config.Default(),
```
gssbzn, line 46: you don't read from the store but from global so similar to above you can move to use the store (good for testing) or remove it

.gitignore
```diff
@@ -7,6 +7,15 @@
 bin/**
 dist/**
 
+# IDE specific
```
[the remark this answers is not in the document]
robcarlan-mlab: It's subjective, but I like having these general files in the projects gitignore as it can make it easier for people to contribute without having to figure out a global gitignore if they haven't already

internal/config/profile.go
robcarlan-mlab: Yeah that will probably be better, as it avoids accidentally changing the config

internal/config/profile.go
```diff
@@ -225,18 +225,38 @@ func (p *profile) SetOrgID(v string) {
 	p.Set(orgID, v)
 }
 
+// GetConfigDescription returns a map describing the configuration
+func GetConfigDescription(name string) (map[string]string, error) {
```
gssbzn, line 229: I think is safe now to remove the error form the return
```suggestion
func GetConfigDescription(name string) map[string]string {
```

internal/cli/config/describe.go
```diff
@@ -0,0 +1,56 @@
...
+
+type ListOpts struct {
+	name string
+}
+
+func (opts *ListOpts) Run() error {
+	configDescription := config.GetConfigDescription(opts.name)
+
+	if len(configDescription) == 0 {
+		fmt.Printf("There is no config with name '%v'\n", opts.name)
```
gssbzn, line 33: [non-blocking] 
```suggestion
		return fmt.Errorf("no profile with name '%s'\n", opts.name)
```
(errors should not start with caps)

internal/cli/config/describe.go
```diff
@@ -0,0 +1,56 @@
...
+	}
+
+	for k, v := range configDescription {
+		fmt.Printf("%v = %v\n", k, v)
```
gssbzn, line 37: [nit]
```suggestion
		fmt.Printf("%s = %s\n", k, v)
```

internal/config/profile.go
```diff
@@ -225,18 +225,38 @@ func (p *profile) SetOrgID(v string) {
...
 	viper.SetConfigName(ToolName)
 	viper.SetConfigPermissions(0600)
 	viper.AddConfigPath(p.configDir)
 
-	viper.SetEnvPrefix(EnvPrefix)
+	if readEnvironmentVars {
+		viper.SetEnvPrefix(EnvPrefix)
+		viper.AutomaticEnv()
+	}
+
```
gssbzn, line 257: could we rollback this? feels like this may be the reason the e2e tests are failing

internal/cli/config/describe.go
```diff
@@ -0,0 +1,56 @@
...
+
+type ListOpts struct {
+	name string
+}
+
+func (opts *ListOpts) Run() error {
+	configDescription := config.GetConfigDescription(opts.name)
+
+	if len(configDescription) == 0 {
+		return fmt.Errorf("no profile with name '%s'\n", opts.name)
```
gssbzn, line 33: the linting does not like the \n at the end

internal/config/profile.go
```diff
@@ -225,18 +225,41 @@ func (p *profile) SetOrgID(v string) {
...
+	return m
+}
+
 // Load loads the configuration from disk
-func Load() error { return p.Load() }
-func (p *profile) Load() error {
+func Load() error { return p.Load(false) }
+func (p *profile) Load(readEnvironmentVars bool) error {
```
gssbzn, line 247: this may be the reason the e2e is failing

internal/config/profile.go
```diff
@@ -225,18 +225,41 @@ func (p *profile) SetOrgID(v string) {
...
+		m[publicAPIKey] = redactedValue
+	}
+
+	return m
+}
+
 // Load loads the configuration from disk
-func Load() error { return p.Load() }
-func (p *profile) Load() error {
+func Load() error { return p.Load(false) }
```
gssbzn, line 246: Found why, this should be true
```suggestion
func Load() error { return p.Load(true) }
```
let's fix this or revert all this

robcarlan-mlab: ha, good spot. Of course it was just something simple

PR #225, opened 2020-06-30
CLOUDP-66131: add config rename command

robcarlan-mlab:
## Proposed changes

Adds `config rename <oldName> <newName>`

When the current profile doesn't exist, it errors:

```
./bin/mongocli config rename new123 default
Error: profile new123 does not exist
```

Normal flow (replace to an unused profile name_:

Config before:
```
cat ~/.config/mongocli.toml

[default]
  private_api_key = "123123123"
  public_api_key = "1212312"
  service = "cloud"

[new]
  private_api_key = "141414"
  public_api_key = "123"
  service = "cloud"
```

```
>./bin/mongocli config rename new new1
```

Config after:

```
cat ~/.config/mongocli.toml

[default]
  private_api_key = "123123123"
  public_api_key = "1212312"
  service = "cloud"

[new1]
  private_api_key = "141414"
  public_api_key = "123"
  service = "cloud"
```

Replacing a profile to a name that already exists:

```
./bin/mongocli config rename new default
? There is already a profile called default. Do you want to replace it? Yes
```

The `new` entry is removed, and the default entry contains what used to be the `new` entry
```
cat ~/.config/mongocli.toml

[default]
  private_api_key = "141414"
  public_api_key = "123"
  service = "cloud"
```

_Jira ticket:_ [mongocli config rename <oldName> <newName>](https://jira.mongodb.org/browse/CLOUDP-66131)

## Checklist

### DIFF
--- a/internal/cli/config/rename.go
+++ b/internal/cli/config/rename.go
@@ -1,0 +1,82 @@
+// Copyright 2020 MongoDB Inc
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package config
+
+import (
+	"fmt"
+
+	"github.com/AlecAivazis/survey/v2"
+	"github.com/mongodb/mongocli/internal/prompt"
+
+	"github.com/mongodb/mongocli/internal/config"
+	"github.com/mongodb/mongocli/internal/description"
+	"github.com/spf13/cobra"
+)
+
+type RenameOpts struct {
+	oldName string
+	newName string
+}
+
+func (opts *RenameOpts) Run() error {
+	oldProfile := config.GetConfigDescription(opts.oldName, false)
+	if len(oldProfile) == 0 {
+		return fmt.Errorf("profile %v does not exist", opts.oldName)
+	}
+
+	shouldReplace := false
+	newProfile := config.GetConfigDescription(opts.newName, false)
+	if len(newProfile) > 0 {
+		p := prompt.NewProfileReplaceConfirm(opts.newName)
+		if err := survey.AskOne(p, &shouldReplace); err != nil {
+			return err
+		}
+
+		if !shouldReplace {
+			return nil
+		}
+	}
+
+	if err := config.DeleteConfig(opts.oldName); err != nil {
+		return err
+	}
+
+	if err := config.DeleteConfig(opts.newName); err != nil {
+		return err
+	}
+
+	if err := config.SetConfig(opts.newName, oldProfile); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func RenameBuilder() *cobra.Command {
+	opts := &RenameOpts{}
+	cmd := &cobra.Command{
+		Use:   "rename <oldName> <newName>",
+		Short: description.ConfigRenameDescription,
+		Long:  fmt.Sprintf(description.ConfigRenameLong),
+		Args:  cobra.ExactArgs(2),
+		RunE: func(cmd *cobra.Command, args []string) error {
+			opts.oldName = args[0]
+			opts.newName = args[1]
+			return opts.Run()
+		},
+	}
+
+	return cmd
+}
--- a/internal/cli/config/rename.go
+++ b/internal/cli/config/rename.go
@@ -30,13 +30,11 @@ type RenameOpts struct {
 }
 
 func (opts *RenameOpts) Run() error {
-	config.SetName(&opts.oldName)
-	if config.IsProfileEmpty() {
+	if config.Exists(opts.oldName) {
 		return fmt.Errorf("profile %v does not exist", opts.oldName)
 	}
 
-	config.SetName(&opts.newName)
-	if !config.IsProfileEmpty() {
+	if !config.Exists(opts.newName) {
 		replaceExistingProfile := false
 		p := prompt.NewProfileReplaceConfirm(opts.newName)
 		if err := survey.AskOne(p, &replaceExistingProfile); err != nil {
@@ -44,6 +42,7 @@ func (opts *RenameOpts) Run() error {
 		}
 
 		if !replaceExistingProfile {
+			fmt.Printf("Profile was not renamed.\n")
 			return nil
 		}
 	}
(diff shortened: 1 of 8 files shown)


internal/cli/config/rename.go
```diff
@@ -0,0 +1,82 @@
...
+	if len(newProfile) > 0 {
+		p := prompt.NewProfileReplaceConfirm(opts.newName)
+		if err := survey.AskOne(p, &shouldReplace); err != nil {
+			return err
+		}
+
+		if !shouldReplace {
+			return nil
+		}
+	}
```
gssbzn, line 50: [q] can this happen? you have a validation that the second arg is needed so could this happen?
also I like this idea so if you want to keep it I'm ok with it let's just make sure it can happen 👍

internal/cli/config/rename.go
```diff
@@ -0,0 +1,82 @@
...
+
+	if err := config.DeleteConfig(opts.oldName); err != nil {
+		return err
+	}
+
+	if err := config.DeleteConfig(opts.newName); err != nil {
```
gssbzn, line 56: this is a bit scary, I would warn users and confirm first

internal/config/profile.go
```diff
@@ -226,20 +229,65 @@ func (p *profile) SetOrgID(v string) {
...
+		} else {
+			newSettings[k] = v
+		}
+	}
+
+	return newSettings
+}
+
+// DeleteConfig deletes an existing configuration
+func DeleteConfig(name string) error {
```
gssbzn, line 251: this approach works for the delete command, I'd even suggest let's do that first since it feels simpler than rename right now, but writing and reloading from disk is a bit slow, specially doing it multiple times for a single comamnd, I'd like rename to do all in memory in one go, find, delete, replace

gssbzn, line 251: [big nit] this may have not being obvious at first glance but this files tries to follow a singleton pattern. `p` is your singleton instance so this method should follow the pattern
short suggestion
```go
func Delete() error { return p.Delete() }
func (p *profile) Delete() error {
...  
 t.Delete(&p.name)
...
}
```
would follow the style of this file for this operation

internal/config/profile.go
```diff
@@ -226,20 +229,65 @@ func (p *profile) SetOrgID(v string) {
...
+	if _, err := f.WriteString(s); err != nil {
+		return err
+	}
+
+	// Force reload, so that viper has the new configuration
+	return Load()
+}
+
+func SetConfig(name string, newConfig map[string]string) error {
+	p.SetName(&name)
```
gssbzn, line 284: Similar to the above comment about the singleton, this could potentially use the same pattern
```go
func SetConfig(c map[string]string) error { return p.SetConfig(c) }
func (p *profile) SetConfig(c map[string]string) error { ... }
```

Another option if you want to keep the function as global I'd avoid relying on the singleton instance and instead do something like

```go
newP := newProfile()
newP.name = &name
```

I think at the end a combination of both could be the ideal solution

internal/config/profile.go
```diff
@@ -226,20 +229,65 @@ func (p *profile) SetOrgID(v string) {
...
+
+	for k, v := range newConfig {
+		p.Set(k, v)
 	}
 
-	return m
+	return p.Save()
```
gssbzn, line 290: none of our other setters set and save so let's not do this, like I mentioned before disk are slow so we rather do everything on memory and then commit

internal/config/profile.go
robcarlan-mlab: I think that might be the easier solution -- I'll make a delete command now based on this. Rename might be easier in a similar way, just editing the toml directly

internal/cli/config/rename.go
robcarlan-mlab: The validation doesn't check that the profile exists -- thinking about it more I like this way too. Means that they dont need to run another command and so it should be easier to use

internal/cli/config/rename.go
robcarlan-mlab: Rename now edits the toml directly, so this should be atomic and so I think we don't need a prompt for this now

internal/cli/config/rename.go
```diff
@@ -0,0 +1,73 @@
...
+}
+
+func (opts *RenameOpts) Run() error {
+	config.SetName(&opts.oldName)
+	if config.IsProfileEmpty() {
+		return fmt.Errorf("profile %v does not exist", opts.oldName)
+	}
+
+	config.SetName(&opts.newName)
+	if !config.IsProfileEmpty() {
```
gssbzn, line 39: I'm ok with the method `IsProfileEmpty` but its use is a bit smelly, this part of setting the name to the old name, then setting it to the new, then setting it to the old again feels a bit weird.
When I designed the singleton in the `config` package was with the idea users would usually deal with a single source of settings, here similar to the `config list` command I think should be ok to be more global having a function like `config.Exists(name)` and that exists could call `List()` which return an array of names, and then just use `search.StringInSlice`

internal/cli/config/rename.go
```diff
@@ -0,0 +1,73 @@
...
+		replaceExistingProfile := false
+		p := prompt.NewProfileReplaceConfirm(opts.newName)
+		if err := survey.AskOne(p, &replaceExistingProfile); err != nil {
+			return err
+		}
+
+		if !replaceExistingProfile {
+			return nil
```
gssbzn, line 47: we should print a message that nothing happened

internal/cli/config/rename.go
```diff
@@ -0,0 +1,73 @@
...
+		}
+	}
+
+	config.SetName(&opts.oldName)
+	if err := config.Rename(opts.newName); err != nil {
+		return err
+	}
+
+	return nil
```
gssbzn, line 56: You can simplify this to be just
```suggestion
	return config.Rename(opts.newName)
```

gssbzn: But I'd keep it as it is if you add a confirmation message at the end

internal/cli/config/rename.go
```diff
@@ -0,0 +1,73 @@
...
+	oldName string
+	newName string
+}
+
+func (opts *RenameOpts) Run() error {
+	if config.Exists(opts.oldName) {
+		return fmt.Errorf("profile %v does not exist", opts.oldName)
+	}
+
+	if !config.Exists(opts.newName) {
```
gssbzn, line 37: [q] should this be 
```suggestion
	if config.Exists(opts.newName) {
```
if it exists confirm?

internal/config/profile.go
```diff
@@ -227,6 +227,12 @@ func (p *profile) SetOrgID(v string) {
 	p.Set(orgID, v)
 }
 
+// Exists returns true if there are any set settings for the profile name.
+func Exists(name string) bool { return p.Exists(name) }
+func (p *profile) Exists(name string) bool {
+	return len(viper.GetStringMap(name)) == 0
```
gssbzn, line 233: similar to above, shouldn't this be 
```suggestion
	return len(viper.GetStringMap(name)) > 0
```

To have semantical sense?

internal/cli/config/rename.go
robcarlan-mlab: Oops yes updating this now

PR #226, opened 2020-07-01
CLOUDP-66292 config delete

robcarlan-mlab:
## Proposed changes

_Jira ticket:_ [mongocli config delete|rm <name>](https://jira.mongodb.org/browse/CLOUDP-66292)

## Checklist

## Further comments

Delete flow:
```
> ./bin/mongocli config ls
Available profiles:
  default
  hi1
  hi3
  hi4

> ./bin/mongocli config rm hi4
? Are you sure you want to delete: hi4 Yes

> cat ~/.config/mongocli.toml

[default]
  private_api_key = "123123123"
  public_api_key = "1212312"
  service = "cloud"

[hi1]
  a = "b"
  service = "cloud"

[hi3]
  private_api_key = "redacted"
  public_api_key = "redacted"
  service = ""

```

Delete non-existent profile:
```
> ./bin/mongocli config rm non-existent
Error: profile non-existent does not exist
```

### DIFF
--- a/internal/cli/config/delete.go
+++ b/internal/cli/config/delete.go
@@ -25,27 +25,28 @@ import (
 	"github.com/spf13/cobra"
 )
 
-type RenameOpts struct {
+type DeleteOpts struct {
 	name string
 }
 
-func (opts *RenameOpts) Run() error {
-	profile := config.GetConfigDescription(opts.name)
+func (opts *DeleteOpts) Run() error {
+	config.SetName(&opts.name)
+
+	profile := config.GetConfigDescription()
 	if len(profile) == 0 {
 		return fmt.Errorf("profile %v does not exist", opts.name)
 	}
 
-	shouldReplace := false
+	shouldDelete := false
 	p := prompt.NewDeleteConfirm(opts.name)
-	if err := survey.AskOne(p, &shouldReplace); err != nil {
+	if err := survey.AskOne(p, &shouldDelete); err != nil {
 		return err
 	}
 
-	if !shouldReplace {
+	if !shouldDelete {
 		return nil
 	}
 
-	config.SetName(&opts.name)
 	if err := config.Delete(); err != nil {
 		return err
 	}
@@ -54,7 +55,7 @@ func (opts *RenameOpts) Run() error {
 }
 
 func DeleteBuilder() *cobra.Command {
-	opts := &RenameOpts{}
+	opts := &DeleteOpts{}
 	cmd := &cobra.Command{
 		Use:     "delete <name>",
 		Aliases: []string{"rm"},
--- a/internal/cli/config/delete.go
+++ b/internal/cli/config/delete.go
@@ -17,8 +17,7 @@ package config
 import (
 	"fmt"
 
-	"github.com/AlecAivazis/survey/v2"
-	"github.com/mongodb/mongocli/internal/prompt"
+	"github.com/mongodb/mongocli/internal/cli"
 
 	"github.com/mongodb/mongocli/internal/config"
 	"github.com/mongodb/mongocli/internal/description"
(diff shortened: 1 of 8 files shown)


internal/config/profile.go
```diff
@@ -227,19 +230,52 @@ func (p *profile) SetOrgID(v string) {
 
 // GetConfigDescription returns a map describing the configuration
 func GetConfigDescription(name string) map[string]string {
-	redactedValue := "redacted"
+	settings := viper.GetStringMapString(name)
+	newSettings := make(map[string]string)
```
robcarlan-mlab, line 234: this was because the settings were being edited in place, so it would actually change & potentially save the api key as "redacted" 😰

internal/cli/config/delete.go
```diff
@@ -0,0 +1,71 @@
...
+	profile := config.GetConfigDescription()
+	if len(profile) == 0 {
+		return fmt.Errorf("profile %v does not exist", opts.name)
+	}
+
+	shouldDelete := false
+	p := prompt.NewDeleteConfirm(opts.name)
+	if err := survey.AskOne(p, &shouldDelete); err != nil {
+		return err
+	}
```
gssbzn, line 44: we have defined a pattern for this as we have a lot of commands that do this (delete and confirm) you can create a composition in your `DeleteOpts` of the `cli.DeleteOpts` and this should give you access to a Prompt() method you can use in the PreRun block of the `cobra.Command`

you can see an example [here](https://github.com/mongodb/mongocli/blob/238a113358ce6add54198173f430cc6d96c369e8/internal/cli/projects/delete.go)

internal/config/profile.go
```diff
@@ -18,8 +18,11 @@ import (
 	"errors"
 	"fmt"
 	"log"
+	"os"
 	"sort"
 
+	"github.com/pelletier/go-toml"
+
```
gssbzn, line 25: similar to above let's remove the empty line

internal/config/profile.go
gssbzn: 😱

internal/config/profile.go
```diff
@@ -226,20 +229,54 @@ func (p *profile) SetOrgID(v string) {
...
+		return err
+	}
+
+	if _, err := f.WriteString(s); err != nil {
+		return err
 	}
 
-	return m
+	// Force reload, so that viper has the new configuration
+	return Load()
```
gssbzn, line 279: [q] do we need this? if yes you may want to use `p.Load()`

robcarlan-mlab: It'd be more correct if we were to do anything after deleting a profile, otherwise the old configuration would still be cached. I guess that's a judgement call though -- would be quicker without it and probably won't happen
+1 on p.load instead

gssbzn: up to you here, I don't see many people deleting their profiles so speed is not my concern for this one

gssbzn: just add to the comment of the method that the profile will be reloaded as we don't do this for other methods

internal/cli/config/delete.go
```diff
@@ -0,0 +1,61 @@
...
+		Aliases: []string{"rm"},
+		Short:   description.ConfigDeleteDescription,
+		Args:    cobra.ExactArgs(1),
+		PreRunE: func(cmd *cobra.Command, args []string) error {
+			opts.Entry = args[0]
+
+			config.SetName(&opts.Entry)
+			profile := config.GetConfigDescription()
+			if len(profile) == 0 {
+				return fmt.Errorf("profile %v does not exist", opts.Entry)
```
robcarlan-mlab, line 50: this is to give the not found precedence over the prompt

PR #228, opened 2020-07-02
fix: alerts e2e

gssbzn:
## Proposed changes

Small change to make the 2e tests of alerts more portable across projects

## Checklist

### DIFF
--- a/e2e/atlas/alerts_test.go
+++ b/e2e/atlas/alerts_test.go
@@ -39,6 +39,7 @@ func TestAlerts(t *testing.T) {
 	if err != nil {
 		t.Fatalf("unexpected error: %v", err)
 	}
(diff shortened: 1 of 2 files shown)


e2e/atlas/alerts_test.go
```diff
@@ -33,12 +33,89 @@ const (
...
 	atlasEntity := "atlas"
 	alertsEntity := "alerts"
-	alertID := "5ecbef6b2359825e889837a7"
+	var alertID string
 
 	cliPath, err := cli()
 	if err != nil {
 		t.Fatalf("unexpected error: %v", err)
 	}
+	t.Run("List", func(t *testing.T) {
```
gssbzn, line 42: the main think is to run listing first, grab the id from there instead of hardcoded and then use it for the rest

PR #229, opened 2020-07-02
CLOUDP-66571: add tests to profile

robcarlan-mlab:
## Proposed changes

_Jira ticket:_ [Add tests to profile](https://jira.mongodb.org/browse/CLOUDP-66571)

## Checklist

## Further comments


go.mod
```diff
@@ -20,6 +20,7 @@ require (
 	github.com/spf13/jwalterweatherman v1.1.0 // indirect
 	github.com/spf13/pflag v1.0.5 // indirect
 	github.com/spf13/viper v1.7.0
+	github.com/stretchr/testify v1.6.1
```
robcarlan-mlab, line 23: Not sure how you feel about adding this library in the PR -- saved a bunch of `if diff ... ` statements and seems to have the same functionality

internal/config/profile_test.go
```diff
@@ -0,0 +1,304 @@
...
+
+	if err := profile.Load(false); err != nil {
+		t.Error(err)
+	}
+
+	availableProfiles := profile.List()
+	assert.Equal(t, 2, len(availableProfiles), "expected to find 2 profiles")
+}
+
+func TestProfileDescribeFullProfile(t *testing.T) {
```
gssbzn, line 124: [nit-style] your test method should match the method name, since you're testing the struct profile, and its method 
`GetConfigDescription` this test should be 
```suggestion
func TestProfile_GetConfigDescription(t *testing.T) {
```
if you have multiple scenarios you can add them like `TestProfile_GetConfigDescription_FullProfile` tho I prefer to use internal functions/closures  like `t.Run("full profile", func(t *testing.T) {` the second let's you do some basic scenarios as code kind of thing, you can check [this test](https://github.com/mongodb/mongocli/blob/f480e0f9ca05bdb5e9177029f38bc3dcc9a97657/internal/convert/cluster_config_test.go#L25) for an example of what I mean (see how I define a name, initial state, changes, expected map and then apply the `t.Run` function to each scenario

internal/config/profile_test.go
```diff
@@ -0,0 +1,304 @@
...
+
+	desc := profile.GetConfigDescription()
+	assert.Equal(t, "default", Name())
+	assert.Equal(t, 9, len(desc))
+	assert.Equal(t, "cloud", Service())
+	assert.Equal(t, "some_public_key", PublicAPIKey())
+	assert.Equal(t, "some_private_key", PrivateAPIKey())
+	assert.Equal(t, "http://om_url.com/", OpsManagerURL())
+	assert.Equal(t, "/path/to/certificate", OpsManagerCACertificate())
+	assert.Equal(t, "false", OpsManagerSkipVerify())
```
gssbzn, line 142: reading the docs of assert they recommend that when you do multiple assertions is better to do 
`a : = assert.New()` then all your `.Equal` should be against `a` like `a.Equal` (I'm using a instead of assert as they docs as our linter will tell you you're shadowing an import that way)

internal/config/profile_test.go
```diff
@@ -0,0 +1,304 @@
...
+
+	profile.SetName(atlasProfile)
+
+	if err := profile.Load(false); err != nil {
+		t.Error(err)
+	}
+
+	desc := profile.GetConfigDescription()
+	assert.Equal(t, atlasProfile, profile.Name(), "expected atlas profile to be described")
+	assert.Equal(t, 3, len(desc))
```
gssbzn, line 174:
```suggestion
	assert.Len(t, 3, desc)
```

internal/config/profile_test.go
```diff
@@ -0,0 +1,304 @@
...
+		t.Error(err)
+	}
+
+	if err := profile.Load(false); err != nil {
+		t.Error(err)
+	}
+
+	desc := profile.GetConfigDescription()
+	assert.Equal(t, DefaultProfile, Name())
+	assert.Equal(t, 4, len(desc))
```
gssbzn, line 157:
```suggestion
	assert.Len(t, 4, desc)
```

internal/config/profile_test.go
```diff
@@ -0,0 +1,304 @@
...
+	if err != nil {
+		t.Error(err)
+	}
+
+	if err := profile.Load(false); err != nil {
+		t.Error(err)
+	}
+
+	availableProfiles := profile.List()
+	assert.Equal(t, 2, len(availableProfiles), "expected to find 2 profiles")
```
gssbzn, line 121:
```suggestion
	assert.Len(t, 2, availableProfiles, "expected to find 2 profiles")
```

internal/config/profile_test.go
```diff
@@ -0,0 +1,304 @@
...
+	profile, err := createProfileWithOneDefaultUserOneNonDefault()
+	if err != nil {
+		t.Error(err)
+	}
+
+	if err := profile.Load(false); err != nil {
+		t.Error(err)
+	}
+
+	assert.Equal(t, true, profile.Exists(DefaultProfile), "default profile should exist")
```
gssbzn, line 229:
```suggestion
	assert.True(t,  profile.Exists(DefaultProfile), "default profile should exist")
```

internal/config/profile_test.go
```diff
@@ -0,0 +1,304 @@
...
+	assert.Equal(t, true, profile.Exists(atlasProfile), "atlas profile should exist")
+	assert.Equal(t, false, profile.Exists("not_a_profile"), "not_a_profile profile should not exist")
```
gssbzn, line 231:
```suggestion
	assert.False(t, profile.Exists("not_a_profile"), "not_a_profile profile should not exist")
```

internal/config/profile_test.go
```diff
@@ -0,0 +1,304 @@
...
+const (
+	atlasProfile   = "atlas"
+	newProfileName = "newProfileName"
+)
+
+var testConfigDir = "/test_dir"
+var fs afero.Fs
+var filename = fmt.Sprintf("%v/%v.toml", testConfigDir, ToolName)
+
+func TestMain(m *testing.M) {
```
gssbzn, line 35: Please consider that having this here means that `internal/config/home_test.go` is also affected by this main

gssbzn:
One unrelated thing here is that I just found out that the rename config does not work and these tests wouldn't have caught that error so I'm adding some e2e to the config command to prevent this from happening again

robcarlan-mlab:
I'll rewrite these as e2e considering you've added the first one already

internal/config/profile.go
```diff
@@ -104,37 +113,41 @@ func newProfile() *profile {
...
 		configDir: configDir,
 		fs:        afero.NewOsFs(),
 	}
 	return np
 }
 
+func SetConfigPath(name string) { p.SetConfigPath(name) }
+func (p *profile) SetConfigPath(name string) {
+	p.configDir = name
+}
```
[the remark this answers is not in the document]
robcarlan-mlab: [removed this]

internal/config/profile_test.go
```diff
@@ -0,0 +1,304 @@
...
+	fs = afero.NewMemMapFs()
+
+	viper.SetFs(fs)
+	SetConfigPath(testConfigDir)
```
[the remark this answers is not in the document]
robcarlan-mlab: Changed this slightly again, so it doesn't need this now

.githooks/pre-commit
```diff
@@ -10,7 +10,8 @@ done
 
 if [[ -n "${STAGED_GO_FILES}" ]]; then
     set -o errexit
-    go test -race ./internal...
+    make test
+    make integration-test
```
robcarlan-mlab, line 14: Probably might not want these here in the future, but I think for now its ok

internal/config/profile_integration_test.go
```diff
@@ -0,0 +1,279 @@
...
+	}
+
+	filename := fmt.Sprintf("%v/%v.toml", testConfigDir, ToolName)
+
+	err = afero.WriteFile(p.fs, filename, []byte(profileContents), 0600)
+	if err != nil {
+		return err
+	}
+
+	return nil
```
gssbzn, line 47: you can just 
```suggestion	
	return afero.WriteFile(p.fs, filename, []byte(profileContents), 0600)
```

internal/config/profile_integration_test.go
```diff
@@ -0,0 +1,279 @@
...
+	"github.com/stretchr/testify/assert"
+)
+
+const (
+	atlasProfile   = "atlas"
+	newProfileName = "newProfileName"
+)
+
+func createProfile(profileContents string) error {
+	p.fs = afero.NewMemMapFs()
```
gssbzn, line 33: I'm a bit worry that we are testing the global object and this could bring flakiness between tests, like if we change order or they interact with each other in some way we are not seeing (go tests also run in parallel)

I think I mentioned before that ideally each test should be testing its own `p` and this is doable
``` go
func TestProfile_Get_FullProfile(t *testing.T) {
  myLocalP := newProfile() // or even &profile{} if you want more control
  myLocalP.fs = afero.NewMemMapFs()
  myLocalP.Load()
  myLocalP.Get()
  // all my assertions
}
```

.githooks/pre-commit
gssbzn: works for me but how do you feel creating a `unit-test` target and making `make test` call both unit and integration?

internal/config/profile_integration_test.go
```diff
@@ -0,0 +1,279 @@
...
+  profile_id = "5cac6a2179358edabd12b572"
+  service = "cloud"
+`
+	return createProfile(contents)
+}
+
+func TestProfile_List(t *testing.T) {
+	err := createProfileWithOneDefaultUserOneNonDefault()
+	if err != nil {
+		t.Error(err)
```
gssbzn, line 95:
```suggestion
		t.Fatal(err)
```
If you fail to create the profile then the rest of the test are going to fail as well

internal/config/profile_integration_test.go
```diff
@@ -0,0 +1,279 @@
...
+	}
+
+	if err := Load(); err != nil {
+		t.Error(err)
```
gssbzn, line 99: Same
```suggestion
		t.Fatal(err)
```

internal/config/profile_integration_test.go
```diff
@@ -0,0 +1,279 @@
...
+
+	SetName(DefaultProfile)
+
+	if err := Load(); err != nil {
+		t.Error(err)
+	}
+
+	desc := Get()
+
+	a := assert.New(t)
```
gssbzn, line 120: welcome to the single letter variables world

internal/config/profile_integration_test.go
```diff
@@ -0,0 +1,279 @@
...
+	a.Equal("some_private_key", PrivateAPIKey())
+	a.Equal("http://om_url.com/", OpsManagerURL())
+	a.Equal("/path/to/certificate", OpsManagerCACertificate())
+	a.Equal("false", OpsManagerSkipVerify())
+}
+
+func TestProfile_Get_Default(t *testing.T) {
+	err := createProfileWithOneDefaultUserOneNonDefault()
+	if err != nil {
+		t.Error(err)
```
gssbzn, line 134: same and multiple of other test, this is how I like to think about it, can my test continue after this error? if so let's use `t.Error` if not (I'll just keep getting errors) let's use `t.Fatal`

.githooks/pre-commit
```diff
@@ -10,7 +10,8 @@ done
 
 if [[ -n "${STAGED_GO_FILES}" ]]; then
     set -o errexit
-    go test -race ./internal...
```
gssbzn, line 13: FYI the reason I don't use make test here is that `make test` ignores cache, I wanted the pre commit to be as fast as possible so I made this change

internal/config/profile_integration_test.go
robcarlan-mlab: I think for me I'd rather use `Fatal` (practically) everywhere as it saves thinking about possible failure conditions on tests, when the test has probably lost any meaningful value as soon as the first error is hit. (It's why I kept using `Error` by default because it doesn't feel intuitive that a test should continue running after reaching an error)

.githooks/pre-commit
robcarlan-mlab: 👍  Im ok with that

internal/config/profile_integration_test.go
robcarlan-mlab: I went back this way mostly to allow testing `Exists` and `List`, I can go either way on which one though

.githooks/pre-commit
robcarlan-mlab: Ok, sounds like it'd be good to keep the cache here

internal/config/profile_integration_test.go
gssbzn: that works for me as well, keep in mind that doing asser.Equal/True/Len multiple times has a similar behaviour to doing `t.Error` so it's quite the "go way"

.githooks/pre-commit
gssbzn: -count=1 is the flag that force tests to ignore cache, I also believe (tho I could be wrong) enabling coverage disables the cache as well

internal/config/profile_integration_test.go
gssbzn: We can ignore for now and see how it goes, if this starts to get flaky then we know what to do

robcarlan-mlab: I was just testing unit & integration tests together and it started failing, probably something to do with this

robcarlan-mlab:
I made the changes but GIthub is having some issues, so the diff isn't up to date

robcarlan-mlab:
Some e2es are still failing, which could be a follow up issue. I dont think its these tests but I will hold off on merging for now

robcarlan-mlab:
Just had to fix conflicts

PR #231, opened 2020-07-03
CLOUDP-66614: Enable linting on e2e tests

gssbzn:
## Proposed changes

## Checklist

## Further comments

I had a suspicious we weren't linting our e2e files but I could not pin the why, now I know

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -75,5 +75,15 @@ linters:
 
 run:
   tests: true
+  build-tags:
+    - e2e
   skip-dirs:
     - internal/mocks
+
+issues:
+  # Excluding configuration per-path, per-linter, per-text and per-source
+  exclude-rules:
+    # Don't check for cyclomatic complexity on e2e file.
+    - path: e2e
+      linters:
+        - gocyclo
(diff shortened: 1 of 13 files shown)


.golangci.yml
```diff
@@ -75,5 +75,15 @@ linters:
...
   skip-dirs:
     - internal/mocks
+
+issues:
+  # Excluding configuration per-path, per-linter, per-text and per-source
+  exclude-rules:
+    # Don't check for cyclomatic complexity on e2e file.
+    - path: e2e
+      linters:
+        - gocyclo
```
andreaangiolillo, line 89: [q] Could you explain me what this piece of code does?

andreaangiolillo: I am not sure what's the purpose of `issues`

gssbzn: what it says at the top
>  # Don't check for cyclomatic complexity on e2e file.

Our e2e files are long and complex and were hitting this lint I don't think there's value to fix them given that they read fine as they are and is an artefact of having to create something and then delete it in the same test

gssbzn: For clarity on `issues`, it's to ignore known issues on your code so you can add golint and not have to fix them all at once

PR #247, opened 2020-07-10
CLOUDP-66900: mongocli atlas cluster(s) onlineArchive(s) delete

gssbzn:
## Proposed changes

_Jira ticket:_ [CLOUDP-66900](https://jira.mongodb.org/browse/CLOUDP-66900)

## Checklist

## Further comments

```zsh
mongocli atlas clusters onlinearchive rm 5f04b0e13fe0b32bb2d78af8 --clusterName testShard 
? Are you sure you want to delete: 5f04b0e13fe0b32bb2d78af8 (y/N)
```


internal/store/online_archives.go
```diff
@@ -53,3 +57,14 @@ func (s *Store) OnlineArchive(projectID, clusterName, archiveID string) (*atlas.
 		return nil, fmt.Errorf("unsupported service: %s", s.service)
 	}
 }
+
+// DeleteOnlineArchive encapsulate the logic to manage different cloud providers
+func (s *Store) DeleteOnlineArchive(projectID, clusterName, archiveID string) error {
+	switch s.service {
+	case config.CloudService:
+		_, err := s.client.(*atlas.Client).OnlineArchives.Delete(context.Background(), projectID, clusterName, archiveID)
```
colm-quinn, line 65: Having the go Atlas client available really makes these easy to implement!

internal/cli/atlas/onlinearchive/delete.go
```diff
@@ -0,0 +1,76 @@
...
+	return opts.Delete(opts.store.DeleteOnlineArchive, opts.ConfigProjectID(), opts.clusterName)
+}
+
+// mongocli atlas cluster(s) onlineArchive(s) delete <id> [--clusterName name][--projectId projectId][--force]
+func DeleteBuilder() *cobra.Command {
+	opts := &DeleteOpts{
+		DeleteOpts: cli.NewDeleteOpts("Archive '%s' deleted\n", "Archive not deleted"),
+	}
+	cmd := &cobra.Command{
+		Use:     "delete <id>",
```
colm-quinn, line 51: [q] Are clusterName and projectId not always required? This seems to read as those are optional parameters, but that's not the case here: https://github.com/mongodb/go-client-mongodb-atlas/blob/master/mongodbatlas/online_archives.go#L137-L144

gssbzn: that's actually a fair point, I've been forgetting to mark the flag as required and the copy/past error has continued 
good catch

gssbzn: @colm-quinn fixed here and other commands

colm-quinn: Cool, looks good.
There's still a few small questions from me on this, part is I'm probably missing what cobra is doing behind the scenes.

1. Is the required on the `projectID` inferred from another place, as I don't see it here?
2. I'm still unsure about what gets printed to screen here. The method comment has `delete <id> [--clusterName name][--projectId projectId][--force]` but the `Use:     "delete <id>"` above doesn't imply to me that they will show in the help.

I'm assuming there's some magic in cobra that does this, but would be good to confirm.

gssbzn: 1. Project ID is more complex because it can come from a flag, env variable or your profile file, so what I've done is a little shared function here https://github.com/mongodb/mongocli/blob/55cce5ce4dbf25efa84fc3301f5e42ead7a3261b/internal/cli/global_opts.go#L53, that validates is present via any of the available options, and is a valid `ObjectId`, we use this function here on line 56

2. On the `Use:` property we only define things that are passed as args and not flags, cobra doesn't enforce this and we do it as a nice to have as when you print usage you get
```
mongocli atlas clusters onlineArchives delete <id> [flags]
```
We don't define flags there, cobra already adds a ` [flags]` as you can see, and since every flag has help you get their help detailed, heres the full example of this command
```
mongocli atlas clusters onlinearchive rm --help
Delete an online archives from a cluster.

Usage:
  mongocli atlas clusters onlineArchives delete <id> [flags]

Aliases:
  delete, rm

Flags:
      --clusterName string   Name of the cluster.
      --force                Don't ask for confirmation.
  -h, --help                 help for delete
      --projectId string     Project ID to use. Overrides configuration file or environment variable settings.

Global Flags:
  -P, --profile string   Profile to use from your configuration file.
```

gssbzn: doing this helped me spot a typo on the description of the command

colm-quinn: Excellent, thanks for the detailed explanation - that makes it a lot clearer.

PR #273, opened 2020-07-21
CLOUDP-67657: atlas certs list

robcarlan-mlab:
## Proposed changes

Added `mongocli dbusers certs list|ls <username>`. Moved DbUsers into a package & clarified the naming in the X509 store. 

_Jira ticket:_ CLOUDP-67657

## Checklist

## Further comments

I had some issues with adding "." in the name, e.g. the command wouldn't work when I made a "rob.carlan" user. But this is also broken when I try via Postman, so I filed https://jira.mongodb.org/browse/CLOUDP-67761

```
./bin/mongocli atlas dbusers certs list newuser
[
  {
    "_id": 2039020905786260978,
    "createdAt": "2020-07-22T14:50:23Z",
    "groupId": "5cac6a2ba6f239298a18a025",
    "notAfter": "2020-10-22T15:50:23Z",
    "subject": "CN=newuser"
  }
]
```


internal/store/x509_auth_database_users.go
```diff
@@ -24,21 +24,26 @@ import (
...
-type X509CertificateSaver interface {
+type X509CertificateConfSaver interface {
 	SaveX509Configuration(string, string) (*atlas.CustomerX509, error)
 }
 
 type X509CertificateDisabler interface {
 	DisableX509Configuration(string) error
 }
 
+type UserCertificateLister interface {
```
gssbzn, line 39: [nit] to avoid confusion 
```suggestion
type DBUserCertificateLister interface {
```

internal/store/x509_auth_database_users.go
```diff
@@ -24,21 +24,26 @@ import (
...
+	GetUserCertificates(string, string) ([]atlas.UserCertificate, error)
```
gssbzn, line 40: same here
```suggestion
	GetDBUserCertificates(string, string) ([]atlas.UserCertificate, error)
```

gssbzn: I would also say these feel more natural in the db_users store but that's totally optional as the files have no meaning  at all

internal/store/x509_auth_database_users.go
```diff
@@ -24,21 +24,26 @@ import (
...
+}
+
 type X509CertificateStore interface {
-	X509CertificateDescriber
-	X509CertificateSaver
+	UserCertificateLister
```
gssbzn, line 44: I would not mix these together (atlas managed/customer managed) but I don't feel strong about it

internal/store/x509_auth_database_users.go
```diff
@@ -65,6 +70,17 @@ func (s *Store) SaveX509Configuration(projectID, certificate string) (*atlas.Cus
 	}
 }
 
+// GetUserCertificates retrieves the current user managed certificates for a database user
+func (s *Store) GetUserCertificates(projectID, username string) ([]atlas.UserCertificate, error) {
```
gssbzn, line 74: similar nit to above, it can be confusing just saying "user" as we have DB users and Atlas users

internal/store/x509_auth_database_users.go
robcarlan-mlab: Cool yeah i wasn't sure whether to keep them together because the client store ends up being the same, but I think separate will be easier

PR #274, opened 2020-07-22
CLOUDP-67735: Handle ca certificate for digest transport

gssbzn:
## Proposed changes

_Jira ticket:_ CLOUDP-67735

## Checklist

## Further comments

We are not able to rely on the go client for setting the certificate, so I had to move some of the logic to the cli, I've tested this manually

### DIFF
--- a/internal/cli/opsmanager/ops_manager.go
+++ b/internal/cli/opsmanager/ops_manager.go
@@ -37,7 +37,11 @@ func Builder() *cobra.Command {
 		Short:   description.OpsManager,
 		PersistentPreRunE: func(cmd *cobra.Command, args []string) error {
 			config.SetService(config.OpsManagerService)
-			return validate.Credentials()
+			// do not validate to create an owner
+			if cmd.CommandPath() != "mongocli ops-manager owner create" {
+				return validate.Credentials()
+			}
+			return nil
 		},
 	}
 
--- a/internal/cli/owner/create.go
+++ b/internal/cli/owner/create.go
@@ -98,7 +98,7 @@ func CreateBuilder() *cobra.Command {
 	cmd.Flags().StringVar(&opts.lastName, flag.LastName, "", usage.LastName)
 	cmd.Flags().StringSliceVar(&opts.whitelistIps, flag.WhitelistIP, []string{}, usage.WhitelistIps)
 
-	_ = cmd.MarkFlagRequired(flag.Username)
+	_ = cmd.MarkFlagRequired(flag.Email)
 	_ = cmd.MarkFlagRequired(flag.FirstName)
 	_ = cmd.MarkFlagRequired(flag.LastName)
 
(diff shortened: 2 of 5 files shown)


internal/cli/opsmanager/ops_manager.go
```diff
@@ -37,7 +37,11 @@ func Builder() *cobra.Command {
 		Short:   description.OpsManager,
 		PersistentPreRunE: func(cmd *cobra.Command, args []string) error {
 			config.SetService(config.OpsManagerService)
-			return validate.Credentials()
+			// do not validate to create an owner
+			if cmd.CommandPath() != "mongocli ops-manager owner create" {
```
gssbzn, line 41: drive by fix, owner is an unauth request

internal/store/store.go
```diff
@@ -52,12 +54,32 @@ type Store struct {
...
+		DialContext: (&net.Dialer{
+			Timeout:   timeout,
+			KeepAlive: keepAlive,
+		}).DialContext,
+		MaxIdleConns:          maxIdleConns,
+		MaxIdleConnsPerHost:   maxIdleConnsPerHost,
+		Proxy:                 http.ProxyFromEnvironment,
+		IdleConnTimeout:       idleConnTimeout,
+		ExpectContinueTimeout: expectContinueTimeout,
+		TLSClientConfig:       tlsClientConfig,
```
gssbzn, line 76: TL;DR: you can only set this on construction and the client was relying on reflection to "re set" this if needed, that reflection was failing when using the `digest.Transport` as the lib is agnostic about the authentication we use

internal/store/store.go
```diff
@@ -69,22 +91,54 @@ func newTransport(c config.Config) *digest.Transport {
...
+}
+
+func defaultClient(c config.Config) (*http.Client, error) {
+	client := http.DefaultClient
+	if caCertificate := c.OpsManagerCACertificate(); caCertificate != "" {
+		dat, err := ioutil.ReadFile(caCertificate)
+		if err != nil {
+			return nil, err
+		}
+		fmt.Println(string(dat))
```
savannaendicott, line 126: Was this used for dev/meant to be removed? Same with above on line 108?

internal/cli/owner/create.go
```diff
@@ -98,7 +98,7 @@ func CreateBuilder() *cobra.Command {
 	cmd.Flags().StringVar(&opts.lastName, flag.LastName, "", usage.LastName)
 	cmd.Flags().StringSliceVar(&opts.whitelistIps, flag.WhitelistIP, []string{}, usage.WhitelistIps)
 
-	_ = cmd.MarkFlagRequired(flag.Username)
+	_ = cmd.MarkFlagRequired(flag.Email)
```
savannaendicott, line 101: drive by fix?

gssbzn: yep

internal/store/store.go
gssbzn: oops, thanks for the catch

gssbzn:
@savannaendicott RFAL

PR #275, opened 2020-07-22
CLOUDP-67741: Add a global --output flag

gssbzn:
## Proposed changes

_Jira ticket:_ CLOUDP-67741

## Checklist

## Further comments

This adds a flag to control output format, it defaults to json (what we do) and there's not much to it currently just trying to do this big change as incremental as possible


internal/config/profile.go
```diff
@@ -129,15 +130,18 @@ func (p *profile) SetName(name string) {
 
 func Set(name, value string) { p.Set(name, value) }
 func (p *profile) Set(name, value string) {
-	viper.Set(fmt.Sprintf("%s.%s", p.name, name), value)
+	settings := viper.GetStringMapString(p.Name())
+	settings[name] = value
+	viper.Set(p.name, settings)
```
gssbzn, line 135: this was driving my mad, I had no idea why so many tests started failing, turns out if you set a key with viper, viper do not merge, it will overwrite and since we were doing dots like `default.output` it would overwrite all default to just be `default.output` this also leads me to believe we may have had some unknown bug around with this

internal/cli/config/rename.go
```diff
@@ -60,9 +60,10 @@ func RenameBuilder() *cobra.Command {
 	const argsN = 2
 	opts := &RenameOpts{}
 	cmd := &cobra.Command{
-		Use:   "rename <oldName> <newName>",
-		Short: description.ConfigRenameDescription,
-		Args:  cobra.ExactArgs(argsN),
+		Use:     "rename <oldName> <newName>",
+		Aliases: []string{"mv"},
```
gssbzn, line 64: adding an alias

internal/cli/config/delete.go
```diff
@@ -57,5 +60,7 @@ func DeleteBuilder() *cobra.Command {
 		},
 	}
 
+	cmd.Flags().BoolVar(&opts.Confirm, flag.Force, false, usage.Force)
```
gssbzn, line 63: adding this so I can test it on evg

internal/config/profile.go
```diff
@@ -101,7 +102,7 @@ func List() []string {
 
 // Exists returns true if there are any set settings for the profile name.
 func Exists(name string) bool {
-	return len(viper.GetStringMap(name)) > 0
+	return search.StringInSlice(List(), name)
```
gssbzn, line 105: the original method was giving me problems due to the same problem we have of viper over writing configs

internal/cli/atlas/logs_download.go
```diff
@@ -114,7 +114,7 @@ func LogsDownloadBuilder() *cobra.Command {
 		},
 	}
 
-	cmd.Flags().StringVarP(&opts.out, flag.Out, flag.OutShort, "", usage.LogOut)
+	cmd.Flags().StringVar(&opts.out, flag.Out, "", usage.LogOut)
```
colm-quinn, line 117: [q] I've spotted the removing `StringVarP` in a few places in this PR. I see in `flags.go` you've added `Output` and `OutputShort`. Any reason why you're not still reading both here?

colm-quinn: Ahh reading a bit more, this looks like you're keeping the legacy support for `Out` and adding the new flags `o` and `output`

colm-quinn:
General question, any idea why the e2e tests are failing at the moment on this?

internal/cli/atlas/logs_download.go
gssbzn: yeah the shorthand is now used for output which is global (and affects this command) and I can't have it be a short hand for this other flag :(

gssbzn:
@colm-quinn the only failing tests is a cloud manager IAM issue we are waiting for this to be fixed on cloud

PR #277, opened 2020-07-23
CLOUDP-67829: Interface cleanup

gssbzn:
## Proposed changes

_Jira ticket:_ CLOUDP-67829

## Checklist

## Further comments

Wanted to change how the profile interfaces work given I'm doing some changes where I need more control but had to fix some wrong mocks along the way


internal/config/profile.go
```diff
@@ -27,9 +27,22 @@ import (
 	"github.com/spf13/viper"
 )
 
-//go:generate mockgen -destination=../mocks/mock_profile.go -package=mocks github.com/mongodb/mongocli/internal/config Setter,Saver,SetSaver,Getter,Config
+//go:generate mockgen -destination=../mocks/mock_profile.go -package=mocks github.com/mongodb/mongocli/internal/config SetSaver
```
gssbzn, line 30: this is the main change cleaning unused interfaces and inverting how this work a bit to make it more in line with go practices (return structs accept interfaces)

internal/store/store.go
```diff
@@ -95,7 +95,16 @@ func skipVerifyTransport() http.RoundTripper {
 	}
 }
 
-func authenticatedClient(c config.Config) (*http.Client, error) {
+type Config interface {
```
gssbzn, line 98: this is the other side of the change, and defining exactly what I need here, nothing else nothing more and gives me the freedom to accept any struct that implements the interface I need

This is in line with another change coming to handle different outputs

PR #279, opened 2020-07-23
CLOUDP-67836: Add a function to handle output formats

gssbzn:
## Proposed changes

_Jira ticket:_ CLOUDP-67836

## Checklist

## Further comments

This adds a function to centralise the output management and changing all commands using the json formatter to use this function instead


internal/output/output.go
```diff
@@ -0,0 +1,44 @@
...
+
+type Config interface {
+	Output() string
+}
+
+const jsonFormat = "json"
+
+// Print outputs v to os.Stdout while handling configured formats,
+// if the optional t is given then it's processed as a go-template
+func Print(c Config, t string, v interface{}) error {
```
gssbzn, line 32: this is the only change that matters, the rest is changing all commands to use this

internal/output/output.go
```diff
@@ -0,0 +1,44 @@
...
+	if c.Output() == jsonFormat {
+		return json.PrettyPrint(v)
+	}
+	if t != "" {
+		tmpl, err := template.New("test").Parse(t)
```
robcarlan-mlab, line 37: Should the name be something like "output"?

gssbzn: indeed

gssbzn:
@robcarlan-mlab merged master and updated your last command to use the new func

PR #281, opened 2020-07-23
CLOUDP-62262: Create a simplified view of atlas cluster

gssbzn:
## Proposed changes

_Jira ticket:_ CLOUDP-62262

## Checklist

## Further comments

This still doesn't default to the simplified output but we are getting there

Cluster list
Screenshot 2020-07-23 at 18 27 00

Cluster Show
Screenshot 2020-07-23 at 18 27 45

Cluster create
Screenshot 2020-07-23 at 18 28 32

The rest of the functions are variations of create so I'm not going to go over them


internal/output/output.go
```diff
@@ -38,7 +39,12 @@ func Print(c Config, t string, v interface{}) error {
 		if err != nil {
 			return err
 		}
-		return tmpl.Execute(os.Stdout, v)
+		w := tabwriter.NewWriter(os.Stdout, 0, 8, 2, '\t', 0)
```
colm-quinn, line 42: [np] The collection of numbers on this call really isn't very self documenting unless you dig into the method signatures. A comment as to the intent would be good here I think.

internal/cli/atlas/clusters/create.go
```diff
@@ -64,6 +64,8 @@ func (opts *CreateOpts) initStore() error {
 	return err
 }
 
+var createTmpl = "Deploying cluster {{.Name}}.\n"
```
colm-quinn, line 67: [q] Thinking aloud - I wonder should all these templates defined in one place and pulled in, rather than in their particular module.

My reasoning is consistency of messaging would be easier to maintain in this manner as you are looking at all templates at the same time.

It's probably not a very go like approach though - what do you think?

gssbzn: you're right there in that's not very go like, we have a global "[flag](https://github.com/mongodb/mongocli/tree/master/internal/flag)" package for the same reason you describe and I was temped to follow the same approach but that `flag` and `usage` packages are becoming unmaintainable and a dumping ground of `const`s and I've been thinking if it would be worth to start splitting them

This is my long way of saying, I see your point but I don't know which one is best

internal/output/output.go
gssbzn: will do

internal/cli/atlas/clusters/create.go
gssbzn: (side note as I open the editor) the advantage of having the template in the command is that you have easier access to see the fields of the struct you're templating) so I think I vote on keeping it here

colm-quinn: It's all trade-offs :) To be clear, I have no problem with the current approach, just throwing ideas around. The fact that you've already tried with flags and are reconsidering the approach is evidence enough for me to keep going this the approach in this PR.

PR #287, opened 2020-07-27
CLOUDP-62263:Add a simplified view for clusters operations in Ops manager

gssbzn:
## Proposed changes

_Jira ticket:_ CLOUDP-62263

## Checklist

## Further comments

Screenshot 2020-07-27 at 15 44 57

Screenshot 2020-07-27 at 15 44 39


e2e/atlas/online_archives_test.go
```diff
@@ -160,5 +162,7 @@ func TestOnlineArchives(t *testing.T) {
 		if err != nil {
 			t.Fatalf("unexpected error: %v, resp: %v", err, string(resp))
 		}
+		expected := fmt.Sprintf("Archive '%s' deleted\n", archiveID)
```
gssbzn, line 165: sneaking this here

internal/cli/opsmanager/clusters/describe.go
```diff
@@ -40,20 +40,34 @@ func (opts *DescribeOpts) initStore() error {
...
 		return err
 	}
 
-	clusterConfigs := convert.FromAutomationConfig(result)
-	for _, rs := range clusterConfigs {
+	return output.Print(config.Default(), describeTemplate, r)
+}
+
+func (opts *DescribeOpts) cluster() (interface{}, error) {
+	if config.Output() == "" {
```
robcarlan-mlab, line 57: Seems like the template is only valid with this Cluster type & OM. Will there be a template for the output mode in L67 / the Atlas cluster type?

I'm wondering if there's a way we can avoid using interface{} here and in the other command, but I'm not sure how easy that'd be with the changes to the store

robcarlan-mlab, line 57: should this be a `config.Simple` rather than empty string?

internal/cli/opsmanager/clusters/describe.go
```diff
@@ -40,20 +40,34 @@ func (opts *DescribeOpts) initStore() error {
...
+	}
+	r := convert.FromAutomationConfig(c)
+	for _, rs := range r {
 		if rs.Name == opts.name {
-			return output.Print(config.Default(), "", rs)
+			return rs, nil
 		}
 	}
-	return fmt.Errorf("replicaset %s not found", opts.name)
+	return nil, fmt.Errorf("replicaset %s not found", opts.name)
```
robcarlan-mlab, line 70: maybe: "replicaset" -> "replica set"

internal/cli/opsmanager/clusters/list.go
```diff
@@ -37,21 +37,46 @@ func (opts *ListOpts) init() error {
...
+{{.ClusterID}}	{{.Name}}	{{.Type}}{{end}}{{end}}
+`
+
 func (opts *ListOpts) Run() error {
+	r, err := opts.clusters()
+	if err != nil {
+		return err
+	}
+
+	return output.Print(config.Default(), opts.template(), r)
```
robcarlan-mlab, line 56: Could also potentially choose the template based on a type switch from the interface, but I don't think it's ideal

internal/cli/opsmanager/clusters/describe.go
gssbzn: simple will be "no output selected", and will be the default eventually, so if you want json you'll have to be explicit about it

gssbzn: if no output is selected we call the clusters api, and use the template
if we specify json we use the automation API and transform it to the cli format

The original implementation of this using the automation config was so that listing or describing would let you take that same config modify it and apply it again, something you can't do using the clusters API

That approach had a downside we had no cluster ID, this was a problem if you needed access to the metrics commands so we this change I'm making sure users get back that missing info.

This is my long way on why of this changes

>I'm wondering if there's a way we can avoid using interface{}

Sadly not with go :(

internal/cli/opsmanager/clusters/list.go
gssbzn: I just wasn't sure on returning three values from that other function

internal/cli/opsmanager/clusters/describe.go
robcarlan-mlab: Makes sense -- perhaps this is the best way for now then

PR #293, opened 2020-07-28
CLOUDP-68210: Take cluster e2e tests out of the PR suite

gssbzn:
## Proposed changes

_Jira ticket:_ CLOUDP-68210

## Checklist

### DIFF
--- a/build/ci/evergreen.yml
+++ b/build/ci/evergreen.yml
@@ -1130,7 +1130,7 @@ buildvariants:
       go_bin: "/opt/golang/go1.14/bin"
     tasks:
       - name: .code_health
-  - name: e2e
+  - name: e2e_generic
     display_name: "E2E Tests Generic"
     run_on:
       - rhel70-small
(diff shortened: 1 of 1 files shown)


andreaangiolillo:
I think that one of the reason our tests are flaky is that sometimes clusters are not deleted. I think we could write a small piece of code that run at the end of each evergreen task that deletes the clusters "alive". If you like the idea I will create a JIRA ticket.

andreaangiolillo:
Another idea is to wait to start an evergreen task until there are less that `n` clusters up on atlas.

gssbzn:
@andreaangiolillo right now one of the reasons we don't delete is because of new push to the branch, this interrupts the current build which leaves all the dangling clusters
Also deleting live clusters is dangerous right now as all e2e tests share the same project, which could lead you to deleting clusters of other unrelated test/pr

One thing I've discussed with @robcarlan-mlab and we can eventually move to this is for each tests to create its own project and run its own set of e2e in isolation but that may take more work than

andreaangiolillo:
Looking at this page `https://docs.atlas.mongodb.com/reference/atlas-limits/ `it seems the the max number of projects within an org is 250 which is quite safe.

gssbzn:
yes and no, there was a bug where if we were to delete a project created with an API key it would still count toward the project limits even after being deleted, this has been recently fixed so we could look into it

PR #304, opened 2020-07-28
CLOUDP-68114: Add CAS field for security certs

robcarlan-mlab:
## Proposed changes

_Jira ticket:_ CLOUDP-68114

Closes #[issue number]

## Checklist

## Further comments

Just outputs the plain certificate:

```
./bin/mongocli atlas security customerCerts create --casFile ~/.ssh/bad_pem.pem
Certificate successfully created.
```

```
./bin/mongocli atlas security customerCerts describe
-----BEGIN CERTIFICATE-----
....
bSQR/NrJDA==
-----END CERTIFICATE-----

```

### DIFF
--- a/internal/cli/atlas/security/customercerts/create.go
+++ b/internal/cli/atlas/security/customercerts/create.go
@@ -26,6 +26,10 @@ import (
 	"github.com/spf13/cobra"
 )
 
+const createTemplate = `CAS
+{{.Cas}}
+`
+
 type SaveOpts struct {
 	cli.GlobalOpts
 	store   store.X509CertificateConfSaver
@@ -52,7 +56,7 @@ func (opts *SaveOpts) Run() error {
 		return err
 	}
 
-	return output.Print(config.Default(), "", r)
+	return output.Print(config.Default(), createTemplate, r)
 }
 
 // mongocli atlas security customercerts create --projectId projectId --casFile /path/to/certificates.pem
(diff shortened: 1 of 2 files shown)


internal/cli/atlas/security/customercerts/create.go
```diff
@@ -26,6 +26,10 @@ import (
 	"github.com/spf13/cobra"
 )
 
+const createTemplate = `CAS
+{{.Cas}}
+`
+
```
colm-quinn, line 32: [q] What did this look like without the template?

I'm looking at the output fro `gpg` and wondering should it just be the key, so you can then pipe the output to file and use it directly somewhere. Right now you'd need to strip the `CAS` line at the start.

gssbzn: @robcarlan-mlab I'm with @colm-quinn in that I feel like the "human" output could be just straight up the certificate, no title, but that for the describe, for the create I think we should follow other commands with just a "success message"

@colm-quinn for context in this command (create) the user is providing the certificate so returning an answer with the same certificate the user provided is kind of a meh for me

colm-quinn: > @colm-quinn for context in this command (create) the user is providing the certificate so returning an answer with the same certificate the user provided is kind of a meh for me

Got it and yes agreed.

robcarlan-mlab: Yeah that makes sense to me too 👍

PR #332, opened 2020-08-04
CLOUDP-68633: Update GOLANGCI_VERSION

gssbzn:
## Proposed changes

_Jira ticket:_ CLOUDP-68633

## Checklist


e2e/atlas/clusters_test.go
```diff
@@ -16,22 +16,25 @@
 package atlas_test
 
 import (
+	"crypto/rand"
 	"encoding/json"
 	"fmt"
-	"math/rand"
```
gssbzn, line 21: gosec is now picky about it, we had no problem using this but changing to a "secure" option was not more complicated than using the safer option

e2e/atlas/clusters_test.go
```diff
@@ -16,22 +16,25 @@
...
 )
 
 func TestClusters(t *testing.T) {
-	r := rand.New(rand.NewSource(time.Now().UnixNano()))
-	clusterName := fmt.Sprintf("e2e-cluster-%v", r.Uint32())
+	n, err := rand.Int(rand.Reader, big.NewInt(1000))
+	if err != nil {
+		t.Fatalf("unexpected error: %v", err)
+	}
+	clusterName := fmt.Sprintf("e2e-cluster-%v", n)
```
colm-quinn, line 37: [np] This is updated in a bunch of places - maybe extract into a small helper method that generates the cluster name with a prefix as a parameter?

gssbzn: done

PR #336, opened 2020-08-04
CLOUDP-67937: mongocli iam project(s) apiKey(s)|apikey(s) delete

andreaangiolillo:
## Proposed changes

_Jira ticket:_ CLOUDP-67937

## Checklist

## Further comments

```
./bin/mongocli iam projects apikeys delete 5f297b4b81552a3d91711585
? Are you sure you want to delete: 5f297b4b81552a3d91711585 Yes
APIKey '5f297b4b81552a3d91711585' deleted
```


e2e/iam/orgs_api_keys_test.go
```diff
@@ -33,15 +33,14 @@ const (
 	apiKeysEntity = "apikeys"
 )
 
-func TestOrgAPIKeys(t *testing.T) {
+func TestOrgsAPIKeys(t *testing.T) {
```
gssbzn, line 36: [nit-grammar] only pluralize the last word, same for the file name
```suggestion
func TestOrgAPIKeys(t *testing.T) {
```

e2e/iam/orgs_api_keys_test.go
```diff
@@ -57,6 +56,7 @@ func TestOrgAPIKeys(t *testing.T) {
 		assert.NotEmpty(t, keys)
 	})
 
+	// This test must run first to grab the ID of the org to later describe
```
gssbzn, line 59: thanks

e2e/iam/projects_api_keys_test.go
```diff
@@ -0,0 +1,95 @@
...
+	var ID string
+
+	// This test must run first to grab the ID of the project to later describe
+	t.Run("Create", func(t *testing.T) {
+		cmd := exec.Command(cliPath, iamEntity,
+			projectEntity,
+			apiKeysEntity,
+			"create",
+			"--desc=e2e-test",
+			"--role=ORG_READ_ONLY")
```
gssbzn, line 45: interesting this works, this should be `GROUP_READ_ONLY`

e2e/iam/projects_api_keys_test.go
```diff
@@ -0,0 +1,95 @@
...
+		}
+		var keys []mongodbatlas.APIKey
+		if err := json.Unmarshal(resp, &keys); err != nil {
+			t.Fatalf("unexpected error: %v", err)
+		}
+		assert.NotEmpty(t, keys)
+	})
+
+	t.Run("Describe", func(t *testing.T) {
+		cmd := exec.Command(cliPath, iamEntity, projectEntity, apiKeysEntity, "describe", ID)
```
gssbzn, line 75: this command doesn't exists, how is this working?

andreaangiolillo: I am not sure about this. I run this test in my local machine and I got `nil` as response 🤷‍♂️

gssbzn: I tried running `mongocli iam projects apiKeys describe asa` and it doesn't return an exit status of error it just prints usage 🤔  I wonder if we should start validating this because it looks like we don't have tests if we accidentally delete a command

andreaangiolillo: yes, I think is a wise thing to do. Maybe instead of validating the output we can just use `assert.NotEmpty `

```
// NotEmpty asserts that the specified object is NOT empty.  I.e. not nil, "", false, 0 or either
// a slice or a channel with len == 0.
```

PR #341, opened 2020-08-05
CLOUDP-68758: Include API keys in creation message

gssbzn:
## Proposed changes

_Jira ticket:_ CLOUDP-68758

## Checklist

## Further comments

When we create an API key this is the only time we get to see the private key, so we need to show it


internal/cli/projects/apikeys/assign.go
```diff
@@ -11,57 +11,57 @@
...
-	fmt.Println("Successfully updated APIKey.")
-	return nil
+	return output.Print(config.Default(), updateTemplate, nil)
 }
 
-// mongocli iam project(s) apiKey(s)|apikey(s) update <ID> [--role role][--projectId projectId]
-func UpdateBuilder() *cobra.Command {
-	opts := new(UpdateOpts)
+// mongocli iam project(s) apiKey(s)|apikey(s) assign <ID> [--role role][--projectId projectId]
+func AssignBuilder() *cobra.Command {
```
gssbzn, line 59: I saw this command is actually to assign the key to a project so make this clearer for users

internal/cli/projects/apikeys/assign.go
```diff
@@ -11,57 +11,57 @@
...
+	opts := new(AssignOpts)
 	cmd := &cobra.Command{
-		Use:     "update <ID>",
-		Aliases: []string{"updates"},
-		Args:    cobra.ExactArgs(1),
-		Short:   description.ProjectOrganizationsAPIKey,
+		Use:   "assign <ID>",
```
colm-quinn, line 62: [q] Do we need to keep an alias to the old update command here, so users of the current version?

gssbzn: I thought about it but we have not released this so I was ok with just a full rename

colm-quinn: sgtm, thanks for the clarification.

PR #343, opened 2020-08-07
CLOUDP-68878: split the description package

gssbzn:
## Proposed changes

_Jira ticket:_ CLOUDP-68878

## Checklist

## Further comments

This is a proposal to start splitting the big description package closer to where the descriptions are needed, when we started we only had the cli package and as we started to add commands we started to group them in packages, I think is time to move from the big description file to smaller ones closer to where they are used


gssbzn:
@colm-quinn this was also a fun exercise as it exposed a couple of commands with missing descriptions or with copy/paste descriptions that did not match, this is another reason I like this, making it to its own package will let us catch this quicker and ina more obvious way

PR #347, opened 2020-08-10
CLOUDP-69013: Automatically generate command aliases

gssbzn:
## Proposed changes

_Jira ticket:_ CLOUDP-69013

## Checklist

### DIFF
--- a/internal/cli/global_opts.go
+++ b/internal/cli/global_opts.go
@@ -16,9 +16,11 @@ package cli
 
 import (
 	"fmt"
+	"strings"
 
 	"github.com/mongodb/mongocli/internal/config"
 	"github.com/mongodb/mongocli/internal/validate"
+	"github.com/tangzero/inflector"
 )
 
 type GlobalOpts struct {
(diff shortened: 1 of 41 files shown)


internal/cli/global_opts.go
```diff
@@ -87,3 +89,27 @@ func (opts *GlobalOpts) PreRunEOrg(cbs ...cmdOpt) error {
 func DeploymentStatus(baseURL, projectID string) string {
 	return fmt.Sprintf("Changes are being applied, please check %sv2/%s#deployment/topology for status\n", baseURL, projectID)
 }
+
+// GenerateAliases return aliases for use such that they are:
+// a version all lower case, a version with dashes, a singular versions with the same rules.
+func GenerateAliases(use string, extra ...string) []string {
```
gssbzn, line 95: new method to handle singulars, lowercase and dash version of the commands

colm-quinn: Automagical parameter creation - I love it!

colm-quinn: [q] What stops a clash here with multiple commands generating the same alias?

gssbzn: if they generate the same alias means they have the same `usage` and let me tell you it doesn't look pretty
```
 cloud-manager MongoDB Cloud Manager operations.
  ops-manager   Ops Manager operations.
  cloud-manager Organization and projects operations.
```
no error and cobra will call the first command with that name

colm-quinn: *cry* I mean, the current solution is no worse than what's in place right now, but that isn't great :/ Feels like we should at least build a list of aliases in memory as we init and exit with a clear error rather than have it run like that.

gssbzn: this is tricky, having same usage/aliases is only a problem for sibling commands, we already have duplication and is expected (ops manager clusters vs atlas clusters)
What you actually need to do is construct a list of sibling commands and check they don't share usage an aliases
I think the risk of this happening is quite low but we can explore it later anyway

PR #370, opened 2020-08-18
CLOUDP-69182: mongocli atlas networking peering create aws

andreaangiolillo:
## Proposed changes

_Jira ticket:_ CLOUDP-69182

## Checklist

## Further comments

```
./bin/mongocli atlas networking peering create aws --accountId 228363220022 --region EU_WEST_1 --vpcId vpc-7b64b41e --routeTableCidrBlock 172.31.0.0/16 --atlasCidrBlock 172.31.0.0/16

Network peering connection '5f3ba92f3f7e890dc369a6a5' created.
```

Screenshot 2020-08-18 at 12 11 33


internal/cli/atlas/networking/peering/create/aws.go
```diff
@@ -0,0 +1,145 @@
...
+	"github.com/mongodb/mongocli/internal/config"
+	"github.com/mongodb/mongocli/internal/flag"
+	"github.com/mongodb/mongocli/internal/output"
+	"github.com/mongodb/mongocli/internal/store"
+	"github.com/mongodb/mongocli/internal/usage"
+	"github.com/spf13/cobra"
+	atlas "go.mongodb.org/atlas/mongodbatlas"
+)
+
+type AwsOpts struct {
```
gssbzn, line 30: Go style prefers all caps for this kind of names
```suggestion
type AWSOpts struct {
```

gssbzn: to be clear, just in case, go style is AWS, all caps or aws, all lower case, no middle ground

internal/cli/atlas/networking/peering/create/azure.go
```diff
@@ -123,12 +125,12 @@ func AzureBuilder() *cobra.Command {
 		},
 	}
 
-	cmd.Flags().StringVar(&opts.directoryID, flag.DirectoryID, "", usage.ClusterName)
-	cmd.Flags().StringVar(&opts.subscriptionID, flag.SubscriptionID, "", usage.Database)
-	cmd.Flags().StringVar(&opts.resourceGroup, flag.ResourceGroup, "", usage.Collection)
-	cmd.Flags().StringVar(&opts.vNetName, flag.VNet, "", usage.DateField)
-	cmd.Flags().StringVar(&opts.region, flag.Region, "", usage.DateField)
-	cmd.Flags().StringVar(&opts.atlasCIDRBlock, flag.AtlasCIDRBlock, "", usage.DateField)
+	cmd.Flags().StringVar(&opts.directoryID, flag.DirectoryID, "", usage.DirectoryID)
```
gssbzn, line 128: oops and thanks

internal/flag/flags.go
```diff
@@ -148,4 +148,8 @@ const (
 	SubscriptionID                  = "subscriptionId"                  // SubscriptionID flag
 	ResourceGroup                   = "resourceGroup"                   // ResourceGroup flag
 	VNet                            = "vnet"                            // VNet flag
+	AccountID                       = "accountId"                       // AccountID flag
+	ContainerID                     = "containerID"                     // ContainerID flag
```
gssbzn, line 152: we don't follow go rules for flags
```suggestion
	ContainerID                     = "containerId"                     // ContainerID flag
```

internal/cli/atlas/networking/peering/create/aws.go
```diff
@@ -0,0 +1,145 @@
...
+	}
+	r, err := opts.store.CreatePeeringConnection(opts.ConfigProjectID(), opts.newPeer(container.ID))
+	if err != nil {
+		return err
+	}
+	return output.Print(config.Default(), createTemplate, r)
+}
+
+func (opts *AwsOpts) containerExists() (*atlas.Container, error) {
+	r, err := opts.store.AllContainers(opts.ConfigProjectID(), nil)
```
gssbzn, line 68: [q] Why all instead of filtering by AWS?

internal/cli/atlas/networking/peering/create/aws.go
```diff
@@ -0,0 +1,145 @@
...
+		PreRunE: func(cmd *cobra.Command, args []string) error {
+			return opts.PreRunE(opts.initStore)
+		},
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return opts.Run()
+		},
+	}
+
+	cmd.Flags().StringVar(&opts.accountID, flag.AccountID, "", usage.AccountID)
+	cmd.Flags().StringVar(&opts.containerID, flag.ContainerID, "", usage.ContainerID)
```
gssbzn, line 131: you shouldn't need this flag

internal/cli/atlas/networking/peering/create/azure.go
```diff
@@ -70,7 +72,7 @@ func (opts *AzureOpts) containerExists() (*atlas.Container, error) {
 		return nil, err
 	}
 	for i := range r {
-		if r[i].Region == opts.region {
+		if strings.EqualFold(r[i].Region, opts.region) {
```
gssbzn, line 75: rather than doing this we should do a normalization of the input, ie, we uppercase the input, which is a once off operation, see my comment in the aws command

internal/flag/flags.go
gssbzn: also we may not need this flag at all

internal/cli/atlas/networking/peering/create/aws.go
```diff
@@ -0,0 +1,145 @@
...
+}
+
+func (opts *AwsOpts) containerExists() (*atlas.Container, error) {
+	r, err := opts.store.AllContainers(opts.ConfigProjectID(), nil)
+	if err != nil {
+		return nil, err
+	}
+	for i := range r {
+		region := strings.ToUpper(opts.region)
+		region = strings.ReplaceAll(region, "-", "_")
```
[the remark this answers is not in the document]
andreaangiolillo: I am not using the same arguments for `ReplaceAll`. Indeed,
Container: Region = `EU_WEST_1`
Peer: Region = `eu-west-1`

Those two lines allow us to have only one flag `region` instead of creating also `AccepterRegionName`.
Given that `ReplaceAll` is called in two different place with different arguments I wouldn’t create a function for it.

internal/flag/flags.go
andreaangiolillo: I agree with you, in the doc is listed as Required but it is not true 🤷‍♂️

internal/cli/atlas/networking/peering/create/aws.go
andreaangiolillo: If omitted, Atlas sets this parameter to AWS. https://docs.atlas.mongodb.com/reference/api/vpc-get-containers-list/

internal/flag/flags.go
gssbzn: it is but ware being sneaky about it, there's this rule where there can only be one container per region per provider, so what we are doing is checking if it exists if not we create it and use that id

internal/cli/atlas/networking/peering/create/aws.go
gssbzn: yes, but this method is for https://docs.atlas.mongodb.com/reference/api/vpc-get-containers-list-all/
you need `ContainersByProvider` also let's be explicit about it to avoid the chance that the behaviour of the API changes later

internal/cli/atlas/networking/peering/create/aws.go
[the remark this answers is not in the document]
gssbzn: lines 73-74
```go
region := strings.ToUpper(opts.region)
region = strings.ReplaceAll(region, "-", "_")
```
lines  83-84
```go
region := strings.ToUpper(opts.region)
region = strings.ReplaceAll(region, "-", "_")
```

They look the same to me

internal/cli/atlas/networking/peering/create/azure.go
gssbzn: also wath out here with normalization as for example
Azure: centralus
Atlas: US_CENTRAL

internal/cli/atlas/networking/peering/create/aws.go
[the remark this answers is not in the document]
andreaangiolillo: ops you’re right 😅

internal/cli/atlas/networking/peering/create/aws_test.go
```diff
@@ -0,0 +1,106 @@
...
+			CreatePeeringConnection(opts.ProjectID, request).
+			Return(&mongodbatlas.Peer{}, nil).
+			Times(1)
+		if err := opts.Run(); err != nil {
+			t.Fatalf("Run() unexpected error: %v", err)
+		}
+	})
+}
+
+func TestNormalizeAtlasRegion(t *testing.T) {
```
gssbzn, line 83: let's use a [table based](https://dave.cheney.net/2019/05/07/prefer-table-driven-tests) test here
if you try to generate the test with IDEA, IDEA already gives you the skeleton for this type of test

internal/store/peering_connections.go
```diff
@@ -30,6 +30,7 @@ type PeeringConnectionLister interface {
 
 type PeeringConnectionCreator interface {
 	AzureContainers(string) ([]atlas.Container, error)
+	AWSContainers(string) ([]atlas.Container, error)
 	CreateContainer(string, *atlas.Container) (*atlas.Container, error)
 	CreatePeeringConnection(string, *atlas.Peer) (*atlas.Peer, error)
 }
```
gssbzn, line 36: I'm already doing this for GCP but let's split this
`type AzurePeeringConnectionCreator interface {` and `type AWSPeeringConnectionCreator interface {` and this way they don't share the List method

internal/usage/usage.go
```diff
@@ -129,6 +129,14 @@ const (
 	WhitelistIPEntry                = "IP address to be whitelisted for a given API key."
 	WhitelistCIDREntry              = "Whitelist entry in CIDR notation to be added for a given API key."
 	PrivateEndpointID               = "Unique identifier of the AWS PrivateLink connection."
+	AccountID                       = "Account ID of the owner of the peer VPC."
+	RouteTableCidrBlock             = "Peer VPC CIDR block or subnet."
+	VpcID                           = "Unique identifier of the peer VPC."
+	AtlasCIDRBlock                  = "CIDR block that Atlas uses for your clusters."
+	VnetName                        = "Name of your Azure VNet."
```
gssbzn, line 136: [nit] 
```suggestion
	VNet                        = "Name of your Azure VNet."
```

internal/store/containers.go
```diff
@@ -64,6 +64,24 @@ func (s *Store) AzureContainers(projectID string) ([]atlas.Container, error) {
 	}
 }
 
+// AWSContainers encapsulates the logic to manage different cloud providers
+func (s *Store) AWSContainers(projectID string) ([]atlas.Container, error) {
```
gssbzn, line 68: thanks

internal/cli/atlas/networking/peering/create/aws_test.go
```diff
@@ -0,0 +1,104 @@
...
+	type test struct {
+		input string
+		want  string
+	}
+
+	tests := []test{
```
gssbzn, line 90: _chef kiss_

internal/cli/atlas/networking/peering/create/aws_test.go
```diff
@@ -0,0 +1,104 @@
...
+		{input: "eu-west-1", want: "EU_WEST_1"},
+		{input: "eu_west-1", want: "EU_WEST_1"},
+		{input: "eu-west_1", want: "EU_WEST_1"},
+		{input: "EU_WEST_1", want: "EU_WEST_1"},
+		{input: "eu_west_1", want: "EU_WEST_1"},
+	}
+
+	for _, tc := range tests {
+		got := normalizeAtlasRegion(tc.input)
+		if !reflect.DeepEqual(tc.want, got) {
```
gssbzn, line 100: you can safely really on `==` since they are strings but let's not make you wait for another evergreen run

PR #379, opened 2020-08-20
CLOUDP-69793: [mongocli] Improve output handling

gssbzn:
## Proposed changes

_Jira ticket:_ CLOUDP-69793

## Checklist

## Further comments

This is somewhat of a big change but mostly do to the number of files I had to update, the main thing here is to have a central point where we can control output, specially as I want to add things like a spinner and colors, where we need to check if the terminal is a terminal or a file

### DIFF
--- a/internal/cli/events/list.go
+++ b/internal/cli/events/list.go
@@ -19,7 +19,6 @@ import (
 	"github.com/mongodb/mongocli/internal/cli"
 	"github.com/mongodb/mongocli/internal/config"
 	"github.com/mongodb/mongocli/internal/flag"
-	"github.com/mongodb/mongocli/internal/output"
 	"github.com/mongodb/mongocli/internal/store"
 	"github.com/mongodb/mongocli/internal/usage"
 	"github.com/spf13/cobra"
@@ -28,6 +27,7 @@ import (
 
 type ListOpts struct {
 	cli.ListOpts
+	cli.OutputOpts
 	orgID     string
 	projectID string
 	eventType []string
@@ -61,7 +61,7 @@ func (opts *ListOpts) Run() error {
 		return err
 	}
 
-	return output.Print(config.Default(), listTemplate, r)
+	return opts.Print(r)
 }
 
 func (opts *ListOpts) newEventListOptions() *atlas.EventListOptions {
@@ -79,6 +79,7 @@ func (opts *ListOpts) newEventListOptions() *atlas.EventListOptions {
 // mongocli atlas event(s) list [--projectId projectId] [--orgId orgId] [--page N] [--limit N] [--minDate minDate] [--maxDate maxDate]
 func ListBuilder() *cobra.Command {
 	opts := &ListOpts{}
+	opts.Template = listTemplate
 	cmd := &cobra.Command{
 		Use:     "list",
 		Short:   listEvents,
@@ -91,6 +92,7 @@ func ListBuilder() *cobra.Command {
 			if opts.orgID == "" && opts.projectID == "" {
 				return fmt.Errorf("--%s or --%s must be set", flag.ProjectID, flag.OrgID)
 			}
+			opts.OutWriter = cmd.OutOrStdout()
 			return opts.initStore()
 		},
 		RunE: func(cmd *cobra.Command, args []string) error {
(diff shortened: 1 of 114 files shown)


internal/cli/output_opts.go
```diff
@@ -0,0 +1,119 @@
...
+
+const (
+	jsonFormat     = "json"
+	goTemplate     = "go-template"
+	goTemplateFile = "go-template-file"
+)
+
+var templateFormats = []string{goTemplate, goTemplateFile}
+
+type OutputOpts struct {
```
gssbzn, line 38: This is the new add, and now every command that wants to print should do so with the help of this struct

internal/templatewritter/templatewritter.go
```diff
@@ -12,28 +12,15 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-package output
+package templatewritter
```
robcarlan-mlab, line 15: package name and filename should be `templatewritter` -> `templatewriter `

internal/jsonwritter/jsonwritter.go
```diff
@@ -12,19 +12,25 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-package json
+package jsonwritter
```
robcarlan-mlab, line 15: `jsonwritter` -> `jsonwriter`

internal/cli/events/list.go
```diff
@@ -79,6 +79,7 @@ func (opts *ListOpts) newEventListOptions() *atlas.EventListOptions {
 // mongocli atlas event(s) list [--projectId projectId] [--orgId orgId] [--page N] [--limit N] [--minDate minDate] [--maxDate maxDate]
 func ListBuilder() *cobra.Command {
 	opts := &ListOpts{}
+	opts.Template = listTemplate
```
robcarlan-mlab, line 82: [q} Can these be written in the same way as https://github.com/mongodb/mongocli/pull/379/files#diff-7dfd7a5005571cf7b5e5368ae7a939bcR62, to make it more consistent?

internal/cli/output_opts.go
```diff
@@ -0,0 +1,119 @@
...
+		return err
+	}
+	if t != "" {
+		return templatewritter.Print(opts.ConfigWriter(), t, v)
+	}
+	_, err = fmt.Fprintln(opts.ConfigWriter(), v)
+	return err
+}
+
+func (opts *OutputOpts) parseTemplate() (string, error) {
```
robcarlan-mlab, line 99: is this allowing user defined templates? Might be worth adding a comment

internal/cli/output_opts.go
```diff
@@ -0,0 +1,119 @@
...
+	return false
+}
+
+func (opts *OutputOpts) IsCygwinTerminal() bool {
+	if f, isFile := opts.OutWriter.(*os.File); isFile {
+		return isatty.IsCygwinTerminal(f.Fd())
+	}
+
+	return false
+}
```
andreaangiolillo, line 89: 💯

internal/cli/events/list.go
gssbzn: We don't have a call to our own `PreRunE` func, so it's a bit ugly since that function is thought to be called with that
as in, to use it here it would be something like `opts.InitOutput(cmd.OutOrStdout(), listTemplate)(),`

I'm thinking we need to make `PreRunE` and pass the validations functions instead of relaying on it doing the project id validation but that will be another day

For this le me know if switching to  `_ = opts.InitOutput(cmd.OutOrStdout(), listTemplate)()` makes sense (also ignore error as that never return an error, is just there to fulfil the signature of `PreRunE`

gssbzn (commented):
I think I addressed your comments, I'll wait to see if all variants pass

andreaangiolillo (approved):
I have just spotted a bunch of stuff that can be changed in another PR. I can’t wait to see how the output will change in the next releases! 🚀  LGTM!! 💯

internal/cli/events/list.go
robcarlan-mlab: Ah got it, I think that last suggestion makes sense, imo (but I wouldn't block on it if you want to change the structure soon)

internal/cli/output_opts.go
gssbzn: added

PR #408, opened 2020-09-03
CLOUDP-70881: Implementing e2e tests for team and users

andreaangiolillo:
## Proposed changes

_Jira ticket:_ CLOUDP-70881

## Checklist


internal/cli/iam/projects/teams/add.go
```diff
@@ -79,7 +79,7 @@ func AddBuilder() *cobra.Command {
 
 	cmd.Flags().StringSliceVar(&opts.roles, flag.Role, []string{}, usage.TeamRole)
 
-	cmd.Flags().StringVar(&opts.OrgID, flag.ProjectID, "", usage.ProjectID)
+	cmd.Flags().StringVar(&opts.ProjectID, flag.ProjectID, "", usage.ProjectID)
```
andreaangiolillo, line 82: the importance of testing 😬

gssbzn: :P 
I actually have a PR in the works to start testing what flags we set to a command

colm-quinn (dismissed):
Mostly doing a parameter check on the method calls to check they are doing what the description says their doing. LGTM.

e2e/iam/helper_test.go
```diff
@@ -82,3 +84,127 @@ func deleteOrgAPIKey(id string) error {
...
+		return "", "", err
+	}
+
+	var users mongodbatlas.AtlasUsersResponse
+	if err := json.Unmarshal(resp, &users); err != nil {
+		return "", "", err
+	}
+
+	if len(users.Results) < userIndex {
+		return "", "", fmt.Errorf(`got="%s" want= len(users.Results) > "%d"`, users.Results, userIndex)
```
gssbzn, line 206: yes on the error but the error message can be better 
```suggestion
		return "", "", fmt.Errorf("invalid index %d for %d users", n, len(users.Results))
```

internal/cli/iam/teams/describe.go
```diff
@@ -83,6 +83,9 @@ func DescribeBuilder() *cobra.Command {
 		Example: `  
   Describe a user by ID
   $ mongocli iam users describe --id teamId --orgId orgId <id>
+
+  Describe a user by Name
```
gssbzn, line 87:
```suggestion
  Describe a team by Name
```

same above

gssbzn (changes requested):
This looks great, those helpers are going to come handy latter.
Left a couple of small nits, and one major point is that you are missing `assert` style test with standard go assertions, please stick to one and avoid mixing them, this is on all tests but I only commented the first couple of files

e2e/iam/helper_test.go
```diff
@@ -82,3 +84,127 @@ func deleteOrgAPIKey(id string) error {
...
+	resp, err := cmd.CombinedOutput()
+
+	if err != nil {
+		return "", "", err
+	}
+
+	var users mongodbatlas.AtlasUsersResponse
+	if err := json.Unmarshal(resp, &users); err != nil {
+		return "", "", err
+	}
```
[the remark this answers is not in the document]
andreaangiolillo: it doesn’t because `atlasUser` and `User` have the "same" fields but yes, this may change in the future so I am going to check if we are running the test against atlas or cloudManager. I hope this `if config.Service() == config.CloudService ` is enough

e2e/iam/helper_test.go
```diff
@@ -82,3 +84,127 @@ func deleteOrgAPIKey(id string) error {
...
+
+	if len(users.Results) < userIndex {
+		return "", "", fmt.Errorf(`got="%s" want= len(users.Results) > "%d"`, users.Results, userIndex)
+	}
+
+	return users.Results[userIndex].Username, users.Results[userIndex].ID, nil
```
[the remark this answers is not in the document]
andreaangiolillo: the reason why is that we have two different user objects: `AtlasUser` and `User`. So if I return `interface{}` I wouldn't be able to select  object's fields.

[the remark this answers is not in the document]
gssbzn: you can select fields, you just need to cast it first, we do this in a couple of places

e2e/iam/project_teams_test.go
```diff
@@ -0,0 +1,162 @@
...
+		if err = json.Unmarshal(resp, &roles); err != nil {
+			t.Fatalf("unexpected error: %v", err)
+		}
+
+		assert.Len(t, roles, 1)
+
+		role := roles[0]
+
+		assert.Equal(t, role.TeamID, teamID)
+		assert.Len(t, role.RoleNames, 2)
```
gssbzn, line 121: [nit] when doing multiple asserts the docs recommend to do an initialization of assert and then reference that struct 
```go
a := assert.New(t)
a.Len(roles, 1)
a.Equal(role.TeamID, teamID)
a.Len(role.RoleNames, 2)

File: e2e/iam/project_teams_test.go [hunk 7]
```diff
@@ -0,0 +1,162 @@
...
+			projectEntity,
+			teamsEntity,
+			"delete",
+			teamID,
+			"--force",
+			"--projectId",
+			projectID)
+		cmd.Env = os.Environ()
+		resp, err := cmd.CombinedOutput()
+		assert.NoError(t, err, string(resp))
```
gssbzn, line 160: Please add an assertion for the message, to confirm the output, we do this for all our `delete`s and has help us to discover bugs

e2e/iam/project_teams_test.go
```diff
@@ -0,0 +1,162 @@
...
+			projectID,
+			"-o=json")
+		cmd.Env = os.Environ()
+		resp, err := cmd.CombinedOutput()
+		assert.NoError(t, err, string(resp))
+
+		var teams mongodbatlas.TeamsAssigned
+		if err = json.Unmarshal(resp, &teams); err != nil {
+			t.Fatalf("unexpected error: %v", err)
+		}
```
gssbzn, line 78: [nit] any reason you decided to keep this with this still and not use `assert.NoError(t, err)`?
if you decide to use the assert approach please take into account that you should probably start using an assert variable (see other comment about this) and approach I've done on other test is to use and `if` here

```go
a := assert.New(t)
if err := json.Unmarshal(resp, &teams); a.NoError(err) {
	// loop to find value
	a.True(found)
}
```

gssbzn: this applies to all the `err = json.Unmarshal()` you left around

gssbzn (changes requested):
couple of small nits, mostly around improving the usage of assert but this is almost done

e2e/iam/project_teams_test.go
```diff
@@ -0,0 +1,171 @@
...
+
+		found := false
+		for _, team := range teams.Results {
+			if team.TeamID == teamID {
+				found = true
+				break
+			}
+		}
+
+		a.True(found)
```
gssbzn, line 96: duplicated

e2e/iam/project_teams_test.go
```diff
@@ -0,0 +1,171 @@
...
+		resp, err := cmd.CombinedOutput()
+		a := assert.New(t)
+		a.NoError(err, string(resp))
+
+		var roles []mongodbatlas.TeamRoles
+		if err = json.Unmarshal(resp, &roles); a.NoError(err) {
+			a.Len(roles, 1)
+
+			role := roles[0]
+			a.Equal(role.TeamID, teamID)
```
gssbzn, line 115: first param is expected and second is actual
```suggestion
			a.Equal(teamID, role.TeamID)
```

e2e/iam/project_teams_test.go
```diff
@@ -0,0 +1,161 @@
...
+			a.Len(role.RoleNames, 2)
+			for _, roleName := range role.RoleNames {
+				a.Contains([]string{roleName1, roleName2}, roleName)
+			}
```
gssbzn, line 119: you can rewrite this as
```suggestion
			a.ElementsMatch([]string{roleName1, roleName2},  role.RoleNames)
```
This ignores order

gssbzn (commented):
one couple of small assert changes

gssbzn (approved):
LGTM

PR #424, opened 2020-09-08
CLOUDP-72402: Add e2e tests for atlas integrations

andreaangiolillo:
## Proposed changes

_Jira ticket:_ CLOUDP-72402

Closes #[issue number]

## Checklist

### DIFF
--- a/e2e/atlas/integrations_test.go
+++ b/e2e/atlas/integrations_test.go
@@ -1,0 +1,307 @@
+// Copyright 2020 MongoDB Inc
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+// +build e2e atlas,generic
+
+package atlas_test
+
+import (
+	"encoding/json"
+	"fmt"
+	"os"
+	"os/exec"
+	"testing"
+
+	"github.com/mongodb/mongocli/e2e"
+	"github.com/stretchr/testify/assert"
+	"go.mongodb.org/atlas/mongodbatlas"
+)
+
+const (
+	integrationsEntity = "integrations"
+	datadogEntity      = "DATADOG"
+	flowdockEntity     = "FLOWDOCK"
+	newRelicEntity     = "NEW_RELIC"
+	opsGenieEntity     = "OPS_GENIE"
+	pagerDutyEntity    = "PAGER_DUTY"
+	victorOpsEntity    = "VICTOR_OPS"
+	webhookEntity      = "WEBHOOK"
+)
+
+func TestIntegrations(t *testing.T) {
+	n, err := e2e.RandInt(255)
+	if err != nil {
+		t.Fatalf("unexpected error: %v", err)
+	}
+
+	key := "51c0ef87e9951c3e147accf0e12" + n.String()
+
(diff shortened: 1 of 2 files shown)


e2e/atlas/integrations_test.go
```diff
@@ -0,0 +1,307 @@
...
+			victorOpsEntity,
+			"--apiKey",
+			key,
+			"-o=json")
+		cmd.Env = os.Environ()
+		resp, err := cmd.CombinedOutput()
+
+		a := assert.New(t)
+		a.NoError(err, string(resp))
+	})
```
andreaangiolillo, line 222: VICTOR_OPS is not inside the returned list. I have created this task to investigate this behaviour [CLOUDP-71439](https://jira.mongodb.org/browse/CLOUDP-71439)

e2e/atlas/integrations_test.go
```diff
@@ -0,0 +1,306 @@
...
+	datadogEntity      = "DATADOG"
+	flowdockEntity     = "FLOWDOCK"
+	newRelicEntity     = "NEW_RELIC"
+	opsGenieEntity     = "OPS_GENIE"
+	pagerDutyEntity    = "PAGER_DUTY"
+	victorOpsEntity    = "VICTOR_OPS"
+	webhookEntity      = "WEBHOOK"
+)
+
+func TestIntegrations(t *testing.T) {
```
gssbzn, line 41: one major thing here, we may get false positives from running this test in parallel given there can only be one of these integrations per project and we're always running with the same project
I think we should reuse your helper to create a new project here and limit the scope of the test and avoid false possitives

e2e/atlas/integrations_test.go
```diff
@@ -0,0 +1,306 @@
...
+		var thirdPartyIntegrations mongodbatlas.ThirdPartyIntegrations
+		if err := json.Unmarshal(resp, &thirdPartyIntegrations); a.NoError(err) {
+			found := false
+			services := thirdPartyIntegrations.Results
+			for i := range services {
+				if services[i].Type == datadogEntity {
+					found = true
+					break
+				}
+			}
```
gssbzn, line 77: I think we should create a helper `integrationExists(name string) bool` and reuse it for these create tests

gssbzn (changes requested):
We may face some false positives here given we can only have one integration type per project so we should try to scope this test to run on its own new project

andreaangiolillo:
I am not sure what this error means
```
e2e/atlas/helper_test.go:145:97: integrationExists - result 0 (bool) is never used (unparam)
func integrationExists(name string, thirdPartyIntegrations mongodbatlas.ThirdPartyIntegrations) bool {
```

e2e/atlas/integrations_test.go
```diff
@@ -100,15 +109,7 @@ func TestIntegrations(t *testing.T) {
...
-			found := false
-			services := thirdPartyIntegrations.Results
-			for i := range services {
-				if services[i].Type == flowdockEntity {
-					found = true
-					break
-				}
-			}
-			a.True(found)
+			integrationExists(flowdockEntity, thirdPartyIntegrations)
```
gssbzn, line 112: the error is you never use the return value of the function, here you're missing a `a.True(integrationExists(...))`

e2e/atlas/integrations_test.go
```diff
@@ -50,6 +50,19 @@ func TestIntegrations(t *testing.T) {
 	if err != nil {
 		t.Fatalf("unexpected error: %v", err)
 	}
+
+	projectName := fmt.Sprintf("e2e-proj-%v", n)
```
gssbzn, line 54: [nit] can we rename this, sometimes test left things around in atlas and this way we can better diagnose where are we leaking data
```suggestion
	projectName := fmt.Sprintf("e2e-integration-proj-%v", n)
```

gssbzn (changes requested):
One small nit and let's fix that linting error

gssbzn (approved):
LGTM thanks for adding these tests

PR #435, opened 2020-09-09
CLOUDP-72054: Investigate why we are getting an error when we do not set --secret to configure webhook

andreaangiolillo:
## Proposed changes

_Jira ticket:_ CLOUDP-72054

## Checklist

## Further comments

This property is optional according to documentation and the atlas UI so this is a problem of the endpoint. I am going to create a ticket for core/iam to verify this behaviour.
In the meantime, I am going to set --secret as required.


andreaangiolillo:
we already set --secret in the e2e test

gssbzn (approved):
LGTM

PR #443, opened 2020-09-11
CLOUDP-72734: add e2e tests for LDAP

andreaangiolillo:
## Proposed changes

_Jira ticket:_ CLOUDP-72734

## Checklist


andreaangiolillo:
@gssbzn

build/ci/evergreen.yml
```diff
@@ -466,6 +466,24 @@ tasks:
           MCLI_PUBLIC_API_KEY: ${atlas_public_api_key}
           MCLI_SERVICE: cloud
           E2E_TAGS: atlas,clusters
+  # LDAP Configuration depends on a cluster running
```
gssbzn, line 469: [q] do they?

e2e/atlas/ldap_test.go
```diff
@@ -0,0 +1,183 @@
...
+	"testing"
+
+	"github.com/mongodb/mongocli/e2e"
+	"github.com/stretchr/testify/assert"
+	"go.mongodb.org/atlas/mongodbatlas"
+)
+
+const (
+	PENDING  = "PENDING"
+	HOSTNAME = "2.tcp.ngrok.io"
```
gssbzn, line 31: [q] why are we exporting this? and also in go there's no all CAPS naming convention (unless they are an acronym like LDAP)

gssbzn, line 31: if this doesn't need to be valid I'd rather use a random string

e2e/atlas/helper_test.go
```diff
@@ -81,6 +84,10 @@ func getHostnameAndPort() (string, error) {
 }
 
 func deployCluster() (string, error) {
+	return deployClusterFromProject("")
+}
+
+func deployClusterFromProject(projectID string) (string, error) {
```
gssbzn, line 90: [nit] couple of things here, first `deployClusterFromProject` seems unused, as in you only use it here, so it seems as an unnecessary change. I guess your goal was to create a new project for ldap test and then deploy for the new project, if you still plan to do that then this method should be `deployClusterForProject` or `deployClusterWithProjectID` and be used, I'd also change `deployCluster` to `deployClusterWithDefaultProject` to avoid confusion later on
Also non blocking this usage where and empty string defaults to the default project (`deployClusterFromProject("")`) should at least be documented, the name of the method and this assumption could be confusing later on so I'd rather have a nutral method shared by both `WithDefaultProject` and `WithProjectID` and let `WithProjectID` fail if no id is given

e2e/atlas/helper_test.go
```diff
@@ -116,11 +135,27 @@ func deployCluster() (string, error) {
 }
 
 func deleteCluster(clusterName string) error {
+	return deleteClusterForProject(clusterName, "")
+}
+
+func deleteClusterForProject(clusterName, projectID string) error {
```
gssbzn, line 141: similar comment to above

gssbzn (changes requested):
This looks great, I have a couple of question or suggestions before moving forward

build/ci/evergreen.yml
andreaangiolillo: yes, they do.
Screenshot 2020-09-14 at 10 57 09

gssbzn: weird but I'll take it 👍

e2e/atlas/ldap_test.go
```diff
@@ -27,8 +27,8 @@ import (
 )
 
 const (
-	PENDING  = "PENDING"
-	HOSTNAME = "2.tcp.ngrok.io"
+	pending  = "pending"
+	hostname = "localhost"
```
gssbzn, line 31: 👍

gssbzn (dismissed):
I've scheduled the e2e test for ldap since they won't run by default on the PR suite, good to merge if it passes on evergreen

e2e/atlas/ldap_test.go
```diff
@@ -27,8 +27,8 @@ import (
 )
 
 const (
-	PENDING  = "PENDING"
-	HOSTNAME = "2.tcp.ngrok.io"
+	pending  = "pending"
```
andreaangiolillo, line 30: Actually here there is a mistake:
```suggestion
	pending  = "PENDING"
```

gssbzn: when you submit this change please schedule the e2e test for ldap on evergreen

andreaangiolillo: done

gssbzn (approved):
LGTM

PR #470, opened 2020-09-30
CLOUDP-70062: mongocli ops-manager admin backup blockstore(s) lists

andreaangiolillo:
## Proposed changes

_Jira ticket:_ CLOUDP-70062

## Checklist

## Further comments

```
  ./bin/mongocli om admin backup blockstore list -P ops -o json
{
  "links": [
    {
      "rel": "self",
      "href": "http://localhost:8080/api/public/v1.0/admin/backup/snapshot/mongoConfigs?pageNum=1\u0026itemsPerPage=100"
    }
  ],
  "results": [
    {
      "id": "blockstore",
      "uri": "mongodb://test1:test1@localhost:22222",
      "ssl": false,
      "assignmentEnabled": true,
      "encryptedCredentials": false,
      "loadFactor": 1
    }
  ],
  "totalCount": 1
}

```
```
./bin/mongocli om admin backup blockstore list -P ops
ID           URI                                     SSL     LOAD FACTOR
blockstore   mongodb://test1:test1@localhost:22222   false   1
```


internal/cli/opsmanager/backup/backup.go
```diff
@@ -34,6 +35,7 @@ func Builder() *cobra.Command {
 		CheckpointsBuilder(),
 		EnableBuilder(),
 		config.Builder(),
+		blockstore.Builder(),
```
gssbzn, line 38: [nit] big thing here, according to spec this should be `mongocli ops-manager admin backup blockstore(s) lists`
notice the admin part on the command, also make sure that this command is only available for Ops Manager and not Cloud Manager

gssbzn (changes requested):
just one thing that this command namespace is not right, we should have a new admin scope and add it there

internal/cli/opsmanager/backup/backup.go
andreaangiolillo: ah okay, good call 👍

internal/cli/opsmanager/admin/backup/description.go
```diff
@@ -0,0 +1,18 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+package backup
+
+const (
+	backup = "Manage backup administration for your project."
```
gssbzn, line 17: here and admin description these are global settings and not per project so we can drop that part

gssbzn (changes requested):
one small nit to drop project from some descriptions as these are global settings

gssbzn (approved):
LGTM

PR #475, opened 2020-10-05
CLOUDP-73721: mongocli ops-manager admin backup blockstore(s) update

andreaangiolillo:
## Proposed changes

_Jira ticket:_ CLOUDP-73721

## Checklist

## Further comments

```
./bin/mongocli om admin backup blockstore update test --assignment  --uri mongodb://localhost:29022 -o json -P ops
{
  "id": "test",
  "uri": "mongodb://localhost:29022",
  "ssl": false,
  "assignmentEnabled": true,
  "encryptedCredentials": false,
  "usedSize": 204800,
  "loadFactor": 1
}
```

```
./bin/mongocli om admin backup blockstore update test --assignment  --uri mongodb://localhost:29022  -P ops
Blockstore configuration 'test' updated.
```


internal/cli/admin_opts.go
```diff
@@ -0,0 +1,62 @@
+// Copyright 2020 MongoDB Inc
```
gssbzn, line 1: [nit] can we move this to the `/opsmanager/admin/backup` package? this way the name is less confusing, `Admin` is to broad of a term to be in the global `cli` package

gssbzn (changes requested):
one small nit to move the AdminOpts closer to where we need it

internal/cli/admin_opts.go
andreaangiolillo: In this way we are going to hit `import cycle not allowed`. Should I rename it as `create_opts`?

gssbzn: feel free then to create a `backupstore` package in the admin one

gssbzn (approved):
LGTM

PR #501, opened 2020-10-16
CLOUDP-73748 mongocli ops-manager serverUsage download, CLOUDP-73749: mongocli ops-manager serverUsage capture

andreaangiolillo:
## Proposed changes

_Jira ticket:_ CLOUDP-73748
_Jira ticket:_ CLOUDP-73749

## Checklist

## Further comments

```
./bin/mongocli om serverUsage capture -P ops
Snapshot captured
```
```
./bin/mongocli om serverUsage download --startDate "2020-01-01" --endDate "2020-10-16" -P ops
Download of serverUsageReport.tar.gz completed.
```
Screenshot 2020-10-16 at 17 46 05


internal/cli/opsmanager/serverusage/download.go
```diff
@@ -0,0 +1,101 @@
...
+func (opts *DownloadOpts) newServerTypeOptions() *opsmngr.ServerTypeOptions {
+	return &opsmngr.ServerTypeOptions{
+		StartDate: opts.startDate,
+		EndDate:   opts.endDate,
+	}
+}
+
+func (opts *DownloadOpts) initDefaultOut() {
+	if opts.Out == "" {
+		opts.Out = "serverUsageReport" + opts.format
```
gssbzn, line 71: how does this format works? the docs say that it always returns a tgz, https://docs.opsmanager.mongodb.com/current/reference/api/diagnostics/get-project-diagnostic-archive/

andreaangiolillo: I think that I may have used the wrong end point. Here I have used https://docs.opsmanager.mongodb.com/current/reference/api/usage/create-one-report/ which is the one that I have implemented. Can you confirm that I have used the wrong endpoint?

gssbzn: no, you used the right one but that create us confusing, one thing here, for that endpoint the format is a query param, which you're not passing as a query param (newServerTypeOptions)

gssbzn: I have created a [docs](https://jira.mongodb.org/browse/DOCSP-12840) ticket to update that documentation

gssbzn (changes requested):
newServerTypeOptions doesn't look to be setting the format query param

internal/usage/usage.go
```diff
@@ -102,6 +102,7 @@ const (
 	EndDate                         = "Timestamp in ISO 8601 date and time format in UTC when the maintenance window ends."
 	ServerUsageStartDate            = "Timestamp in ISO 8601 date format when the list of host assignments starts."
 	ServerUsageEndDate              = "Timestamp in ISO 8601 date format when the list of host assignments ends."
+	ServerUsageFormat               = "Compression format of the resulting report. Accepted values for this option are: .zip, .tar.gz"
```
gssbzn, line 105: [super nit] could we avoid having to use the `.` it doesn't really add anything here as a param, you may need to handle this better in the file name is empty but it will reduce typing for users when setting it

internal/cli/opsmanager/serverusage/download.go
```diff
@@ -0,0 +1,102 @@
...
+	return &opsmngr.ServerTypeOptions{
+		StartDate:  opts.startDate,
+		EndDate:    opts.endDate,
+		FileFormat: opts.format,
+	}
+}
+
+func (opts *DownloadOpts) initDefaultOut() {
+	if opts.Out == "" {
+		opts.Out = "serverUsageReport" + opts.format
```
gssbzn, line 72: having resolved the format, could we improve the naming? the UI returns a name with the date filters in it, "serverUsage-{start-date}-{end-date}.{format}" can we do something similar here please? it makes it easier for users avoid to have to set names if creating different files for different date ranges, also with my other comment, we should not need to pass the dot(.) on the format, that's a naming problem you can solve here

gssbzn (changes requested):
two small nits on generating the file name

internal/cli/opsmanager/serverusage/download.go
andreaangiolillo: sure 👍

internal/cli/opsmanager/serverusage/download.go
```diff
@@ -0,0 +1,103 @@
...
+		StartDate:  opts.startDate,
+		EndDate:    opts.endDate,
+		FileFormat: opts.format,
+	}
+}
+
+func (opts *DownloadOpts) initDefaultOut() {
+	opts.format = "." + opts.format
+	if opts.Out == "" {
+		opts.Out = "serverUsage-" + opts.startDate + "-" + opts.endDate + opts.format
```
gssbzn, line 73: I think you're missing a dot(.) here, opts.endDate + "." + opts.format
Also optional, using `fmt.Sprintf` could make this a bit more readable

gssbzn (changes requested):

internal/cli/opsmanager/serverusage/download.go
andreaangiolillo: I don't because I am inserting `.` at line 71

internal/cli/opsmanager/serverusage/download.go
```diff
@@ -0,0 +1,103 @@
...
+func (opts *DownloadOpts) newServerTypeOptions() *opsmngr.ServerTypeOptions {
+	return &opsmngr.ServerTypeOptions{
+		StartDate:  opts.startDate,
+		EndDate:    opts.endDate,
+		FileFormat: opts.format,
+	}
+}
+
+func (opts *DownloadOpts) initDefaultOut() {
+	opts.format = "." + opts.format
```
gssbzn, line 71: for zip to work is has to be exactly `zip` with no dots, anything else the API assumes tar.gz

gssbzn (changes requested):

internal/cli/opsmanager/serverusage/download.go
andreaangiolillo: I may have misunderstood your comment here https://github.com/mongodb/mongocli/pull/501#discussion_r507611698.
The behaviour now is that the user has to specify the format flag as`--format zip` and `--format tar.gz`. 
Before it was `--format .zip` and `--format .tar.gz`

andreaangiolillo: mongocli use `tar.gz` if the flag is not specified

andreaangiolillo: when `opts.format = zip`, `opts.Out = "serverUsage-startDate-endDate.zip"`. Am I missing something?

gssbzn: yes, the behaviour is correct, but you're adding a dot to the query param, so the server will get `.zip` which will not match the check the server does `"zip".equals(param)` and then defaulting to tar.gz

andreaangiolillo: got it, Thanks for the explanation

gssbzn: this is not about the file name, but what you're sending to the server, adding it here solves the file name but breaks the API request

andreaangiolillo: I wrote this message https://github.com/mongodb/mongocli/pull/501#discussion_r507661018 before reading your comment (I didn’t refresh the page)

internal/cli/opsmanager/serverusage/download.go
```diff
@@ -0,0 +1,102 @@
...
+		},
+	}
+
+	cmd.Flags().StringVar(&opts.startDate, flag.StartDate, "", usage.ServerUsageStartDate)
+	cmd.Flags().StringVar(&opts.endDate, flag.EndDate, "", usage.ServerUsageEndDate)
+	cmd.Flags().StringVar(&opts.format, flag.Format, "tar.gz", usage.ServerUsageFormat)
+	cmd.Flags().StringVar(&opts.Out, flag.Out, "", usage.LogOut)
+	cmd.Flags().BoolVar(&opts.Force, flag.Force, false, usage.ForceFile)
+
+	_ = cmd.MarkFlagFilename(flag.Out)
```
gssbzn, line 97: is this required? you do a check to see if it's empty and set a default name

internal/cli/opsmanager/serverusage/download.go
andreaangiolillo: I am not sure why I thought that I need to pass ".zip" and ".tar.gz" to the api 🤷‍♂️

internal/cli/opsmanager/serverusage/download.go
andreaangiolillo: good call

gssbzn (approved):
LGTM, sorry about all the little changes

andreaangiolillo:
Not at all! Thanks for your review!

PR #504, opened 2020-10-19
CLOUDP-75271: Add TLS options to ProcessConfig

andreaangiolillo:
## Proposed changes

_Jira ticket:_ CLOUDP-75271

## Checklist


internal/convert/process_config.go
```diff
@@ -29,28 +29,29 @@ const (
...
+	Port                int          `yaml:"port" json:"port"`
+	Priority            *float64     `yaml:"priority,omitempty" json:"priority,omitempty"`
+	ProcessType         string       `yaml:"processType" json:"processType"`
+	SlaveDelay          *float64     `yaml:"slaveDelay,omitempty" json:"slaveDelay,omitempty"`
+	Version             string       `yaml:"version,omitempty" json:"version,omitempty"`
+	Votes               *float64     `yaml:"votes,omitempty" json:"votes,omitempty"`
+	ArbiterOnly         *bool        `yaml:"arbiterOnly,omitempty" json:"arbiterOnly,omitempty"`
+	Disabled            bool         `yaml:"disabled" json:"disabled"`
+	Hidden              *bool        `yaml:"hidden,omitempty" json:"hidden,omitempty"`
+	TLS                 *opsmngr.SSL `yaml:"tls,omitempty" json:"tls,omitempty"`
```
gssbzn, line 54: you can't directly use `opsmngr.SSL` it doesn't have yaml tgas so a yaml configuration won't be serialised here

gssbzn: this is also not as easy as just adding this field, you need to do a mapping between, opsmngr <-> convert, when reading from ops manager set the new fields in convert, when reading from convert set the right fields in ops manager

You may only need to update https://github.com/mongodb/mongocli/blob/023747d66b626d06dd0eaf72fab06622bab24d7b/internal/convert/process_config.go#L201-L207 to set ops manager TLS options since TLS is part of NET
When reading from ops manager you may need to update https://github.com/mongodb/mongocli/blob/023747d66b626d06dd0eaf72fab06622bab24d7b/internal/convert/process_config.go#L98

gssbzn (changes requested):
this is missing a lot of pieces to make it work

gssbzn (approved):
LGTM, thanks for all the work here

PR #505, opened 2020-10-19
CLOUDP-75269: Simplify PreRunE and PreRunEOrg

andreaangiolillo:
## Proposed changes

_Jira ticket:_ CLOUDP-75269

## Checklist


andreaangiolillo:
@gssbzn ❓ ❓ ❓ I would like to know what you think about this solution ❓ ❓ ❓

internal/cli/global_opts.go
```diff
@@ -51,14 +52,20 @@ func (opts *GlobalOpts) ConfigOrgID() string {
...
+	if opts.PreRunEOrg {
+		err := validateOrgID(opts)
+		if err != nil {
+			return err
+		}
+	} else {
+		err := validateProjectID(opts)
+		if err != nil {
+			return err
+		}
```
gssbzn, line 66: I'd rather stay away from an `if` here, this avoids the question, why `opts.PreRunEOrg` and not `opts.PreRunEProj` or why not both?
I think I mentioned this in the ticket, we should aim for a full [functional config](https://sagikazarmark.hu/blog/functional-options-on-steroids/) here and pushing the need to validate orgs or projects when calling `PreRunE`, as `opts.PreRunE(opts.validateProjectID, ...)`

andreaangiolillo: yes, I was looking for a way to change less files as possible. Thanks for the feedback!

gssbzn: I understand the fear and is one of the reasons we have not done this so far but the longer we wait the worst it'll get

internal/cli/global_opts_test.go
```diff
@@ -25,40 +25,37 @@ import (
 func TestGlobalOpts_PreRunE(t *testing.T) {
 	t.Run("empty project ID", func(t *testing.T) {
 		o := &GlobalOpts{}
-		if err := o.PreRunE(); err != errMissingProjectID {
+		if err := o.PreRunE(o.ValidateProjectID); err != errMissingProjectID {
```
gssbzn, line 28: you should probably change this test to be, `TestGlobalOpts_ValidateProjectID` and skip the `o.PreRunE` part given this is mostly testing the `ValidateProjectID` method

for `o.PreRunE` I'm not sure if we can have a proper unit tests here given what it does but feel free to try something

internal/cli/global_opts_test.go
```diff
@@ -25,40 +25,37 @@ import (
...
 			t.Fatalf("PreRunE() unexpected error %v\n", err)
 		}
 	})
-}
-
-func TestGlobalOpts_PreRunEOrg(t *testing.T) {
 	t.Run("empty org ID", func(t *testing.T) {
 		o := &GlobalOpts{}
-		if err := o.PreRunEOrg(); err != ErrMissingOrgID {
+		if err := o.PreRunE(o.ValidateOrgID); err != ErrMissingOrgID {
```
gssbzn, line 46: similar to above you should turn this into a test for `ValidateOrgID`

gssbzn (dismissed):
💯  massive effort here, just a couple of small nits with the tests for `ValidateProjectID` and `ValidateOrgID`

internal/cli/global_opts_test.go
gssbzn: like I said you don't need to call `o.PreRunE`
this can be just `if err := o.ValidateProjectID(); err != errMissingProjectID`, `PreRunE`, has no impact on this test so let's not mix them

gssbzn (changes requested):
one of my previous comments is still not addressed here

internal/cli/global_opts_test.go
andreaangiolillo: I am not sure if I agree here. In this way we are also testing `o.PreRunE`. in addition, `ValidateProjectID` and `ValidateOrgID` are used only with `o.PreRunE`. I understand that the generally idea is to not mix things together but I would argue that those two methods are bundled together.

gssbzn: ValidateProjectID and ValidateOrgID, is a method of its own, so it should have its own set of unit tests, the same goes for `PreRunE`, it should be its own set of unit tests that should not depend on `ValidateProjectID` or `ValidateOrgID`. This is basic unit testing, you don't want your test to have external dependencies because if the external dependency changes means unrelated tests now need to be updated

The fact that `PreRunE` always calls ValidateProjectID or ValidateOrgID doesn't mean this will always be the case, we have now open the possibility to use `PreRunE` in a couple of places where we were not doing it before because it will always validate

So ideally a test for `PreRunE` would inject its own set of dependencies.

internal/cli/global_opts_test.go
```diff
@@ -17,53 +17,78 @@
...
 	t.Run("valid org ID", func(t *testing.T) {
 		o := &GlobalOpts{OrgID: "5e98249d937cfc52efdc2a9f"}
-		if err := o.PreRunEOrg(); err != nil {
+		if err := o.ValidateOrgID(); err != nil {
 			t.Fatalf("PreRunE() unexpected error %v\n", err)
 		}
 	})
 }
 
+func TestGlobalOpts_PreRunE(t *testing.T) {
```
gssbzn, line 68: 💯  perfect

internal/cli/global_opts_test.go
```diff
@@ -17,53 +17,78 @@
...
+	noErrorFunc := func() error {
+		return nil
+	}
+
+	errorFunc := func() error {
+		return errors.New("error")
+	}
```
gssbzn, line 75: [non-blocking] you could define each inside the test that uses it but this is not a blocking change

gssbzn (approved):
LGTM, thanks for doing this, this was a tech debt item I've long wanted to address

gssbzn (approved):
LGTM :shipit:

PR #510, opened 2020-10-22
CLOUDP-75190: [mongocli] Add a command to fully remove clusters for OM/CM

andreaangiolillo:
## Proposed changes

_Jira ticket:_ CLOUDP-75190

## Checklist

## Further comments

```
./bin/mongocli om cluster unmanage myReplicaSet -P ops
? Are you sure you want to delete: myReplicaSet Yes
Cluster deleted
```

**Before:**

Screenshot 2020-10-22 at 20 17 01

**After:**

Screenshot 2020-10-22 at 20 19 08


internal/cli/opsmanager/clusters/delete.go
```diff
@@ -27,10 +29,13 @@ import (
 	"go.mongodb.org/ops-manager/atmcfg"
 )
 
+const defaultWait = 4 * time.Second
```
gssbzn, line 32: [nit] feel free to use [watch_opts](https://github.com/mongodb/mongocli/blob/master/internal/cli/watch_opts.go) to handle this and  the polling mechanism, it would be nice to add the spinner as this command is going to lock the terminal (a bit of an odd one as we usually don't do that)

internal/cli/opsmanager/clusters/delete.go
```diff
@@ -43,39 +48,162 @@ func (opts *DeleteOpts) Run() error {
 	if !opts.Confirm {
 		return nil
 	}
+
+	err := opts.newHostIds()
```
gssbzn, line 52: [nit] usually i'd expect a "new" method to return something, this seems more like an `init` kind of.
`init` methods are not great as this means keeping a context, in this case you're keeping it in `opts` where we usually reserve `opts` for the external parameters that users set.
I'd expect this to be more like  `opts.HostIDs` where it returns a slice of ids to pass to another function

internal/cli/opsmanager/clusters/delete.go
```diff
@@ -43,39 +48,162 @@ func (opts *DeleteOpts) Run() error {
...
+	r := convert.FromAutomationConfig(current)
+	hostname := make(map[string][]int32)
+
+	for _, rs := range r {
+		if rs.Name == opts.Entry {
+			for _, s := range rs.ProcessConfigs {
+				hostname[s.Hostname] = append(hostname[s.Hostname], int32(s.Port))
+			}
+		}
+	}
```
gssbzn, line 93: [nit] this is going to be quite slow, convert to the cli presentation just to group per replica set, to then find the replicaset you want, processes have [cluster field](https://github.com/mongodb/go-client-mongodb-ops-manager/blob/bfe7ba34b319e2a9fb17a0d60f59f5b3c5cbc07d/opsmngr/automation_config.go#L330) which should match the name of the cluster you're deleting, you can then just search over all the processes by this field

internal/cli/opsmanager/clusters/delete.go
```diff
@@ -43,39 +48,162 @@ func (opts *DeleteOpts) Run() error {
...
 
+	hosts, err := opts.store.Hosts(opts.ConfigProjectID(), nil)
```
gssbzn, line 95: we may get pagination problems here, ie, what if the host you're looking for is in the second page?, two options here, try to pass the max value allowed for items per page, or use 
[get-one-host-by-hostname-port](https://docs.opsmanager.mongodb.com/current/reference/api/hosts/get-one-host-by-hostname-port/)
Given you recently collected all processes and should have the hostname and port info, this may do more network calls but it may be the safest option in case of pagination edge cases

internal/cli/opsmanager/clusters/delete.go
```diff
@@ -43,39 +48,162 @@ func (opts *DeleteOpts) Run() error {
...
 		DeleteOpts: cli.NewDeleteOpts("", "Cluster not deleted\""),
 	}
 	cmd := &cobra.Command{
-		Use:     "delete <name>",
-		Aliases: []string{"rm"},
-		Short:   "This is an internal and undocumented command.",
-		Long:    "This commands only removes entries from the automation config but does not actually remove a cluster.",
+		Use:     "unmanage <name>",
+		Aliases: []string{"rm", "delete"},
+		Short:   DeleteCluster,
```
gssbzn, line 205: sorry if this was not clear from the ticket, we want to keep the old delete. the old delete should be renamed to `unmanage`, users may still want to just unmanaged a deployment
Then we want a new `delete` command, which does what you just implemented

internal/store/automation.go
```diff
@@ -56,6 +56,14 @@ type CloudManagerClustersDescriber interface {
 	OpsManagerClusterDescriber
 }
 
+type CloudManagerClustersDeleter interface {
+	AutomationGetter
+	AutomationUpdater
```
gssbzn, line 61: [nit]
```suggestion
	AutomationPatcher
```

gssbzn (changes requested):
This is a great start, left a couple of nits and suggestions, main thing here is that we want to keep the old delete, that should be renamed to unmanage, and then we can add a new delete command which does what you just implemented

internal/cli/opsmanager/clusters/clusters.go
```diff
@@ -37,6 +37,8 @@ func Builder() *cobra.Command {
 		DeleteBuilder(),
 		ApplyBuilder(),
 		IndexesBuilder(),
+		DeleteBuilder(),
```
gssbzn, line 40: [nit] duplicated?

internal/cli/opsmanager/clusters/delete.go
```diff
@@ -43,44 +44,147 @@ func (opts *DeleteOpts) Run() error {
...
+	err = opts.stopMonitoring(hostIds)
+	if err != nil {
+		return err
+	}
+
+	fmt.Print("Cluster deleted\n")
+	return nil
+}
+
+func (opts *DeleteOpts) newHostIds() ([]string, error) {
```
gssbzn, line 75: [nit] following go no getter and caps on ids this should be just `hostIDs`

internal/cli/opsmanager/clusters/delete.go
```diff
@@ -43,44 +44,147 @@ func (opts *DeleteOpts) Run() error {
...
 	current, err := opts.store.GetAutomationConfig(opts.ConfigProjectID())
+	if err != nil {
+		return nil, err
+	}
+
+	hostnameMap := make(map[string][]int32)
+	for _, process := range current.Processes {
+		if process.Args26.Replication.ReplSetName == opts.Entry {
+			hostnameMap[process.Hostname] = append(hostnameMap[process.Hostname], int32(process.Args26.NET.Port))
+		}
```
gssbzn, line 85: [q] does this work for a sharded cluster as well?

internal/cli/opsmanager/clusters/delete.go
```diff
@@ -43,44 +44,147 @@ func (opts *DeleteOpts) Run() error {
...
+}
+
 // mongocli cloud-manager cluster(s) delete <name> --projectId projectId [--force]
 func DeleteBuilder() *cobra.Command {
 	opts := &DeleteOpts{
 		DeleteOpts: cli.NewDeleteOpts("", "Cluster not deleted\""),
 	}
 	cmd := &cobra.Command{
-		Use:     "delete <name>",
+		Use:     "delete [name]",
```
gssbzn, line 178: [nit] the name is not optional, brackets for posix mean optional

gssbzn (changes requested):
thanks for looking into this, I left some minor comments and questions but nothing major. I may try this command before merging given how much it does I want to make sure the workflow is right

internal/cli/opsmanager/clusters/delete.go
andreaangiolillo: I am not sure, I have to test this scenario

andreaangiolillo: No it doesn’t. I see that we have `process.cluster` for `mongos` but we don't for `mongod`. `mongods` have 

```
replication": {"replSetName": "myShard_0"},
```

We may ask to the user to provide the shard name and use `contains("shardName")` here

gssbzn: `process` has a `Cluster` prop, it's what I used when grouping processes of a sharded cluster but I don't remember if that's set for a replica set as well

andreaangiolillo: I am wondering what is the workflow with a shared cluster: should I shutdown the mongod first? configRS? or mongos? I am going to have a look to the API call that is used in the atlas ui.

gssbzn: I'm testing this and it looks like the shutdown has a bug that us not sutting down the config RS nor the mongos, see https://github.com/mongodb/go-client-mongodb-ops-manager/blob/3bd88d15acfe2afed2bd8293c6457acfd99e3c2d/atmcfg/atmcfg.go#L29-L36

So we may need to fix that first

Then we have the issue of getting host ids, and maybe we should do something like the shutdown method and just walk thru all the arrays collecting what matches

andreaangiolillo:
I tested my changes with a sharded cluster and it works

e2e/cloud_manager/deploy_sharded_cluster_test.go
```diff
@@ -120,3 +120,58 @@ func TestDeployCluster(t *testing.T) {
 		}
 	})
 }
+
+func TestDeployDeleteCluster(t *testing.T) {
```
gssbzn, line 124: 💯  thanks for adding this as well

internal/cli/opsmanager/clusters/delete.go
```diff
@@ -43,27 +45,149 @@ func (opts *DeleteOpts) Run() error {
...
+	}
 
+	if len(hostIds) == 0 {
+		return nil, fmt.Errorf("cluster '%s' doesn't exist", opts.Entry)
+	}
+
+	return hostIds, nil
+}
+
+func (opts *DeleteOpts) replicaSetHostNames(automationConfig *opsmngr.AutomationConfig, hostnameMap map[string][]int32, name string) {
```
gssbzn, line 104: [nit] given you don't need `opts`
```suggestion
func replicaSetHostNames(automationConfig *opsmngr.AutomationConfig, hostnameMap map[string][]int32, name string) {
```

internal/cli/opsmanager/clusters/delete.go
```diff
@@ -43,27 +45,149 @@ func (opts *DeleteOpts) Run() error {
...
+			opts.replicaSetHostNames(automationConfig, hostnameMap, sharding.ConfigServerReplica)
+			break
+		}
+	}
+}
+
+func (opts *DeleteOpts) stopMonitoring(hostIds []string) error {
+	for _, id := range hostIds {
+		err := opts.store.StopMonitoring(opts.ConfigProjectID(), id)
+		if err != nil {
```
gssbzn, line 128: nit 
```suggestion		
		if err := opts.store.StopMonitoring(opts.ConfigProjectID(), id); err != nil {
```

internal/cli/opsmanager/clusters/description.go
```diff
@@ -24,4 +24,6 @@ const (
 	StartUpCluster  = "Start up a cluster for your project."
 	Indexes         = "Manage cluster rolling indexes for your project."
 	CreateIndex     = "Create a rolling index for your MongoDB cluster."
+	DeleteCluster   = "Delete a cluster for your project."
+	UnmanageCluster = "Remove a cluster entry from the automation config."
```
gssbzn, line 28: super nit
```suggestion
	DeleteCluster   = "Completely removes a cluster from your project."
	UnmanageCluster = "Stop managing a cluster via automation."
```

gssbzn (changes requested):
💯  this is a great add, couple of minor nits but the main feature is done

gssbzn (approved):
LGTM :shipit:

PR #512, opened 2020-10-28
CLOUDP-77992: mongocli atlas quickStart

andreaangiolillo:
## Proposed changes
`mongocli atlas quickstart` allows the user to quickly deploy a cluster and connect into it.
This command will have the following functionalities:

- no required flags: the user has optional flags to personalise the cluster
- adds client's public IP to access list
- creates an atlas dbuser
- deploys a cluster: if no optional flags are used, the command deploys a cluster on AWS with the following properties:

  - ClusterType:` ReplicaSet
  - MongoDBMajorVersion: 4.4
  - DiskSizeGB: 2GB
  - Tier: M2
  - Members: 3

- Waits for the cluster to be deployed
- Prints the following message once it has finished: 
```
Now you can connect to your Atlas cluster with:
mongo -u quickstartUser -p Password1! mongodb+srv://getstarted.ajlj3.mongodb-dev.net
```

## Checklist

## Further comments

I tested it with Mac/Linux/Windows

ezgif com-gif-maker


internal/cli/atlas/quickstart/quick_start.go
```diff
@@ -0,0 +1,248 @@
...
+		AWSIAMType:   none,
+		LDAPAuthType: none,
+		DatabaseName: convert.AdminDB,
+		Username:     opts.dbUsername,
+	}
+}
+
+// newIPAddress returns client's public ip
+func (opts *Opts) newIPAddress() (string, error) {
+	command := "curl ifconfig.me"
```
gssbzn, line 146: [nit] this won't work on windows, ie `curl` may not be installed, you may want to implement a little go client to get this info

internal/cli/atlas/quickstart/quick_start.go
```diff
@@ -0,0 +1,248 @@
...
+			return opts.Run()
+		},
+	}
+
+	cmd.Flags().StringVar(&opts.clusterName, flag.ClusterName, "GetStarted", usage.ClusterName)
+	cmd.Flags().StringVar(&opts.provider, flag.Provider, "AWS", usage.Provider)
+	cmd.Flags().StringVarP(&opts.region, flag.Region, flag.RegionShort, "US_EAST_1", usage.Region)
+	cmd.Flags().StringVar(&opts.ipAddress, flag.IP, "", usage.AccessListIPEntry)
+	cmd.Flags().StringVar(&opts.dbUsername, flag.Username, "quickstartUser", usage.DBUsername)
+	cmd.Flags().StringVar(&opts.dbUserPassword, flag.Password, "Password1!", usage.Password)
```
gssbzn, line 243: [nit] please don't provide defaults for this, ask them both as part of the flow if empty

internal/cli/atlas/quickstart/quick_start.go
```diff
@@ -0,0 +1,248 @@
...
+		},
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return opts.Run()
+		},
+	}
+
+	cmd.Flags().StringVar(&opts.clusterName, flag.ClusterName, "GetStarted", usage.ClusterName)
+	cmd.Flags().StringVar(&opts.provider, flag.Provider, "AWS", usage.Provider)
+	cmd.Flags().StringVarP(&opts.region, flag.Region, flag.RegionShort, "US_EAST_1", usage.Region)
+	cmd.Flags().StringVar(&opts.ipAddress, flag.IP, "", usage.AccessListIPEntry)
```
gssbzn, line 241: [nit] it would be great to give the option to input this in the wizard if empty or use current IP

internal/cli/atlas/quickstart/quick_start.go
```diff
@@ -0,0 +1,248 @@
...
+			)
+		},
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return opts.Run()
+		},
+	}
+
+	cmd.Flags().StringVar(&opts.clusterName, flag.ClusterName, "GetStarted", usage.ClusterName)
+	cmd.Flags().StringVar(&opts.provider, flag.Provider, "AWS", usage.Provider)
+	cmd.Flags().StringVarP(&opts.region, flag.Region, flag.RegionShort, "US_EAST_1", usage.Region)
```
gssbzn, line 240: similar to other comments, let's not have defined one, we should ask users if none is given
Also please consider using a defined list for selecting the region/provider combo, so users should not have to type the region but select from a list

gssbzn (changes requested):
This is great, I have a couple of recommendations to improve this a little
Main one, `curl` will not work on windows by default, you may want to implement this in go
The other in general terms let's avoid setting defaults for the flags and ask users their input if no value is given

mmarcon:
Any chance we can check whether `mongosh` is installed and if it is, suggest to use that to connect?

gssbzn:
@andreaangiolillo any update on this? it would be really nice to finish this and get it ready for review

andreaangiolillo:
Hope to have some time in the next 2 weeks or during my help rota week to complete this and the cluster delete command.

internal/cli/atlas/quickstart/quick_start.go
andreaangiolillo: I left out the cluster name flag because I would like to show as less wizards as possible

internal/cli/atlas/quickstart/quick_start.go
```diff
@@ -0,0 +1,407 @@
...
+		publicIP, err := newIPAddress()
+		message := "Insert the IP entry to add to the Access List"
+		if err == nil {
+			message = fmt.Sprintf(`Insert the IP entry to add to the Access List [Press Enter to use your public IP "%s"]`, publicIP)
+		}
+
+		publicIPPrompt := survey.Input{
+			Message: message,
+			Help:    usage.AccessListIPEntry,
+		}
```
gssbzn, line 302: you can set a default for a prompt and it will show like this
```bash
mongocli config --service cloud-manager -P broken
You are configuring a profile for mongocli.

All values are optional and you can use environment variables (MCLI_*) instead.

Enter [?] on any option to get help.

? Public API Key: [? for help] (JQVAXGIA)
```

that value between parenthesis is the value already set

gssbzn (changes requested):
Great improvements now with everything optional and with prompts, now let's make those prompts really work for use

jamesbroadhead:
I know that this is counter to Gustavo's suggestion - but I really liked the idea of this taking no required flags. The use case I was imagining was something like a sticker or card we could hand out at .Live which says 
`brew install mongocli ; mongocli atlas quickstart` 
and that that would result in a running database. I think it's a really compelling use-case. 

The only hard blocker is API keys - I'd prefer to take `whoami` as the username, generate a password and select an IP allowlist by figuring out what Atlas sees from my current network. 

== 
Second thought is about the educational value of this quickstart command. 
It'd be really cool if as it took each action, it printed the command that you "would have had to run" to do the same -- something like 

``` 
$ mongocli atlas quickstart

Running: mongocli atlas cluster create ... 
Running: mongocli atlas cluster whitelist ... 
... 
```

internal/cli/atlas/quickstart/quick_start.go
```diff
@@ -0,0 +1,407 @@
...
+		LDAPAuthType: none,
+		DatabaseName: convert.AdminDB,
+		Username:     opts.dbUsername,
+	}
+}
+
+// newIPAddress returns client's public ip
+func newIPAddress() (string, error) {
+	publicIP := ""
+	for _, uri := range APIURIs {
```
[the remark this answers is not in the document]
andreaangiolillo: No, I didn’t. The only reason why I am iterating over this list is to make the function more reliable so that, if the first link of the list returns an error, this function would still work.

internal/cli/atlas/quickstart/quick_start.go
```diff
@@ -0,0 +1,407 @@
...
+					publicIP = string(responseBytes)
+					break
+				}
+			}
+		}
+	}
+
+	if publicIP == "" {
+		return publicIP, errors.New("error in finding your public IP, please use --ip to provide your public ip")
+	}
```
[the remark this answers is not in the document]
andreaangiolillo: This logic is already implemented (**lines 291-302** of **internal/cli/atlas/quickstart/quick_start.go**): if `newIPAddress()` returns an error, we would show the message `Insert the IP entry to add to the Access List` and let the user types the ip address.

```java
if opts.ipAddress == "" {
		answer := ""
		publicIP, err := newIPAddress()
		message := "Insert the IP entry to add to the Access List"
		if err == nil {
			message = fmt.Sprintf(`Insert the IP entry to add to the Access List [Press Enter to use your public IP "%s"]`, publicIP)
		}

		publicIPPrompt := survey.Input{
			Message: message,
			Help:    usage.AccessListIPEntry,
		}
```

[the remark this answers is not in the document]
andreaangiolillo: with that said, I do agree that the error message should be more generic given that we want to reuse this function.

internal/cli/atlas/quickstart/quick_start.go
```diff
@@ -0,0 +1,407 @@
...
+func (opts *Opts) newWhitelist() *atlas.ProjectIPWhitelist {
+	return &atlas.ProjectIPWhitelist{
+		GroupID:   opts.ConfigProjectID(),
+		Comment:   accessListComment,
+		IPAddress: opts.ipAddress,
+	}
+}
+
+func (opts *Opts) newCluster() *atlas.Cluster {
+	diskSizeGB := float64(diskSizeGB)
```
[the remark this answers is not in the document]
andreaangiolillo: are you suggesting to show a wizard for the diskSize like we do for the region?

internal/cli/atlas/quickstart/quick_start.go
```diff
@@ -0,0 +1,407 @@
...
+		}
+		opts.dbUsername = dbUsername
+
+		if opts.dbUsername == "" {
+			n, err := e2e.RandInt(maxRandNum)
+			if err != nil {
+				return err
+			}
+			opts.dbUsername = "quickStart_" + n.String()
+		}
```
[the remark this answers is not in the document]
andreaangiolillo: leftover, sorry 🤦  What do you think about using `whoami ` here for the username as James suggested?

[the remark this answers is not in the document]
andreaangiolillo: I honestly like the idea and I think that `whoami` would work with window as well 😃

internal/cli/atlas/quickstart/quick_start.go
```diff
@@ -0,0 +1,353 @@
...
+			{
+				Name: "clusterName",
+				Prompt: &survey.Input{
+					Message: "Insert the cluster name",
+					Help:    usage.Provider,
+					Default: clusterName,
+				},
+			},
+		}
+		qs = append(qs, q...)
```
gssbzn, line 285: I have not fully reviewed this but FYI you do `q...` because is a slice, you can simplify this to be 
```suggestion
		q := &survey.Question{
			Name: "clusterName",
			Prompt: &survey.Input{
				Message: "Insert the cluster name",
				Help:    usage.Provider,
				Default: clusterName,
			},
		}
		qs = append(qs, q)
```

internal/cli/atlas/quickstart/quick_start.go
```diff
@@ -0,0 +1,429 @@
...
+	if publicIP != "" {
+		message = fmt.Sprintf("Insert the IP entry to add to the Access List [Press Enter to use your public IP address '%s']", publicIP)
+	}
+
+	return &survey.Question{
+		Name: "ipAddress",
+		Prompt: &survey.Input{
+			Message: message,
+			Help:    usage.AccessListIPEntry,
+			Default: publicIP,
```
andreaangiolillo, line 300: I am not sure why lint is failing given that `publicIP` is used also outside the if

andreaangiolillo: I don't get this error when running `make lint`

gssbzn: your version of the linter is older probably also this may be a false possitive

andreaangiolillo: I have added` //nolint`

internal/net/net.go
```diff
@@ -0,0 +1,80 @@
...
+		http.MethodGet,
+		uri,
+		nil,
+	)
+
+	if err != nil {
+		return ""
+	}
+
+	req.Header.Add("Accept", "application/json")
```
gssbzn, line 68: you set this but it looks like you are getting is actually a string (`text/plain`) and not a json object, maybe some endpoints do return a json in which case this method may break so more to my point of sticking to one service

internal/net/net.go
```diff
@@ -0,0 +1,80 @@
...
+	"http://ipecho.net/plain",
+	"http://icanhazip.com",
+	"http://ifconfig.me/ip",
+	"http://ident.me",
+	"http://checkip.amazonaws.com",
+	"http://bot.whatismyipaddress.com",
+	"http://whatismyip.akamai.com",
+	"http://wgetip.com",
+	"http://ip.tyk.nu",
+}
```
gssbzn, line 36: [nit] let's simplify this to just one, maybe `https://checkip.amazonaws.com/`, also I'm reading that service may return multiple IPs if the users is behind a proxy, also please keep in mind that I like to move this to an in house endpoint

internal/net/net.go
```diff
@@ -0,0 +1,80 @@
...
+	if err != nil {
+		return ""
+	}
+
+	req.Header.Add("Accept", "application/json")
+	res, err := http.DefaultClient.Do(req)
+
+	if err == nil {
+		responseBytes, err1 := ioutil.ReadAll(res.Body)
+		err2 := res.Body.Close()
```
gssbzn, line 73: you usually want to defer the close, please check how we handle this kind of things on the atlas or ops manager client

internal/cli/atlas/quickstart/quick_start.go
```diff
@@ -0,0 +1,431 @@
...
+		return nil
+	}
+	return &survey.Question{
+		Name: "region",
+		Prompt: &survey.Select{
+			Message: "Insert the physical location of your MongoDB cluster",
+			Help:    usage.Region,
+			Options: DefaultRegions[strings.ToUpper(opts.Provider)],
+		},
+	}
```
gssbzn, line 353: [nit] let's follow. asimilar pattern to the config command where we moved any `survey.Prompt` or `survey.Question
` to a `prompt.go` file, it'll make this easier to review

internal/cli/atlas/quickstart/quick_start.go
```diff
@@ -0,0 +1,431 @@
...
+)
+
+const quickstartTemplate = "Now you can connect to your Atlas cluster with: mongo -u %s -p %s %s\n"
+
+const (
+	replicaSet        = "REPLICASET"
+	diskSizeGB        = 20.0
+	mdbVersion        = "4.4"
+	shards            = 1
+	tier              = "M20"
```
gssbzn, line 45: shouldn't the default be `M2`?

internal/cli/atlas/quickstart/quick_start.go
```diff
@@ -0,0 +1,431 @@
...
+	"github.com/mongodb/mongocli/internal/usage"
+	"github.com/spf13/cobra"
+	atlas "go.mongodb.org/atlas/mongodbatlas"
+)
+
+const quickstartTemplate = "Now you can connect to your Atlas cluster with: mongo -u %s -p %s %s\n"
+
+const (
+	replicaSet        = "REPLICASET"
+	diskSizeGB        = 20.0
```
gssbzn, line 42: you can use, https://github.com/mongodb/go-client-mongodb-atlas/blob/943e1c852f6a1389f7281720566b71eaff148214/mongodbatlas/clusters.go#L174

gssbzn (commented):
I left a couple of initial comments, main thing is trying to simplify the file so this may become easier to review, please check what we did with config where the survey questions were separated to a different file

internal/cli/atlas/quickstart/quick_start.go
```diff
@@ -0,0 +1,431 @@
...
+			message = fmt.Sprintf("Insert the Username for authenticating to MongoDB [Press Enter to use '%s']", usrDefault)
+		}
+		q := &survey.Question{
+			Name: "dbUsername",
+			Prompt: &survey.Input{
+				Message: message,
+				Help:    usage.DBUsername,
+				Default: usrDefault,
+			},
+			Validate: func(val interface{}) error {
```
gssbzn, line 249: [nit] let's move this function to the top level

internal/cli/atlas/quickstart/quick_start.go
```diff
@@ -0,0 +1,431 @@
...
+var DefaultRegions = map[string][]string{
+	"AWS":   {"US_EAST_1", "US_WEST_2", "AP_SOUTH_1", "AP_EAST_2", "EU_WEST_1", "EU_CENTRAL_1", "ME_SOUTH_1", "AF_SOUTH_1"},
+	"GCP":   {"CENTRAL_US", "CANADA_CENTRAL", "WESTERN_EUROPE", "ASIA_SOUTH_EAST", "SOUTH_AFRICA_NORTH", "UAE_NORTH"},
+	"AZURE": {"US_EAST_2", "US_WEST", "EUROPE_NORTH"},
+}
+
+type Opts struct {
+	cli.GlobalOpts
+	cli.WatchOpts
+	cli.ListOpts
```
gssbzn, line 65: [nit] the user is not paginating so this is kind of a hacky way to do what I think you're doing

internal/cli/atlas/quickstart/quick_start.go
```diff
@@ -0,0 +1,431 @@
...
+	// dbUsername can only contain ASCII letters, numbers, hyphens and underscores
+	out := strings.TrimSpace(string(stdout))
+	var re = regexp.MustCompile("([^A-Za-z0-9_-])")
+	return re.ReplaceAllString(out, "_")
+}
+
+// newClusterName returns an auto-generate Cluster name
+func (opts *Opts) newClusterName() string {
+	cs, _ := opts.store.ProjectClusters(opts.ConfigProjectID(), opts.NewListOptions())
+	if clusters, ok := cs.([]atlas.Cluster); ok {
```
gssbzn, line 375: what happens if there are more than 100 clusters in the project?

internal/cli/atlas/quickstart/quick_start.go
```diff
@@ -0,0 +1,431 @@
...
+			Help:    usage.Region,
+			Options: DefaultRegions[strings.ToUpper(opts.Provider)],
+		},
+	}
+}
+
+// dbUsername returns the username of the user by running the command 'whoami'
+func dbUsername() string {
+	command := "whoami"
+	cmd := exec.Command("bash", "-c", command)
```
gssbzn, line 359: bash won't work on windows (and I just tested this)

internal/cli/atlas/quickstart/quick_start.go
```diff
@@ -0,0 +1,431 @@
...
+}
+
+func (opts *Opts) newClusterQuestions() []*survey.Question {
+	var qs []*survey.Question
+
+	if opts.ClusterName == "" {
+		clusterName := opts.newClusterName()
+		message := "Insert the cluster name"
+		if clusterName != "" {
+			message = fmt.Sprintf("Insert the cluster name [Press Enter to use the auto-generated name '%s']", clusterName)
```
gssbzn, line 314: having the name as both the message and the default makes it render like
>Insert the cluster name [Press Enter to use the auto-generated name 'QuickstartCluster610148'] QuickstartCluster610148

so more of a question for design on having it double

internal/cli/atlas/quickstart/quick_start.go
andreaangiolillo: mh... This is strange. I tested on windows (using msi) and it worked: It is the Administrator user that you see here 

Screenshot 2021-03-05 at 14 24 59

internal/cli/atlas/quickstart/quick_start.go
```diff
@@ -0,0 +1,431 @@
...
+
+	// dbUsername can only contain ASCII letters, numbers, hyphens and underscores
+	out := strings.TrimSpace(string(stdout))
+	var re = regexp.MustCompile("([^A-Za-z0-9_-])")
+	return re.ReplaceAllString(out, "_")
+}
+
+// newClusterName returns an auto-generate Cluster name
+func (opts *Opts) newClusterName() string {
+	cs, _ := opts.store.ProjectClusters(opts.ConfigProjectID(), opts.NewListOptions())
```
gssbzn, line 374: to add to my comment above about embedding list options, you can also do 
```suggestion
	cs, _ := opts.store.ProjectClusters(opts.ConfigProjectID(), nil)
```

andreaangiolillo: I will use the get instead of the list here so this won’t be needed

internal/cli/atlas/quickstart/quick_start.go
gssbzn: depends if you have something like `mingw`, which I don't, and we should cater for that kind of users as well, I went quite basic even using `cmd` over `powershell` and I have no dev tools installed

internal/net/net.go
jamesbroadhead: +1 to trusting amazon over whoever is running those others

internal/cli/atlas/quickstart/quick_start.go
jamesbroadhead: +1 to guss's point - you'll need to call a windows API to get the username. 
tbqh, the unix version of this should also be updated to call a unix API to get the username rather than running a shell :) 

this looks promising: https://golangcode.com/get-the-current-username-name-and-home-dir-cross-platform/ 

that said: you could merge this like this & do that change in a separate ticket

internal/store/clusters.go
```diff
@@ -62,6 +62,16 @@ type AtlasClusterGetterUpdater interface {
 	ClusterUpdater
 }
 
+type AtlasClusterQuickStarter interface {
+	ClusterStarter
```
gssbzn, line 66: [nit] I think you don't need `ClusterStarter`

internal/cli/atlas/quickstart/quick_start.go
```diff
@@ -0,0 +1,259 @@
...
+	}
+
+	// Get cluster's connection string
+	cluster, err := opts.store.AtlasCluster(opts.ConfigProjectID(), opts.ClusterName)
+	if err != nil {
+		return err
+	}
+	opts.connectionString = cluster.SrvAddress
+
+	fmt.Printf(quickstartTemplate, opts.DBUsername, opts.DBUserPassword, opts.connectionString)
```
gssbzn, line 109: [q] why `opts.connectionString`? any advantage over
```suggestion
	fmt.Printf(quickstartTemplate, opts.DBUsername, opts.DBUserPassword, cluster.SrvAddress)
```

gssbzn (commented):
almost there, main thing is to try to simplify the `prompt.go` file, it's better to separate presentation from business logic and right now they are still quite mixed up
Another thing is trying to use go implementation of the os/user package as we can avoid using bash(or powershell) if it works correctly

internal/cli/atlas/quickstart/prompt.go
```diff
@@ -0,0 +1,198 @@
...
+}
+
+// dbUsername returns the username of the user by running the command 'whoami'
+func dbUsername() string {
+	command := "whoami"
+	cmd := exec.Command("bash", "-c", command)
+
+	if runtime.GOOS == "windows" {
+		cmd = exec.Command("powershell", "", command)
+	}
```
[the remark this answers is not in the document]
andreaangiolillo: yes, I did but I thought to create a separate PR for that. Anyway, I will try to use it here

internal/cli/atlas/quickstart/quick_start.go
```diff
@@ -0,0 +1,356 @@
...
+	dbUser, err := opts.store.DatabaseUser(convert.AdminDB, opts.ConfigProjectID(), username)
+	if err != nil {
+		if !strings.Contains(err.Error(), fmt.Sprintf("No user with username %s exists.", username)) {
+			return err
+		}
+	}
+
+	if dbUser != nil {
+		return errors.New("a user with this username already exists")
+	}
```
gssbzn, line 289: [nit] you can use modern go error parsing and the atlas error struct to simplify this
```suggestion
	var target *atlas.ErrorResponse
	if err != nil && errors.As(err, &target) {
		if err.StatusCode == "USERNAME_NOT_FOUND" {
			return nil
		}
		return err
	}
	return fmt.Errorf("username %s already exists", username)
```

internal/cli/atlas/quickstart/quick_start.go
```diff
@@ -0,0 +1,356 @@
...
+
+	// dbUsername can only contain ASCII letters, numbers, hyphens and underscores
+	out := strings.TrimSpace(userStruct.Username)
+	var re = regexp.MustCompile("([^A-Za-z0-9_-])")
+	return re.ReplaceAllString(out, "_")
+}
+
+// newClusterName returns an auto-generate Cluster name
+func (opts *Opts) newClusterName() string {
+	cs, _ := opts.store.ProjectClusters(opts.ConfigProjectID(), nil)
```
gssbzn, line 309: [nit] you should check this error, what happens on a network error for example?

internal/cli/atlas/quickstart/quick_start.go
```diff
@@ -0,0 +1,356 @@
...
+	return ""
+}
+
+// mongocli atlas dbuser(s) quickstart [--clusterName clusterName] [--provider provider] [--region regionName] [--projectId projectId] [--username username] [--password password]
+func Builder() *cobra.Command {
+	opts := &Opts{}
+	cmd := &cobra.Command{
+		Use: "quickstart",
+		Example: `mongocli atlas quickstart
+mongocli atlas quickstart --clusterName Test --provider GPC --username dbuserTest --password Test!
```
gssbzn, line 330: [nit] `mongocli atlas quickstart` is already part of the help, is better if you explain what this does, ex:
```suggestion
		Example: `
  Skip setting cluster name, provider or database username by using the command options
  $ mongocli atlas quickstart --clusterName Test --provider GPC --username dbuserTest
```

internal/search/search.go
```diff
@@ -40,3 +41,14 @@ func ClusterExists(c *opsmngr.AutomationConfig, name string) bool {
 
 	return rsFound || shardedFound
 }
+
+// IsClusterFound returns true if a cluster exists for the given name
+func IsClusterFound(clusters []atlas.Cluster, name string) bool {
```
gssbzn, line 46: consider the package part of your name, `search.IsClusterFound` is probably not the best name, some alternatives
`search.AtlasClusterByName`, `search.AtlasClusterExists`

internal/cli/atlas/quickstart/quick_start.go
```diff
@@ -0,0 +1,356 @@
...
+func (opts *Opts) askClusterFlags() error {
+	var qs []*survey.Question
+
+	message := "Insert the cluster name"
+	clusterName := opts.ClusterName
+	if clusterName == "" {
+		clusterName = opts.newClusterName()
+		if clusterName != "" {
+			message = fmt.Sprintf("Insert the cluster name [Press Enter to use the auto-generated name '%s']", clusterName)
+		}
```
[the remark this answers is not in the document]
andreaangiolillo: yes, when  `opts.store.ProjectClusters(opts.ConfigProjectID(), nil)` returns an error, in this scenario we wouldn’t go inside the for loop
```
// newClusterName returns an auto-generate Cluster name
func (opts *Opts) newClusterName() string {
	cs, _ := opts.store.ProjectClusters(opts.ConfigProjectID(), nil)
	i := 0
	if clusters, ok := cs.([]atlas.Cluster); ok {
		for {
			clusterName := "QuickstartCluster" + strconv.Itoa(i)
			if !search.IsClusterFound(clusters, clusterName) {
				return clusterName
			}
			i++
		}
	}

	return ""
}
```

internal/cli/atlas/quickstart/quick_start.go
```diff
@@ -0,0 +1,356 @@
...
+			message = fmt.Sprintf("Insert the IP entry to add to the Access List [Press Enter to use your public IP address '%s']", publicIP)
+		}
+		q := newAccessListQuestion(publicIP, message)
+		qs = append(qs, q)
+	}
+
+	if len(qs) > 0 {
+		if err := survey.Ask(qs, opts); err != nil {
+			return err
+		}
```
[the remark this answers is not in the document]
andreaangiolillo: I think that this piece of code behaves in a different way: I want to leave the function only if I got an error

internal/cli/atlas/quickstart/quick_start.go
```diff
@@ -0,0 +1,356 @@
...
+		opts.DBUserPassword = pwd
+	}
+
+	return nil
+}
+
+func (opts *Opts) validateUniqueUsername(val interface{}) error {
+	username, _ := val.(string)
```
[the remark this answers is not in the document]
andreaangiolillo: See: https://github.com/mongodb/mongocli/pull/512#discussion_r590393087. We wouldn’t go inside the for loop and the function would return `""`

internal/cli/atlas/quickstart/quick_start.go
andreaangiolillo: See: https://github.com/mongodb/mongocli/pull/512#discussion_r590393087. We wouldn’t go inside the for loop and the function would return `""`

internal/cli/atlas/quickstart/description.go
```diff
@@ -0,0 +1,19 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+package quickstart
+
+const (
+	QuickStart     = "Create and access an Atlas Cluster."
+	LongQuickStart = "This command creates a cluster, adds your public IP to the atlas access list and creates a db user to access your MongoDB instance. "
+)
```
gssbzn, line 19: I'm just remembering we don't do description files anymore so let's inline these

internal/cli/atlas/quickstart/prompt.go
```diff
@@ -0,0 +1,91 @@
...
+}
+
+func newRegionQuestions(region, provider string) *survey.Question {
+	if region != "" {
+		return nil
+	}
+	return &survey.Question{
+		Name: "region",
+		Prompt: &survey.Select{
+			Message: "Insert the physical location of your MongoDB cluster",
```
gssbzn, line 42:
```suggestion
			Message: "Select the physical location of your MongoDB cluster",
```

internal/cli/atlas/quickstart/quick_start.go
```diff
@@ -0,0 +1,357 @@
...
+	var target *atlas.ErrorResponse
+
+	if err != nil && errors.As(err, &target) {
+		if target.ErrorCode == "USERNAME_NOT_FOUND" {
+			return nil
+		}
+		return err
+	}
+
+	return errors.New("a user with this username already exists")
```
gssbzn, line 292: [nit] please let add the username to the error, it may be helpful if the person doesn't remember what they typed or even if they had a typo

internal/cli/atlas/quickstart/quick_start.go
```diff
@@ -0,0 +1,357 @@
...
+		opts.DBUserPassword = pwd
+	}
+
+	return nil
+}
+
+func (opts *Opts) validateUniqueUsername(val interface{}) error {
+	username, ok := val.(string)
+	if !ok {
+		return errors.New("the username is not valid")
```
gssbzn, line 279: [nit] please add the value to the error so people can read what they typed

internal/cli/atlas/quickstart/quick_start.go
gssbzn: my question stand, let's say there's a network error, you don't get the list, it probably is null, or empty, what happens? this method should return an error as well as an empty string if somethings goes wrong you may not be able to safely assume a name

gssbzn (commented):
some minor comments, mostly copy

internal/cli/atlas/quickstart/quick_start.go
andreaangiolillo: in case of an error this piece of code `if clusters, ok := cs.([]atlas.Cluster); ok {` would be false so the function would return an empty string. Therefore the UI would show the message `"Insert the cluster name"` instead of `"Insert the cluster name [Press Enter to use the auto-generated name '%s']"`

```
	clusterName := opts.ClusterName
	if clusterName == "" {
		clusterName = opts.newClusterName()
		if clusterName != "" {
			message = fmt.Sprintf("Insert the cluster name [Press Enter to use the auto-generated name '%s']", clusterName)
		}
		qs = append(qs, newClusterNameQuestion(clusterName, message))
	}
```

gssbzn (approved):
LGTM thanks for all the work put into this :shipit:

PR #521, opened 2020-11-12
CLOUDP-76844: Discover replica-set protocol when FCV is only set in the processes

gssbzn:
## Proposed changes

_Jira ticket:_ CLOUDP-76844

## Checklist

### DIFF
--- a/internal/convert/rs_config.go
+++ b/internal/convert/rs_config.go
@@ -80,7 +80,8 @@ func (c *RSConfig) patchConfigServer(out *opsmngr.AutomationConfig, name string)
 func (c *RSConfig) protocolVer() (string, error) {
 	fcVersion := c.FCVersion
 	if fcVersion == "" {
-		// search per process, this may be the case whe users do a get first and update it
+		// search per process, this may be the case when users get a cluster description,
+		// manually update it and then try to apply that updated config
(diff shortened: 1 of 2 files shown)


internal/convert/rs_config.go
```diff
@@ -77,10 +78,20 @@ func (c *RSConfig) patchConfigServer(out *opsmngr.AutomationConfig, name string)
 // protocolVer determines the appropriate protocol based on FCV
 // returns "0" for versions <4.0 or "1" otherwise
 func (c *RSConfig) protocolVer() (string, error) {
-	if c.FCVersion == "" {
-		return "", nil
+	fcVersion := c.FCVersion
+	if fcVersion == "" {
+		// search per process, this may be the case whe users do a get first and update it
```
andreaangiolillo, line 83:
```suggestion
		// search per process, this may be the case when users do a get first and update it
```

[q] Could you explain a bit more what you mean here with "get first and update it"? Just to make clear what's the scenario here

andreaangiolillo (dismissed):
LGTM, I have one question.

internal/convert/rs_config.go
gssbzn: updated

andreaangiolillo (approved):
LGTM, Thanks for small change

PR #533, opened 2020-11-27
Added featurePolicies e2e tests

andreaangiolillo:
## Proposed changes

Added featurePolicies e2e tests

## Checklist

### DIFF
--- a/e2e/cloud_manager/feature_policies_test.go
+++ b/e2e/cloud_manager/feature_policies_test.go
@@ -1,0 +1,109 @@
+// Copyright 2020 MongoDB Inc
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+// +build e2e cloudmanager,generic
+
+package cloud_manager_test
+
+import (
+	"encoding/json"
+	"os"
+	"os/exec"
+	"testing"
+
+	"github.com/mongodb/mongocli/e2e"
+	"go.mongodb.org/ops-manager/opsmngr"
+)
+
+func TestFeaturePolicies(t *testing.T) {
+	const policyDisableSetMongoDConfig = "EXTERNALLY_MANAGED_LOCK"
+	const policyDisableUserManagement = "DISABLE_USER_MANAGEMENT"
+
+	cliPath, err := e2e.Bin()
+	if err != nil {
+		t.Fatalf("unexpected error: %v", err)
+	}
+
+	var externalManagSystemName string
+
+	t.Run("List", func(t *testing.T) {
+		cmd := exec.Command(cliPath,
+			entity,
+			featurePolicies,
+			"list",
+			"-o=json",
+		)
+
+		cmd.Env = os.Environ()
+		resp, err := cmd.CombinedOutput()
+
+		if err != nil {
+			t.Fatalf("unexpected error: %v, resp: %v", err, string(resp))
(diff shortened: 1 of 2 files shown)


e2e/cloud_manager/feature_policies_test.go
```diff
@@ -0,0 +1,114 @@
...
+		externalManagSystemName = policy.ExternalManagementSystem.Name
+	})
+
+	t.Run("Update", func(t *testing.T) {
+		cmd := exec.Command(cliPath,
+			entity,
+			featurePolicies,
+			"update",
+			"--name",
+			externalManagSystemName,
```
gssbzn, line 72: I'm I bit worry about running this in parallel (multiple tests) as if we may not be getting as response from the current test but from a parallel one
I think the `externalManagSystemName` doesn't matter so we could do these as random values
@colm-quinn what do you think?

gssbzn (dismissed):
LGTM but let's hear from colm in case we can improve this a littel

e2e/cloud_manager/feature_policies_test.go
colm-quinn: > I'm I bit worry about running this in parallel (multiple tests) as if we may not be getting as response from the current test but from a parallel one

Good spot. Last write wins with the endpoint, so if you're running in parallel I'd suggest creating a randomised group, so you can ensure the update is only from your test call.

> I think the externalManagSystemName doesn't matter so we could do these as random values
@colm-quinn what do you think?

Yes to randomised names so we can validate, however, in reference to my point above, that will just ensure the correct update, but still be potentially flaky.

e2e/cloud_manager/feature_policies_test.go
```diff
@@ -0,0 +1,114 @@
...
+
+		var policy *opsmngr.FeaturePolicy
+		if err := json.Unmarshal(resp, &policy); err != nil {
+			t.Fatalf("unexpected error: %v", err)
+		}
+
+		if len(policy.Policies) == 0 {
+			t.Error("No policies found")
+		}
+
```
colm-quinn, line 95: [q] What about ensuring only the requested policies were written.

```suggestion
if len(policy.Policies) != 2 {
			t.Error("Policy count mismatch")
		}
```

e2e/cloud_manager/feature_policies_test.go
```diff
@@ -0,0 +1,114 @@
...
+	t.Run("Update", func(t *testing.T) {
+		cmd := exec.Command(cliPath,
+			entity,
+			featurePolicies,
+			"update",
+			"--name",
+			externalManagSystemName,
+			"--policy",
+			policyExternallyManagedLock,
```
colm-quinn, line 74: [q] How do you specify the `disabledParams` for a policy with this command line? Those are needed for the `DISABLE_SET_MONGOD_CONFIG` policy in particular and should be settable.

colm-quinn (changes requested):
I agree we should make sure parallel tests don't impact on each other.

Also I'd like to see disabledParams tested for the DISABLE_SET_MONGOD_CONFIG, as that is a use case for the operator and it would be useful if the cli can update the params.
 
If that feature doesn't work, then we should open a ticket.

e2e/cloud_manager/feature_policies_test.go
andreaangiolillo: we actually don't allow the user to set `disabledParams` from the command line. I am not sure if we want to add this flag as part of this PR though. @gssbzn what do you think?

gssbzn: Let's open a ticket and bring it for planning, let's focus on test for this PR

colm-quinn: Yup, def a separate ticket!

gssbzn (approved):
LGTM thanks for those fixes

e2e/cloud_manager/feature_policies_test.go
andreaangiolillo: [CLOUDP-77797](https://jira.mongodb.org/browse/CLOUDP-77797) has been created.

colm-quinn (approved):
LGTM, thanks for the updates! Groot approves

PR #541, opened 2020-12-03
CLOUDP-78065: [mongocli] atlas logs download doesn’t work if --out is not provided

gssbzn:
## Proposed changes

_Jira ticket:_ CLOUDP-78065

## Checklist

### DIFF
--- a/internal/cli/atlas/logs/download.go
+++ b/internal/cli/atlas/logs/download.go
@@ -65,10 +65,11 @@ func (opts *DownloadOpts) Run() error {
 	return f.Close()
 }
 
-func (opts *DownloadOpts) initDefaultOut() {
+func (opts *DownloadOpts) initDefaultOut() error {
 	if opts.Out == "" {
 		opts.Out = strings.ReplaceAll(opts.name, ".gz", ".log.gz")
 	}
+	return nil
 }
 
 func (opts *DownloadOpts) newDateRangeOpts() *atlas.DateRangetOptions {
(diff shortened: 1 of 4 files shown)


internal/cli/atlas/logs/download.go
```diff
@@ -65,10 +65,11 @@ func (opts *DownloadOpts) Run() error {
 	return f.Close()
 }
 
-func (opts *DownloadOpts) initDefaultOut() {
+func (opts *DownloadOpts) initDefaultOut() error {
 	if opts.Out == "" {
 		opts.Out = strings.ReplaceAll(opts.name, ".gz", ".log.gz")
 	}
+	return nil
 }
```
andreaangiolillo, line 73: [q] in which scenario this function would return an error?

andreaangiolillo (commented):
I have one question but in general, I don't see which was the cause of the error

internal/cli/atlas/logs/download.go
gssbzn: never, this is just so this function can now be used in
```go
opts.PreRunE(opts.ValidateProjectID, opts.initStore, opts.initDefaultOut)
```

internal/cli/atlas/logs/download.go
```diff
@@ -89,12 +90,11 @@ func DownloadBuilder() *cobra.Command {
...
 		PreRunE: func(cmd *cobra.Command, args []string) error {
-			opts.initDefaultOut()
-			return opts.PreRunE(opts.ValidateProjectID, opts.initStore)
-		},
-		RunE: func(cmd *cobra.Command, args []string) error {
 			opts.host = args[0]
 			opts.name = args[1]
+			return opts.PreRunE(opts.ValidateProjectID, opts.initStore, opts.initDefaultOut)
+		},
+		RunE: func(cmd *cobra.Command, args []string) error {
```
gssbzn, line 97: @andreaangiolillo the error is here, I was calling `initDefaultOut` before `opts.name = args[1]` had any value so it was being overwritten by the Run command

andreaangiolillo: oh got it, thanks for the clarification

andreaangiolillo (approved):
LGTM

PR #542, opened 2020-12-07
NOJIRA-0: update .gitignore for IntelliJ and Makefile for downloading go.mod dependencies

shalka:
## Proposed changes

_Jira ticket:_ CLOUDP-#

Closes #[issue number]

## Checklist

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -24,3 +24,7 @@ gon.json
 
 # We don't want to commit env variables
 *.env
+
+# IntelliJ
+.idea/
+
--- a/Makefile
+++ b/Makefile
@@ -24,8 +24,13 @@ export PATH := ./bin:$(PATH)
 export GO111MODULE := on
 export E2E_BINARY
 
+.PHONY: deps
+deps:  ## Download go module dependencies
+	@echo "==> Installing go.mod dependencies..."
+	go mod download
+
 .PHONY: setup
-setup:  ## Install dev tools
+setup: deps ## Install dev tools
 	@echo "==> Installing dependencies..."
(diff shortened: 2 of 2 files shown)


.gitignore
```diff
@@ -24,3 +24,7 @@ gon.json
 
 # We don't want to commit env variables
 *.env
+
+# IntelliJ
```
gssbzn, line 28: [not-blocking] I personally prefer to set things like personal dev environments in a global gitignore but I think this is the second time someone tries to commit this so I'm open to make it official

Makefile
```diff
@@ -24,8 +24,13 @@ export PATH := ./bin:$(PATH)
...
 export E2E_BINARY
 
+.PHONY: deps
+deps:  ## Download go module dependencies
+	@echo "==> Installing go.mod dependencies..."
+	go mod download
+
 .PHONY: setup
-setup:  ## Install dev tools
+setup: deps ## Install dev tools
```
gssbzn, line 33: This change right now is breaking our linting step in evergreen. I have not looked at the [evergreen file](https://github.com/mongodb/mongocli/blob/master/build/ci/evergreen.yml) but I guess we are not setting up the go tool chain for that step given we just needed to download a binary and run it and now we would need a go environment 

With that said do we really need this change? We've been relaying on `make build`, `make install` or `make test` to also download dependencies, `golangci-lint` is just a bit special as they don't recommend to install it via deps

If you want to keep this change let's then try to fix evergreen before we merge this

gssbzn (changes requested):
Hi and thanks for the contributions, right now the changes to the makefile are breaking our CI so let's try to address that before we can look this any further

gssbzn:
Given all the time that has passed I'll close this PR, feel free to reopen it if the latests changes we did to the Makefile don't cover your use case

PR #557, opened 2021-01-08
CLOUDP-79643: Running generic tests on ops manager 4.4 and 4.2

andreaangiolillo:
## Proposed changes

_Jira ticket:_ CLOUDP-79643: Running generic tests on ops manager 4.4 and 4.2

## Checklist

## Further comments

Evergreen Tasks:

- [ops_manager_generic_4_4_e2e](https://evergreen.mongodb.com/task/mongocli_master_e2e_generic_ops_manager_generic_4_4_e2e_patch_79f568234a8243f4661294696d7a2ffd9930bf8b_60183443e3c3311e5170bb8d_21_02_01_17_02_59)
- [ops_manager_generic_4_2_e2e](https://evergreen.mongodb.com/task/mongocli_master_e2e_generic_ops_manager_generic_4_2_e2e_patch_79f568234a8243f4661294696d7a2ffd9930bf8b_60183443e3c3311e5170bb8d_21_02_01_17_02_59)


build/ci/install-mms-spawn-host.sh
```diff
@@ -0,0 +1,60 @@
...
+
+  #install ego
+  curl -sL https://raw.githubusercontent.com/mongodb-labs/ego/master/install.sh | bash
+
+  source ~/.bashrc
+
+  #install mms
+  ego ops_manager_install_version --version 4.4.6 --mongodb-version 4.2.8
+
+  exit
```
andreaangiolillo, line 57: here I am installing ego on the spawn host and then use ego to install mms

build/ci/set-up-mms.sh
```diff
@@ -0,0 +1,106 @@
...
+# This mongocli command returns an error when the user has been created with mongocli om owner create but the project is created anyway
+# More info: https://jira.mongodb.org/browse/CLOUDP-76824
+set +e
+echo "create project"
+./bin/mongocli iam projects create myProj --orgId "${organizationID}" -o json
+
+set -e
+echo "get project id"
+./bin/mongocli iam project list --orgId "${organizationID}" -o json > project.json
+cat project.json
```
andreaangiolillo, line 106: I have to call the `project list` because the previous command fails with 500 error https://jira.mongodb.org/browse/CLOUDP-76824

build/ci/set-up-mms.sh
```diff
@@ -0,0 +1,106 @@
...
+echo "set service"
+  ./bin/mongocli config set service ops-manager
+
+echo "set ops_manager_url"
+  ./bin/mongocli config set ops_manager_url "http://${host}:9080/"
+done
+
+echo "create first user"
+./bin/mongocli om owner create --firstName evergreen --lastName evergreen --email evergreenTest@gmail.com --password "evergreen1234_" -o json > apikeys.json
+
```
andreaangiolillo, line 52: This endpoint returns the programmatic keys only for mms 4.4 or later. I am not sure how we can get the programmatic keys for 4.2, maybe we can ask for a backport.

build/ci/run-e2e-test-mms.sh
```diff
@@ -0,0 +1,100 @@
...
+    hosts = json.load(hostsfile)
+    print(hosts[0]["dns_name"])
+EOF
+
+)
+
+export MCLI_OPS_MANAGER_URL="http://${host}:9080/"
+
+cd ..
+cd ..
```
gssbzn, line 45: [nit] see: https://en.wikipedia.org/wiki/Pushd_and_popd

build/ci/evergreen.yml
```diff
@@ -92,6 +92,24 @@ functions:
           TEST_CMD: gotestsum --junitfile e2e-tests.xml --format standard-verbose --
           <<: *go_options
         command: make e2e-test
+  "e2e test mms":
```
gssbzn, line 95: [nit] mms is an internal name, let's please call it ops manager

build/ci/set-up-mms.sh
```diff
@@ -0,0 +1,103 @@
...
+
+echo "set service"
+  ./bin/mongocli config set service ops-manager
+
+echo "set ops_manager_url"
+  ./bin/mongocli config set ops_manager_url "http://${host}:9080/"
+
+echo "create first user"
+./bin/mongocli om owner create --firstName evergreen --lastName evergreen --email evergreenTest@gmail.com --password "evergreen1234_" -o json > apikeys.json
```
gssbzn, line 48: [nit] couple of things here,
1) can we handle service and base url via ENV variables/evg expansions? we already do this for atlas and cloud manager

2) when creating the first user can we use a random password here? if not I'd rather handle that password via and evg expansion

3) instead of output as json you could use a go template in the toml format of the config and directly write the config from this command, check how we do something like this for gon, https://github.com/mongodb/mongocli/blob/79f568234a8243f4661294696d7a2ffd9930bf8b/build/ci/evergreen.yml#L213

build/ci/set-up-mms.sh
gssbzn: it's ok if we don't fully test 4.2, we offer a best effort for 4.2 but as we move to 5.0 we need to keep compatibility with 4.4

build/ci/set-up-mms.sh
```diff
@@ -0,0 +1,103 @@
...
+    print(user["programmaticApiKey"]["privateKey"])
+EOF
+
+)
+
+echo "set public_api_key"
+./bin/mongocli config set public_api_key "${publicKey}"
+
+echo "set private_api_key"
+./bin/mongocli config set private_api_key "${privateKey}"
```
gssbzn, line 77: this gores with previous comment that we can simplify all this

build/ci/set-up-mms.sh
gssbzn: a workaround for that error should be to invite the global owner to the first org

build/ci/run-e2e-test-mms.sh
```diff
@@ -0,0 +1,100 @@
...
+  cat <<EOF | python - project.json
+import sys
+import json
+with open(sys.argv[1]) as jsonfile:
+    proj = json.load(jsonfile)
+    print(proj["results"][0]["id"])
+EOF
+
+)
+export MCLI_PROJECT_ID
```
gssbzn, line 97: it seems like you generate the config file but at the same time you use env variables, I'd rather use env variables so another thing you can do is when running ownwer create, org create and such is to use custom go templates to write something like -o="go-templatre=export MCLI_..." > setting.bash, and then run that bash before echa test in silent mode

gssbzn (commented):
this is a great start left some recommendations here and there

build/ci/set-up-mms.sh
andreaangiolillo: I have tried here https://evergreen.mongodb.com/task_log_raw/mongocli_master_e2e_generic_ops_manager_generic_e2e_patch_60349802f2b9d4a011e9f9f8f189cd2a814b0911_6012ff01850e6109c8a31a00_21_01_28_18_14_36/9?type=T but I got the same error.

```
 [2021/01/29 09:57:34.196] invite owner to organization
 [2021/01/29 09:57:34.196] The user 'evergreentest1@gmail.com' has been invited.
 [2021/01/29 09:57:34.196] Invited users do not have access to the project until they accept the invitation.
 [2021/01/29 09:57:34.391] Error: POST http://ec2-54-226-39-245.compute-1.amazonaws.com:9080/api/public/v1.0/groups: 500 (request "UNEXPECTED_ERROR") Unexpected error.
 [2021/01/29 09:57:34.516] create project
 [2021/01/29 09:57:34.516] get project id
```

gssbzn: you can set up om to skip invitations, https://docs.opsmanager.mongodb.com/current/reference/configuration/index.html#mms.user.bypassInviteForExistingUsers

Also please don't use emails you don't control `evergreentest1@gmail.com`

andreaangiolillo: Could you suggest an email that I can use here?

build/ci/run-e2e-test-mms.sh
andreaangiolillo: I am not sure I understand here what you mean. The variables exported from this bash script are not visible in the session that runs the tests.

build/ci/evergreen.yml
```diff
@@ -108,6 +108,22 @@ functions:
...
+        include_expansions_in_env:
+          - MCLI_SERVICE
+          - TEST_CMD
+          - E2E_TAGS
+        working_dir: src/github.com/mongodb/mongocli
+        env:
+          XDG_CONFIG_HOME: ${workdir}
+          TEST_CMD: gotestsum --junitfile e2e-tests.xml --format standard-verbose --
+          <<: *go_options
+        command: make e2e-test
```
andreaangiolillo, line 126: I am wondering if I can reuse `"e2e test"` 🤔

andreaangiolillo: I am running a patch to see if I can. I will update accordingly once it has finished

andreaangiolillo: nope 😞 https://evergreen.mongodb.com/task/mongocli_master_e2e_generic_ops_manager_generic_4_2_e2e_patch_60349802f2b9d4a011e9f9f8f189cd2a814b0911_60183a5c3e8e861a46ea7f7f_21_02_01_17_29_10

build/ci/set-up-ops-manager.sh
```diff
@@ -0,0 +1,81 @@
...
+  service = "${ops_manager_service:?}"
+  public_api_key = "${email}"
+  private_api_key = "${MCLI_PRIVATE_API_KEY}"
+  org_id = "${MCLI_ORG_ID}"
+  project_id = "${MCLI_PROJECT_ID}"
+
+EOF
+
+echo "print mongocli.toml"
+cat "${XDG_CONFIG_HOME}/mongocli.toml"
```
gssbzn, line 81: [nit] please remove this

build/ci/set-up-ops-manager.sh
```diff
@@ -0,0 +1,81 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -euo pipefail
+
+while getopts 'h:' opt; do
+  case ${opt} in
+  h) hostsFile="${OPTARG}" ;; # Output of Evergreen host.list
+  *) exit 1 ;;
```
gssbzn, line 22: [nit] same as above, it's nice to have usage if it fails

gssbzn (changes requested):
Thanks for the work deploying Ops manager here, left a couple of questions and recommendations

build/ci/evergreen.yml
gssbzn: it's failing with an error running the command, what's the error?

build/ci/evergreen.yml
```diff
@@ -150,6 +166,37 @@ functions:
           LC_GROUP_ID: ${automation_agent_mms_group_id}
           BASE_URL: ${mcli_ops_manager_url}
         args: ['-u', 'ubuntu', '-i', 'ssh_id', '-h', 'hosts.json']
+  "install ops manager 4.4":
```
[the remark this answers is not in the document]
andreaangiolillo: how do I know when I want to install 4.2 vs 4.4 inside a script? I think that is much easier to define a new evergreen operation to handle this instead using bash. (I personally prefer to use bash as less as possible)

build/ci/set-up-ops-manager.sh
```diff
@@ -0,0 +1,81 @@
...
+./bin/mongocli config set private_api_key "${MCLI_PRIVATE_API_KEY}"
+
+echo "create organization"
+MCLI_ORG_ID=$(./bin/mongocli iam organizations create myOrg -o="go-template={{.ID}}")
+
+echo "create project"
+MCLI_PROJECT_ID=$(./bin/mongocli iam projects create myProj --orgId "${MCLI_ORG_ID}" -o="go-template={{.ID}}")
+
+
+cat <<EOF > "${XDG_CONFIG_HOME}/mongocli.toml"
```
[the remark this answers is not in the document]
andreaangiolillo: I am not following. Could you give me more information on this, please (maybe with an example)? I am setting all these variables (`MCLI_ORG_ID`, `MCLI_PROJECT_ID`, `MCLI_PRIVATE_API_KEY` etc) so that I can create the toml.

build/ci/evergreen.yml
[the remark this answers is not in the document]
gssbzn: you don't need bash at all, look how we do `"e2e test"`, is a function and we pass different parameters depending if it's Atlas or cloud manager, same applies here, you can have a single `install ops manager` function that receives  `ARCHIVE` as a parameter. this goes along with your comment on replacing `"e2e test ops manager":` with the already existing `"e2e test"`, why does it fail? it probably misses a couple of settings that you need to pass to the call of  `"e2e test"`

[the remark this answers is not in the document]
andreaangiolillo: got it! Thanks. Regarding `e2e test`, it is passing more expansions and env var than `e2e test ops manager` so my guess is that the e2e tests are using those vars instead of the toml.

build/ci/set-up-ops-manager.sh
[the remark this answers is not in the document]
gssbzn: you do a lot of `/bin/mongocli config set <something>` those are creating a new `mongocli.toml` that at the end you're not using as you're generating a new file here.
If the reason you are doing `config set` is to have settings for `./bin/mongocli om owner create` and the rest, you don't have to

you can do 
```bash
export MCLI_OPS_MANAGER_BASE_URL="${MCLI_OPS_MANAGER_URL}"
export MCLI_SERVICE="ops-manager"
./bin/mongocli om owner create
```
Or even better now that I write this, rename `MCLI_OPS_MANAGER_URL` to the correct name and you don't even need to export it here 
Also this is valid tho not better in case of secrets
```bash
MCLI_OPS_MANAGER_BASE_URL="value" MCLI_SERVICE="value" mongocli [command]
```
And no need to export, again exporting is fine just shoing you examples that you don't need to call `config set`

You can keep exporting as needed instead of `config set`

build/ci/evergreen.yml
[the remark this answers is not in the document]
gssbzn: `e2e test` just exports them, if they are not set when you call it then it should be ok, my bet is you are not passing the `XDG_CONFIG_HOME` which is the only difference I see, again you may want to debug that as I don't think we should need a new way to run the e2e tests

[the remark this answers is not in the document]
andreaangiolillo: it was failing because `e2e test` was setting `MCLI_OPS_MANAGER_URL`

build/ci/evergreen.yml
andreaangiolillo: it was failing because `e2e test` was setting `MCLI_OPS_MANAGER_URL`

build/ci/install-agent-spawn-host.sh
```diff
@@ -17,12 +17,19 @@
 
 set -euo pipefail
 
+_print_usage() {
+    echo
+    echo '  --i <keyfile>                SSH identity file'
+    echo '  --u <user>                   Username on the remote host'
+    echo '  --h <hostsFile>              Output of Evergreen host.list'
```
gssbzn, line 24: single dash
```suggestion
    echo '  -i <keyfile>                SSH identity file'
    echo '  -u <user>                   Username on the remote host'
    echo '  -h <hostsFile>              Output of Evergreen host.list'
```

build/ci/install-ops-manager-spawn-host.sh
```diff
@@ -0,0 +1,67 @@
...
+# limitations under the License.
+
+
+set -euo pipefail
+
+_print_usage() {
+    echo
+    echo '  --i <keyfile>                SSH identity file'
+    echo '  --u <user>                   Username on the remote host'
+    echo '  --h <hostsFile>              Output of Evergreen host.list'
```
gssbzn, line 24: same as above
```suggestion
    echo '  -i <keyfile>                SSH identity file'
    echo '  -u <user>                   Username on the remote host'
    echo '  -h <hostsFile>              Output of Evergreen host.list'
```

gssbzn (changes requested):
couple of small nits, other than that great work here 🎉

gssbzn:
Your tests are failing because of https://github.com/mongodb/mongocli/pull/598

gssbzn (approved):
LGTM, great work here

PR #558, opened 2021-01-11
CLOUDP-80160: mongocli] Update golangci-lint

gssbzn:
## Proposed changes

_Jira ticket:_ CLOUDP-80160

## Checklist

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -44,15 +44,18 @@ linters:
     - govet
     - ineffassign
     - interfacer
+    - makezero
     - misspell
     - nakedret
     - noctx
     - nolintlint
+    - predeclared
     - rowserrcheck
     - scopelint
     - staticcheck
     - structcheck
     - stylecheck
+    - thelper
     - typecheck
     - unconvert
     - unparam
--- a/internal/test/test.go
+++ b/internal/test/test.go
@@ -10,6 +10,7 @@ import (
 // CmdValidator helps validate a cobra.Command, verifying the number of sub commands
 // and the flags that are being defined for it
 func CmdValidator(t *testing.T, subject *cobra.Command, nSubCommands int, flags []string) {
+	t.Helper()
 	a := assert.New(t)
 	a.Len(subject.Commands(), nSubCommands)
 	if len(flags) == 0 {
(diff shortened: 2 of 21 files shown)


internal/test/test.go
```diff
@@ -10,6 +10,7 @@ import (
 // CmdValidator helps validate a cobra.Command, verifying the number of sub commands
 // and the flags that are being defined for it
 func CmdValidator(t *testing.T, subject *cobra.Command, nSubCommands int, flags []string) {
+	t.Helper()
```
andreaangiolillo, line 13: [q] what the purpose of this piece of code?

.golangci.yml
```diff
@@ -44,15 +44,18 @@ linters:
     - govet
     - ineffassign
     - interfacer
+    - makezero
```
andreaangiolillo, line 47: Why are we adding this check?
```
makezero:
    # Allow only slices initialized with a length of zero. Default is false.
```

Could you give me an example where `-predeclared` is used?
I honestly don’t understand this definition
```
  predeclared:
    # comma-separated list of predeclared identifiers to not report on
```

All checks seem to be enabled by default.
```
  thelper:
    # The following configurations enable all checks. It can be omitted because all checks are enabled by default.
    # You can enable only required checks deleting unnecessary checks.
```

andreaangiolillo (commented):
a couple of questions ❓ ❓

.golangci.yml
gssbzn: [predeclared](https://github.com/nishanths/predeclared), enabling this linter caught that the `delete` strings we had were shadowing the native go `delete` method, this linter will catch this kind of things

[makezero](https://github.com/ashanbrown/makezero), we did not have any violation for this and there's no harm in enabling it by default

[thelper](https://github.com/kulti/thelper), this is the reason I'm adding all those `t.Helper`, is a way go has to flag helper function for tests, this is useful as if we error within the helper then the line/method reported by go test is the helper method, when you call `t.Helper` then the line reported is the test that called the function

internal/test/test.go
gssbzn: https://golang.org/pkg/testing/#T.Helper

andreaangiolillo (approved):
LGTM. Thanks for the clarifications

PR #591, opened 2021-01-27
CLOUDP-79641: Add jsonpath as output

andreaangiolillo:
## Proposed changes

_Jira ticket:_ CLOUDP-79641

## Checklist

## Further comments

**Examples:**

```
./bin/mongocli iam organization apikeys list -o=json-path='$[? @.id=="5f3b9508320c484ff32cb7e8"].desc'
[CLI Gustavo ]
```
```
./bin/mongocli iam organization apikeys list -o=json-path='$[0].id'
5f3b9508320c484ff32cb7e8
```
```
./bin/mongocli atlas privateendpoint aws get 60100269480a253a4044da93 -o json-path='$.status'
WAITING_FOR_USER
```


internal/jsonpathwriter/jsonpathwriter.go
```diff
@@ -0,0 +1,44 @@
...
+	jsonString, err := json.Marshal(obj)
+	if err != nil {
+		return err
+	}
+
+	val := interface{}(nil)
+	err = json.Unmarshal(jsonString, &val)
+	if err != nil {
+		return err
+	}
```
andreaangiolillo, line 35: I have tried to pass obj directly to `jsonpath` but I keep getting errors

internal/cli/output_opts.go
```diff
@@ -105,8 +113,8 @@ func (opts *OutputOpts) Print(v interface{}) error {
 }
 
 // parseTemplate will try to find if the given format is a user given template, either by string or file and use it.
-// Current available user templates are  "go-template=Template string" and "go-template-file=path/to/template"
-func (opts *OutputOpts) parseTemplate() (string, error) {
+// Current available user templates are  "go-template=Template string", "go-template-file=path/to/template" and "json-path=path"
+func (opts *OutputOpts) parseTemplate() (outputTemplate, v string, e error) {
```
gssbzn, line 117: [super-nit] this method is called `parseTemplate` but json path is not technically a template, you could rename this to make the intention clearer or a better option in my opinion would be to split this in two methods, one to get the output type and value, another to read from file if file is given, I think we only error here because of a file and that would bring back the returned value per method to two

internal/jsonpathwriter/jsonpathwriter.go
gssbzn: this seems to be a limitation on json path, it only works with `map[string]interface{}` or `[]interface{}` but not with `interface{}`, weird I know

internal/jsonpathwriter/jsonpathwriter.go
```diff
@@ -0,0 +1,44 @@
...
+)
+
+func Print(w io.Writer, path string, obj interface{}) error {
+	jsonString, err := json.Marshal(obj)
+	if err != nil {
+		return err
+	}
+
+	val := interface{}(nil)
+	err = json.Unmarshal(jsonString, &val)
```
gssbzn, line 32: [nit] this is the right approach given json path limitations but you can optimize this a bit
```suggestion
	var val interface{}
	err = json.Unmarshal(jsonString, &val)
```
Saves you an instaciation you won't need later, also you could group this in the if `if err2 := json.Unmarshal(...` kind of thing but that's optional

internal/cli/output_opts.go
```diff
@@ -93,10 +95,16 @@ func (opts *OutputOpts) Print(v interface{}) error {
...
-	t, err := opts.parseTemplate()
+
+	outputFormat, t, err := opts.parseTemplate()
 	if err != nil {
 		return err
 	}
+
+	if outputFormat == jsonPath && t != "" {
+		return jsonpathwriter.Print(opts.ConfigWriter(), t, v)
+	}
```
gssbzn, line 106: [q] if `t` is empty and json path is not should we return an error? what would happen in that case?

gssbzn (changes requested):
Some small nits but in general this looks great

internal/cli/output_opts.go
andreaangiolillo: it returns something weird. I will add a check for an empty path in the writer
```
./bin/mongocli atlas privateendpoint aws get 60100269480a253a4044da93 -o json-path=
&{60100269480a253a4044da93   com.amazonaws.vpce.eu-central-1.vpce-svc-0d2136ac750f0244b  [] []   WAITING_FOR_USER}
```

internal/cli/output_opts_test.go
```diff
@@ -45,21 +45,22 @@ func TestOutputOpts_parseTemplate(t *testing.T) {
 			want:    "default",
 			wantErr: false,
```
gssbzn, line 46: since this function no longer returns an error you may want to remove this

gssbzn (changes requested):
almost there, couple of nits on the change for `outputTypeAndValue` mainly some simplification that can be done

internal/cli/output_opts.go
```diff
@@ -89,35 +91,47 @@ func (opts *OutputOpts) IsCygwinTerminal() bool {
...
+
+	outputType, val := opts.outputTypeAndValue()
+	if outputType == jsonPath {
+		return jsonpathwriter.Print(opts.ConfigWriter(), val, o)
+	}
+
+	t, err := opts.template(outputType, val)
 	if err != nil {
-		return err
+		return nil
```
gssbzn, line 106: [q] unintended?
```suggestion
		return err
```

internal/jsonpathwriter/jsonpathwriter.go
```diff
@@ -0,0 +1,47 @@
...
+	"encoding/json"
+	"fmt"
+	"io"
+
+	"github.com/PaesslerAG/jsonpath"
+)
+
+func Print(w io.Writer, path string, obj interface{}) error {
+	if path == "" {
+		return fmt.Errorf("path not found. Please use -o json-path='path'")
```
gssbzn, line 27: [nit] since you're not formating you can use `errors.New
```suggestion
		return errors.New("empty jsonpath")
```

gssbzn (commented):
two questions

gssbzn (approved):
LGTM thanks for adding this

PR #593, opened 2021-01-28
tests: test in parallel

gssbzn:
## Proposed changes

Our tests are becoming slower as we add more so adding a couple of `t.Parallel` where we can.
I don't think this is helping much so far comparing with previous runs but hopefully we can keep this pattern around

## Checklist


andreaangiolillo (approved):
LGTM

PR #594, opened 2021-02-01
CLOUDP-81627: Improve Atlas cluster create dynamic version for limited internet envs

gssbzn:
## Proposed changes

_Jira ticket:_ CLOUDP-81627

## Checklist

## Further comments

Default timeouts were too long for users using ops manager with no access to Atlas and just doing `mongocli --version` could take a long time

I'm handling this in two ways, not longer use of default timeouts and also lowering some of the ones we had. and we'll no longer include the Atlas branch if we don't need to, saving the call completely

### DIFF
--- a/cmd/mongocli/mongocli.go
+++ b/cmd/mongocli/mongocli.go
@@ -89,13 +89,13 @@ func rootBuilder() {
 		rootCmd.AddCommand(atlas.Builder())
 	}
 	if len(argsWithoutProg) == 0 || argsWithoutProg[0] == cloudmanager.Use {
-		cloudmanager.Builder()
+		rootCmd.AddCommand(cloudmanager.Builder())
 	}
 	if len(argsWithoutProg) == 0 || argsWithoutProg[0] == opsmanager.Use {
-		opsmanager.Builder()
+		rootCmd.AddCommand(opsmanager.Builder())
 	}
 	if len(argsWithoutProg) == 0 || argsWithoutProg[0] == iam.Use {
-		iam.Builder()
+		rootCmd.AddCommand(iam.Builder())
 	}
 	rootCmd.AddCommand(
 		completionCmd,
--- a/cmd/mongocli/mongocli.go
+++ b/cmd/mongocli/mongocli.go
@@ -32,6 +32,8 @@ import (
 )
 
 var (
+	profile string
+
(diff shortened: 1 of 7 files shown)


cmd/mongocli/mongocli.go
```diff
@@ -106,5 +118,8 @@ func initConfig() {
 }
 
 func main() {
+	cobra.EnableCommandSorting = false
+	cobra.OnInitialize(initConfig)
```
andreaangiolillo, line 122: [q] I am not sure why we need these two lines

cmd/mongocli/mongocli.go
```diff
@@ -65,30 +67,40 @@ no additional shell configuration is necessary, see https://docs.brew.sh/Shell-C
...
+	if len(argsWithoutProg) != 0 && argsWithoutProg[0] == "--version" {
+		return
+	}
+	rootCmd.AddCommand(cliconfig.Builder())
 
-func init() { //nolint:gochecknoinits // This is the cobra way
+	// We realistically only care about not adding Atlas to the chain of commands
+	// but given we can reuse the pattern for the rest we can save some initialization time here
+	if len(argsWithoutProg) == 0 || argsWithoutProg[0] == atlas.Use {
+		rootCmd.AddCommand(atlas.Builder())
```
andreaangiolillo, line 89: [q] how do you know that the first argument is the service? This question may be related to the fact that I don't know what `[1:]` returns.

andreaangiolillo (approved):
LGTM. Just a couple of questions

cmd/mongocli/mongocli.go
gssbzn: those lines used to be in L87 and L91 I just got rid of the init function
cobra by default sorts command alphabetically, we prefer to use our own order, so `cobra.EnableCommandSorting`

Also we call `initConfig` to initialize viper

cmd/mongocli/mongocli.go
gssbzn: if you call the cli like `mongocli atlas clusters` then `os.Args` is `["mongocli", "atlas", "clusters"]`, `os.Args[1:]` would be `["atlas", "clusters"]`

andreaangiolillo: thanks!!!

PR #601, opened 2021-02-03
CLOUDP-79644: Running replicaset/shared cluster tests on ops manager 4.4 and 4.2

andreaangiolillo:
## Proposed changes

_Jira ticket:_ CLOUDP-79644

## Checklist

## Further comments

[E2E Ops Manager Remote Host Tests](https://evergreen.mongodb.com/build/mongocli_master_e2e_ops_manager_remote_patch_bdfed7240e53f4c6cd951f66169e1f950dae45f5_601a7091850e611ed2ce52da_21_02_03_09_44_50)


andreaangiolillo:
replica set and sharded clusters tests are failing because of this 
`400 (request "INVALID_ATTRIBUTE") Invalid attribute onlineArchiveModules specified.`
Do we want to want to have the sharded/replicaset tests for 4.2 as well? if so I think we need to create new e2e tests for 4.2

gssbzn:
4.2 seems broken, and I think I know what it may be, we may need a change on the client

build/ci/install-agent-spawn-host.sh
```diff
@@ -33,6 +33,35 @@ while getopts 'i:h:g:u:a:b:' opt; do
   esac
 done
 
+
+flags=()
+# get the information that you need from the toml file if exists
+if test -f "${XDG_CONFIG_HOME}/mongocli.toml"; then
```
gssbzn, line 39: [q] why not `[[ ]]` instead of test?

build/ci/install-agent-spawn-host.sh
```diff
@@ -33,6 +33,35 @@ while getopts 'i:h:g:u:a:b:' opt; do
...
+    fi
+
+    if [[ "${name}" == "ops_manager_url" ]]; then
+      BASE_URL=$(echo "${value}" | tr -d '"')
+      export BASE_URL
+    fi
+  done < "${XDG_CONFIG_HOME}/mongocli.toml"
+
+  flags+=("--ops-manager")
+fi
```
gssbzn, line 61: [nb] fair play on getting this done on bash but this goes again with one of my recommendations before where you may better off creating a bash file of the format
```bash
export MCLI_*=...
```
and then executing that file over generating a toml not going to block this PR on this but something to consider

gssbzn (commented):
Great working getting this tests going, mostly a couple of questions and recommendations but can be added a future improvements.

Also 4.2 seems broken and it may need intervention on the client, feel free to comment 4.2 while we resolve that

gssbzn:
>replica set and sharded clusters tests are failing because of this
400 (request "INVALID_ATTRIBUTE") Invalid attribute onlineArchiveModules specified.
Do we want to want to have the sharded/replicaset tests for 4.2 as well? if so I think we need to create new e2e tests for 4.2

I had aq conversation the other day about this, that field changed from omit empty to always there recently in the client this was done by internal teams for cloud manager but we may need to revisit it to fix 4.2

gssbzn:
Client change that may be at fault https://github.com/mongodb/go-client-mongodb-ops-manager/pull/72

build/ci/evergreen.yml
```diff
@@ -794,6 +794,91 @@ tasks:
...
+      - func: "clone"
+      - func: "install gotestsum"
+      - func: "build"
+      - func: "deploy spawn host"
+      - func: ssh-ready
+      - func: "install ops manager"
+        vars:
+          ARCHIVE: ${ops_manager_4_4_archive}
+      - func: "set-up ops manager"
+      - func: "install automation agent"
```
gssbzn, line 814: food for thought, should we have a single ops manager e2e sanity tests? installing OM takes a long time here so I'm not sure if it would be better to install and run the whole e2e against it

build/ci/set-up-ops-manager.sh
```diff
@@ -69,6 +71,15 @@ cat <<EOF > "${XDG_CONFIG_HOME}/mongocli.toml"
...
+  agent_api_key = "${AGENT_API_KEY}"
+
+EOF
 
+popd
+cat <<EOF > automation_agent_settings.sh
+export BASE_URL=${MCLI_OPS_MANAGER_URL}
+export LC_AGENT_KEY=${AGENT_API_KEY}
+export LC_GROUP_ID=${MCLI_PROJECT_ID}
+export MCLI_SERVICE=${MCLI_SERVICE}
```
gssbzn, line 83: [nb] technically this should be safe to always set to `ops-manager`, ins't it?

gssbzn (dismissed):
LGTM, thanks for all the changes here, left a couple of question but good to merge on green

build/ci/set-up-ops-manager.sh
andreaangiolillo: yes

build/ci/evergreen.yml
andreaangiolillo: yes, I don't see why we may need different tasks. Deploying a cluster doesn’t require a lot of time and we can always create a new project if one test interferences with another one.

andreaangiolillo: we may want to wait for the fix on 4.2 before doing this so that we can run something on 4.2.

gssbzn: works for me, I'm not blocking on this just flagging it as something we may want to do

gssbzn (approved):
LGTM

PR #604, opened 2021-02-05
tests: pin the cache action as 2.1.4 introduced a bug

gssbzn:
See: https://github.com/actions/cache/issues/527

### DIFF
--- a/.github/workflows/go-test.yml
+++ b/.github/workflows/go-test.yml
@@ -21,7 +21,7 @@ jobs:
         with:
           go-version: ${{ matrix.golang }}
       - name: Cache Dependencies
-        uses: actions/cache@v2
+        uses: actions/cache@v2.1.3
         with:
           path: ~/go/pkg/mod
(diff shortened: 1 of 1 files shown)


colm-quinn:
Reading the linked ticket, it looks like this should fix itself as `v2` was moved back to 2.1.3. Should we still pin? 
https://github.com/actions/cache/issues/527#issuecomment-774152927

If you are pinning, can you add a CLOUDP ticket to track removing the pin once the issue is fixed.

Adding LGTM in case you're blocked, with the above cavaets.

colm-quinn (approved):

threebee (commented):
@gssbzn This change is not required anymore, I re-ran tests on macos-latest and they are working fine now. https://github.com/mongodb/mongocli/runs/1844724645

gssbzn:
This was kind of blocking https://github.com/mongodb/mongocli/pull/603 re running those tests and let's see if I can close this

PR #607, opened 2021-02-08
CLOUDP-82831: Improve argument validation

gssbzn:
## Proposed changes

_Jira ticket:_ CLOUDP-82831

## Checklist

## Further comments

This relates to a recent bug report where the users was trying to use emails instead of ids so adding a little validation here and there


internal/cli/require/require.go
```diff
@@ -49,6 +50,17 @@ func ExactArgs(n int) cobra.PositionalArgs {
...
+// any of those args is not an ObjectID.
+func ExactObjectIDArgs(n int) cobra.PositionalArgs {
+	return func(cmd *cobra.Command, args []string) error {
+		if err := ExactArgs(n)(cmd, args); err != nil {
+			return err
+		}
+		return objectIDArgs(args)
+	}
+}
+
```
andreaangiolillo, line 63: [q] Could this function takes `func objectIDArgs()` as input? so we won't have to define the same method if we want to apply other validations to the args.

internal/cli/require/require.go
```diff
@@ -81,6 +93,38 @@ func MinimumNArgs(n int) cobra.PositionalArgs {
...
+// MinimumNObjectIDArgs returns an error if there is not at least N args and
+// any of those args is not an ObjectID.
+func MinimumNObjectIDArgs(n int) cobra.PositionalArgs {
+	return func(cmd *cobra.Command, args []string) error {
+		if err := MinimumNArgs(n)(cmd, args); err != nil {
+			return err
+		}
+		return objectIDArgs(args)
+	}
+}
```
andreaangiolillo, line 105: [q] same question as `ExactObjectIDArgs`

andreaangiolillo (approved):
LGTM. Left some questions

internal/cli/require/require.go
gssbzn: At this point that would be a premature abstraction, specially if you mean here is to have something like `ExactCustomValidArgs(n int, f func([]string) error) cobra.PositionalArgs` and specially as you can already do that at the cobra level, we may just have enought of a excuse to make the objectID validation aa thing as many commands will take an objectID. What other custom validations you see here?

andreaangiolillo: I wasn’t thinking about any particular validation but I was wondering if this change could make sense here. Thanks for the answer!

PR #614, opened 2021-02-17
CLOUDP-82834: remove description files with only one or two descriptions

gssbzn:
## Proposed changes

_Jira ticket:_ CLOUDP-82834

## Checklist

## Further comments

One of the reasons checks are starting to take longer is the number of files growing (specially for things like fmt), we had these patter of creating the descriptions file to keep command description but they offer little to no advantage over inline values, specially when used for only one or two descriptions


andreaangiolillo (commented):
[q]why aren’t we removing the description file for all the commands? I personally think we should keep all the commands consistent. We should decide if we want to delete the descriptions files or creating only one file as it was before and we currently do with flags/usage.

gssbzn:
@andreaangiolillo I want to delete them all but this was already a large PR so started with the low hanging ones

andreaangiolillo:
@gssbzn let me help you with this. I will delete the description files for atlas

andreaangiolillo (approved):
LGTM

gssbzn:
>let me help you with this. I will delete the description files for atlas

Thanks, no rush on this

PR #616, opened 2021-02-18
CLOUDP-82909: Allow to append privileges if the privilege already exists

gssbzn:
## Proposed changes

_Jira ticket:_ CLOUDP-82909

## Checklist


internal/cli/atlas/customdbroles/update.go
```diff
@@ -63,30 +63,41 @@ func (opts *UpdateOpts) Run() error {
...
+	actionMap := make(map[string]atlas.Action)
+	for _, action := range existingActions {
+		actionMap[action.Action] = action
+	}
+	for i, action := range newActions {
+		if a, ok := actionMap[action.Action]; ok {
+			action.Resources = append(action.Resources, a.Resources...)
+			newActions[i] = action
+			delete(actionMap, action.Action)
+		}
```
andreaangiolillo, line 91: [q] what is going to happen if I have more than 2 actions that are the same? `FIND@db1.collection1,FIND@db2.collection2,FIND@db3.collection3`

internal/cli/atlas/customdbroles/update.go
```diff
@@ -63,30 +63,41 @@ func (opts *UpdateOpts) Run() error {
...
 
+func appendActions(existingActions, newActions []atlas.Action) []atlas.Action {
+	actionMap := make(map[string]atlas.Action)
+	for _, action := range existingActions {
+		actionMap[action.Action] = action
+	}
+	for i, action := range newActions {
+		if a, ok := actionMap[action.Action]; ok {
+			action.Resources = append(action.Resources, a.Resources...)
+			newActions[i] = action
```
andreaangiolillo, line 89: [not bloking] I personally don't like adding/removing/updating an element while iterating over a list.

internal/cli/atlas/customdbroles/create.go
```diff
@@ -28,7 +28,7 @@ import (
 	atlas "go.mongodb.org/atlas/mongodbatlas"
 )
 
-const createTemplate = "Custom Database Role {{.RoleName}} successfully created.\n"
+const createTemplate = "Custom database role '{{.RoleName}}' successfully created.\n"
```
andreaangiolillo, line 31: [q] Why can't we provide this functionality to the create command?

internal/cli/atlas/customdbroles/update_test.go
```diff
@@ -73,3 +73,149 @@ func TestUpdateBuilder(t *testing.T) {
...
+						},
+						{
+							Collection: "test",
+							Db:         "test2",
+						},
+					},
+				},
+			},
+		},
+	}
```
andreaangiolillo, line 212: [q] Can we provide a scenario where we have more than 2 actions that are the same? `FIND@db1.collection1,FIND@db2.collection2,FIND@db3.collection3`

andreaangiolillo (commented):
Left some questions

internal/cli/atlas/customdbroles/update.go
gssbzn: good catch but that's a different bug of the input so I'll addressed in a different way

internal/cli/atlas/customdbroles/create.go
gssbzn: what do you mean here?

andreaangiolillo: I am just wondering why we are not allowing the user to create a custom role with an action which is applied to multiple resources with the `create `command

something like :
`./bin/mongocli atlas customDBRoles create myrole5 --privilege FIND@db1.coll2,FIND@db1.ss,UPDATE@db2.coll3
Error: POST https://cloud-dev.mongodb.com/api/atlas/v1.0/groups/5e4e593f70dfbf1010295836/customDBRoles/roles: 400 (request "ATLAS_DUPLICATE_CUSTOM_ROLE_ACTION") A custom role cannot have duplicate actions.`

andreaangiolillo: but I see now that the `joinActions` should allow this

andreaangiolillo (approved):
LGTM. Thanks for the changes

internal/cli/atlas/customdbroles/create.go
gssbzn: Oh I thought for this comment it was about the line and I was confused

PR #625, opened 2021-03-03
CLOUDP-83921: Offer a suggestion when missing credentials

gssbzn:
## Proposed changes

_Jira ticket:_ CLOUDP-83921

## Checklist

## Further comments

This PR tries to do something similar to how we offer usage on missing args

```bash
mongocli config get
Error: "mongocli config describe" requires 1 argument

Usage:  mongocli config describe <name> [flags]
```

If there are no credentials then we get something like
```bash
mongocli iam projects ls
Error: missing credentials

To set credentials, run: mongocli config
```

### DIFF
--- a/internal/validate/validate.go
+++ b/internal/validate/validate.go
@@ -92,7 +92,7 @@ func ObjectID(s string) error {
 func Credentials() error {
 	if config.PrivateAPIKey() == "" || config.PublicAPIKey() == "" {
 		return fmt.Errorf(
-			"missing credentials\n\nStart with:  %s %s",
+			"missing credentials\n\nTo set credentials, run: %s %s",
 			config.ToolName,
 			"config",
 		)
--- a/internal/validate/validate.go
+++ b/internal/validate/validate.go
@@ -16,14 +16,12 @@ package validate
 
 import (
 	"encoding/hex"
-	"errors"
 	"fmt"
 	"net/url"
 	"strings"
 
-	"github.com/mongodb/mongocli/internal/search"
-
 	"github.com/mongodb/mongocli/internal/config"
+	"github.com/mongodb/mongocli/internal/search"
 )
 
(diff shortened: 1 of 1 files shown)


internal/validate/validate.go
```diff
@@ -96,7 +91,11 @@ func ObjectID(s string) error {
 // Credentials validates public and private API keys have been set
 func Credentials() error {
 	if config.PrivateAPIKey() == "" || config.PublicAPIKey() == "" {
-		return errors.New("missing credentials")
+		return fmt.Errorf(
+			"missing credentials\n\nStart with:  %s %s",
```
andreaangiolillo, line 95: [nit] I personally don't like `Start with`, I think that `Run "mongocli config" to set your credentials` would be more useful.

andreaangiolillo (dismissed):
LGTM

internal/validate/validate.go
gssbzn: I agree with you, I don't like `Start with` but I was trying to keep it in line with `Usage: <command>` over a phrase, I've asked docs for input here so I'll wait for their comments

melissamahoney-mongodb: I see what you mean about trying to keep the syntax the same, but like the phrase too. I think it gives more context. Maybe we could combine them into something like `To set credentials, run: mongocli config` ?

gssbzn: thanks, I think that may work, @andreaangiolillo what do you think?

andreaangiolillo: SGTM!

gssbzn:
@melissamahoney-mongodb @andreaangiolillo RFAL

andreaangiolillo (approved):
LGTM

melissamahoney-mongodb (approved):

PR #631, opened 2021-03-15
CLOUDP-84841: Allow to update security and setParams

gssbzn:
## Proposed changes

_Jira ticket:_ CLOUDP-84841

## Checklist

## Further comments

This seems to be from a time when we were not supporting these values, we would keep the old just in case, that's no longer true


andreaangiolillo (approved):
LGTM

PR #635, opened 2021-03-22
CLOUDP-84973: Quickstart without a default profile

andreaangiolillo:
## Proposed changes

_Jira ticket:_ CLOUDP-84973

## Checklist

## Further comments

```
./bin/mongocli atlas quickstart
missing default profile.
? Do you want to create an Atlas account [This will open www.mongodb.com on your browser]? Yes
? Do you want more information to set up your profile [This will open www.mongodb.com on your browser]? Yes
Error: missing credentials

To set credentials, run: mongocli config
```


internal/cli/atlas/quickstart/quick_start.go
```diff
@@ -450,8 +482,12 @@ func Builder() *cobra.Command {
 		Short: "Create and access an Atlas Cluster.",
 		Long:  "This command creates a cluster, adds your public IP to the atlas access list and creates a db user to access your MongoDB instance.",
 		PreRunE: func(cmd *cobra.Command, args []string) error {
+			if err := opts.ValidateProjectID(); err != nil {
```
gssbzn, line 485: this may not be enough check, you need to check the three values (project, public and private key) since you no longer do a "validate credentials"

It may not be the common case but all these are optional when doing `mongocli config` so you need to check that all have been set when running this command

gssbzn (changes requested):

internal/cli/atlas/quickstart/quick_start.go
andreaangiolillo: I moved the check of the project ID where it was before, I believe we should open the browser only if the API keys are not set.

gssbzn: agree

internal/cli/atlas/quickstart/quick_start.go
```diff
@@ -456,6 +459,35 @@ func askMongoShellAndSetConfig() error {
 	return nil
 }
 
+func askAtlasAccountAndProfile() error {
+	fmt.Println("missing default profile.")
```
gssbzn, line 463: [nit] let's output to stderr and change the copy given we changed the logic to check this
```suggestion
	_ = fmt.Printf(os.Stderr, "No API credentials set.\n")
```
(I don't remember if this is the correct syntax for staderr)

gssbzn (approved):
LGTM

PR #637, opened 2021-03-23
CLOUDP-85676: Update default value of clusterName and DBuser username

andreaangiolillo:
## Proposed changes

_Jira ticket:_ CLOUDP-85676

Closes #[issue number]

## Checklist


internal/cli/atlas/quickstart/quick_start.go
```diff
@@ -507,7 +497,7 @@ func (opts *Opts) newClusterName() string {
 	i := 0
 	if clusters, ok := cs.([]atlas.Cluster); ok {
 		for {
-			clusterName := "QuickstartCluster" + strconv.Itoa(i)
+			clusterName := autogeneratedName()
```
gssbzn, line 500: I think we can remove the  the loop and API call right?

gssbzn (changes requested):
I think there's more code that can be deleted

internal/cli/atlas/quickstart/quick_start.go
andreaangiolillo: yes, correct

internal/store/clusters.go
```diff
@@ -63,7 +63,6 @@ type AtlasClusterGetterUpdater interface {
 }
 
 type AtlasClusterQuickStarter interface {
-	ClusterLister
```
gssbzn, line 66: 💯

internal/cli/atlas/quickstart/quick_start.go
```diff
@@ -262,7 +260,7 @@ func (opts *Opts) askClusterOptions() error {
 	clusterName := opts.ClusterName
 	if clusterName == "" {
 		message := ""
-		clusterName = opts.newClusterName()
+		clusterName = autogeneratedName()
```
gssbzn, line 263: not for this PR but it would be interesting to see if we can set the default of the flag at runtime, like
```go
cmd.Flags().StringVar(&opts.ClusterName, flag.ClusterName, autogeneratedName(), usage.ClusterName)
```

I think I tried something like that before and cobra would not like it but I don't remember what the problem was back then

gssbzn (approved):
LGTM

internal/cli/atlas/quickstart/quick_start.go
andreaangiolillo: I think this may not work with our logic: now we are asking the cluster name/db user questions when the flags are empty, I am not sure how to adapt this logic with this change (maybe if the `clusterName` starts with `Quickstart-` ask the question?).

PR #638, opened 2021-03-24
CLOUDP-84975: Use the new atlas endpoint in the quickstart command

andreaangiolillo:
## Proposed changes

_Jira ticket:_ CLOUDP-84975

## Checklist


internal/store/cloud_provider_regions.go
```diff
@@ -0,0 +1,45 @@
+// Copyright 2020 MongoDB Inc
```
gssbzn, line 1:
```suggestion
// Copyright 2021 MongoDB Inc
```

internal/cli/atlas/quickstart/quick_start.go
```diff
@@ -86,11 +86,11 @@ const (
...
 
-// DefaultRegions represents the regions available for each cloud service provider
+// DefaultRegions represents the regions available for each cloud service provider for the M2 tier
 var DefaultRegions = map[string][]string{
-	"AWS":   {"US_EAST_1", "US_WEST_2", "AP_SOUTH_1", "AP_EAST_2", "EU_WEST_1", "EU_CENTRAL_1", "ME_SOUTH_1", "AF_SOUTH_1"},
-	"GCP":   {"CENTRAL_US", "CANADA_CENTRAL", "WESTERN_EUROPE", "ASIA_SOUTH_EAST", "SOUTH_AFRICA_NORTH", "UAE_NORTH"},
-	"AZURE": {"US_EAST_2", "US_WEST", "EUROPE_NORTH"},
+	"AWS":   {"US_EAST_1"},
+	"GCP":   {"CENTRAL_US"},
+	"AZURE": {"US_EAST_2"},
```
gssbzn, line 93: do we need to keep this? I assume we would remove it with the new API

internal/cli/atlas/quickstart/quick_start.go
```diff
@@ -284,6 +285,66 @@ func (opts *Opts) askClusterOptions() error {
 	return nil
 }
 
+func (opts *Opts) defaultRegions() []string {
+	cloudProviders, err := opts.store.CloudProviderRegions(opts.ConfigProjectID(), opts.Provider, tier, false)
+
+	if err != nil {
+		return DefaultRegions[strings.ToUpper(opts.Provider)]
+	}
```
gssbzn, line 293: Oh, I see you use it as a fallback, what are the chances that the api call will fail?

internal/cli/atlas/quickstart/quick_start.go
```diff
@@ -284,6 +285,66 @@ func (opts *Opts) askClusterOptions() error {
...
+		if v.Name != popularRegion {
+			defaultRegions[i] = v.Name
+			i++
+		}
+	}
+
+	return defaultRegions
+}
+
+func validateDefaultRegions(cloudProviders *atlas.CloudProviders) error {
```
gssbzn, line 325: [q] this seems like a test method? why would users need to see this?

internal/cli/atlas/quickstart/quick_start.go
```diff
@@ -284,6 +285,66 @@ func (opts *Opts) askClusterOptions() error {
...
+	i := 1
+
+	for _, v := range availableRegions {
+		if v.Name != popularRegion {
+			defaultRegions[i] = v.Name
+			i++
+		}
+	}
+
+	return defaultRegions
```
gssbzn, line 322: [nit] if you find the index of the default you could move it in place to be the first, this is a bit more expected in go

Another option instead of using your own index is using `append` to the array

Finally please consider creating a separate function to find default

gssbzn (commented):
Thanks for already adding this, some small nits before approving

internal/cli/atlas/quickstart/quick_start.go
andreaangiolillo: I kept this in case the API returns an error.

internal/cli/atlas/quickstart/quick_start.go
andreaangiolillo: Extremely low but I think that been prepared for the impossible is not a bad thing 😅

internal/cli/atlas/quickstart/quick_start.go
andreaangiolillo: What do you mean with  
> why would users need to see this?
The user doesn’t see anything, I use this function to ensure that I can access the array with indexes later

internal/cli/atlas/quickstart/quick_start.go
gssbzn: let's discuss over the error comment

internal/cli/atlas/quickstart/quick_start.go
gssbzn: low, but what would cause it to fail? would the user be able to continue with other API calls if this one fail?
My point here is, how likely is that if this fails then the next API calls will also fail and in turn we should probably not be so quiet about it

internal/cli/atlas/quickstart/quick_start.go
gssbzn: oh I see you don't do anything with the text of the error, so why so much effort on the error message?
Anyway my original questions stand, why test the API result? are you not able to trust the API?

andreaangiolillo: I don't like using this piece of code `cloudProviders.Results[0].InstanceSizes` without testing for possible NPE. No, I don't trust my code 😅 .

internal/cli/atlas/quickstart/quick_start.go
andreaangiolillo: yes, the user would be able to continue with other API calls with the current implementation.
I believe the user should be able to use the quickstart command even if the API call starts to fail. In addition, this strategy has been used also in other parts of the command.
Regarding what would cause this API to fail:
- changes in the atlas codebase which impacts the endpoint and have not been captured by the integration test
- no internet connection
- wrong API keys and/or projectID

For the last two points, the command will eventually fail

gssbzn: > the command will eventually fail anyway.

Why wait?what's the point of asking so many questions if at the end you're going to fails?

andreaangiolillo: okay, I will remove this piece of code

internal/cli/atlas/quickstart/quick_start.go
gssbzn: you'll get nil panics or NPE if the API returns an error. again this code reads like a test, so maybe this is probably not the best place for this code

use `len(cloudProviders.Results) > 0 && len(cloudProviders.Results[0].InstanceSizes) > 0` or soem combination of it `len(nil) ==0` btw

andreaangiolillo:
@gssbzn ready for review

internal/cli/atlas/quickstart/quick_start.go
```diff
@@ -284,6 +281,47 @@ func (opts *Opts) askClusterOptions() error {
...
 
+func (opts *Opts) defaultRegions() ([]string, error) {
+	cloudProviders, err := opts.store.CloudProviderRegions(opts.ConfigProjectID(), opts.Provider, tier, false)
+
+	if err != nil {
+		return nil, err
+	}
+
+	if len(cloudProviders.Results) == 0 || len(cloudProviders.Results[0].InstanceSizes) == 0 {
+		return nil, fmt.Errorf("no regions available")
```
gssbzn, line 292: [nit] if you don't need interpolation just use `errors.New`
```suggestion
		return nil, errors.New("no regions available")
```

internal/cli/atlas/quickstart/quick_start.go
```diff
@@ -284,6 +281,47 @@ func (opts *Opts) askClusterOptions() error {
...
+	}
+
+	availableRegions := cloudProviders.Results[0].InstanceSizes[0].AvailableRegions
+
+	var defaultRegions []string
```
gssbzn, line 297: [not blocking] FYI, you can still use make, 
```suggestion
	defaultRegions := make([]string, 0, len(availableRegions))
```

the third param is the capacity of the array but the it still has a len of zero so append works as you may expect
no need to change anything here just an FYI

gssbzn (commented):
almost there, one small nit to just use `errors.New` over `fmt.Errorf` if you don't need to format a string

gssbzn (approved):
LGTM

PR #640, opened 2021-03-25
CLOUDP-85812: use the atlas endpoint to get the public ip address of the user within quickstart

andreaangiolillo:
## Proposed changes

_Jira ticket:_ CLOUDP-85812

## Checklist


internal/store/store.go
```diff
@@ -247,7 +247,7 @@ func NewPrivateUnauth(c Config) (*Store, error) {
 	s.baseURL = atlas.CloudURL
 
 	if configURL := c.OpsManagerURL(); configURL != "" {
-		s.baseURL = s.apiPath(configURL)
+		s.baseURL = configURL
```
andreaangiolillo, line 250: drive-by fix

andreaangiolillo, line 250: drive-by fix

internal/net/net.go
```diff
@@ -14,44 +14,20 @@
 package net
 
 import (
-	"context"
-	"io/ioutil"
-	"net/http"
-	"strings"
+	"github.com/mongodb/mongocli/internal/config"
```
gssbzn, line 17: [q] should we keep this package? I feel we can be better off without and rely on our store package alone

internal/store/store.go
gssbzn: [q] are sure? this is used to create the owner and I'm pretty sure it was working as expected, remember that ops manager doesn't set up the `/api` segment as part of the base url

gssbzn: also this is used for the default cluster version, I'll add cluster deployment tests to see if this works

gssbzn: actually this may be right, I think this is only used for the default version and I think that one always uses teh default setting because of reasons

gssbzn (dismissed):
LGTM left a question about maybe removing the net package, also I've bene wanting to refactor how we create clients, it's growing wild

internal/cli/atlas/quickstart/quick_start.go
```diff
@@ -378,7 +377,7 @@ func (opts *Opts) askAccessListOptions() error {
 	}
 
 	message := ""
-	publicIP := net.IPAddress()
+	publicIP := store.IPAddress()
```
gssbzn, line 380: [q] any reason you can't do `opts.store.IPInfo()`?

gssbzn (approved):
LGTM

andreaangiolillo (commented):
@gssbzn I am not sure if your question is gone for a problem with github or if you deleted it. I will put my answer here anyway

I believe it wouldn’t work because I need to use `NewPrivate(config.Default())` to call `https://cloud-dev/api/private/ipInfo`. If I use directly `opts.store.IPInfo()`, I would call `https://cloud-dev/api/public/v1.0/api/private/ipInfo`.

andreaangiolillo (commented):
I have just some questions to understand completely these changes. LGTM once my questions are answers. Thanks for all these changes!

PR #646, opened 2021-04-04
CLOUDP-87128: Add the inventory of the e2e tests in mongocli

andreaangiolillo:
## Proposed changes
I have added [E2E-TESTS.md](https://github.com/mongodb/mongocli/blob/e2e_test_audit/e2e/E2E-TESTS.md) to keep track of how many commands have an e2e test and updated the PR template.

_Jira ticket:_ CLOUDP-87128

## Checklist

- [x] I have updated `e2e-test-audit.csv` (if a new command or e2e test has been added)


e2e-test-audit.csv
```diff
@@ -0,0 +1,292 @@
+Command,E2E-Atlas,E2E-OM,E2E-CM,Deprecated,Atlas,OM,CM,Location,E2E Test Location
```
gssbzn, line 1: [q] what are you plans on maintaining this and consider that this is a public repo and we welcome external contributors, this format seems hard to maintain specially when adding new commands?

gssbzn, line 1: also, a file like this should probably be in the `e2e` folder

e2e-test-audit.csv
```diff
@@ -0,0 +1,292 @@
...
+performanceAdvisor namespace list,,N,N,,X,X,X,internal/cli/performanceadvisor,
+performanceAdvisor slowQueryLogs list,,N,N,,X,X,X,,
+performanceAdvisor suggestedIndexes list,,N,N,,X,X,X,,
+,,,,,,,,,
+,,,,,,,,,
+,,,,,,,,,
+N,57,135,57,,,,,,
+Y,112,19,52,,,,,,
+Total,169,154,109,,,,,,
+% Coverage,66.27%,12.34%,47.71%,,,,,,
```
gssbzn, line 292: similar to above, are contributors expected to calculate this when updating this file?

gssbzn (commented):
Thanks for adding this, the idea seems promising but in its current format it looks to be adding extra complexity for both maintainers and external contributors and I worry specially about the later, I'd like to look into ways to either automate this file generation or have a simpler format that is easy to update for anyone opening a PR

e2e-test-audit.csv
andreaangiolillo: Given that this file is a `.csv`, it can be opened with any editor and easily updated

e2e-test-audit.csv
andreaangiolillo: Ideally yes, but they probably won’t

e2e-test-audit.csv
colm-quinn: I think the point here is not that it's easy to update, but that it's hard to know what update is needed. Think even if you look at the format in 6 months, will you find it easy to remember what each field does?

If we were using a file input like the above, a son would be more powerful as it can be more expressive and easier to validate.

gssbzn: even as a `.csv` simple text editors (VS code/Atom/Sublime) may show this in a not tabular (excel) way. I would say [markdown table](https://guides.github.com/features/mastering-markdown/#GitHub-flavored-markdown) could be a better presentation as you could preview the file in your text editor and not need something like excel to make sure you didn't break it

this also goes with my other comment on percentages, I doubt people will take care to update them, so should we keep them? 

I'd like to keep this requirement for an e2e inventory as simple and welcoming for someone not familiar with the codebase as possible so if we need to make changes to how we structure e2e tests to make this inventory even easier I'm ok with that

e2e/E2E-TESTS.md
```diff
@@ -0,0 +1,38 @@
+# E2E Test Inventory
+
+#### Fields Description
+
+- Command: the MongoDB CLI command without `mongocli atlas|om|cm`
```
andreaangiolillo, line 5: @gssbzn @colm-quinn Could you tell me what you think about this format? I would like to gather some feedback before adding all the commands

andreaangiolillo: you can have a look at the file here https://github.com/mongodb/mongocli/blob/e2e_test_audit/e2e/E2E-TESTS.md

colm-quinn: I like the inventory and markdown table. I'd probably summarise the Y/N steps into a single line so it reads more easily. Something like:

```
- E2E Atlas: 
    -  Possible values: ['Y'|'N']
    - Indicates if an e2e test for the command is present ('Y') or not present ('N') for Atlas
...
```

.github/pull_request_template.md
```diff
@@ -33,6 +33,7 @@ We're here to help! This is simply a reminder of what we are going to look for b
 - [ ] I have signed the [MongoDB CLA](https://www.mongodb.com/legal/contributor-agreement)
 - [ ] I have added tests that prove my fix is effective or that my feature works
 - [ ] I have added any necessary documentation (if appropriate)
+- [ ] I have updated `e2e-test-audit.csv` (if a new command or e2e test has been added)
```
colm-quinn, line 36: [np] It's not a csv anymore, as you're using markdown table.

gssbzn (approved):
Love this, I'd wait for any comment from @colm-quinn as someone less familiar with the code I think his input could be great here

.github/pull_request_template.md
gssbzn: I would also link to the
```suggestion
- [ ] I have updated [e2e/E2E-TESTS.md](e2e/E2E-TESTS.md) (if a new command or e2e test has been added)
```

colm-quinn: One last comment I have on the checklist is there is nothing to force the user to update the table. I see we have a code health check. In a follow-up check-in, I think it would be good to add a check that the table has been updated if a new test has been added, otherwise I can see it falling out of sync very easily.

colm-quinn:
> Love this, I'd wait for any comment from @colm-quinn as someone less familiar with the code I think his input could be great here

Just waiting on the final updates and will give this a review - all looking good so far!

.github/pull_request_template.md
andreaangiolillo: yes, you're right. What I would like to do in a follow-up PR is to automatically update this file. If I cannot achieve it I will add a check.

gssbzn (approved):
LGTM same condition on letting @colm-quinn have final say

e2e/E2E-TESTS.md
```diff
@@ -4,35 +4,309 @@
 
 - Command: the MongoDB CLI command without `mongocli atlas|om|cm` 
 - E2E Atlas:    
-    -  `Y`: the e2e test for the related command has been implemented for Atlas
-    -  `N`: no e2e test for the related command has been implemented for Atlas
+    -  Possible values: `('Y'|'N')`
```
colm-quinn, line 7: 💯

colm-quinn (approved):
LGTM, thanks for those changes, looks great!

.github/pull_request_template.md
colm-quinn: sgtm 👍

PR #660, opened 2021-04-13
task: refactor how we create stores

gssbzn:
## Proposed changes

Move from having 4 ways to create a new `Store` to a single one that takes functional options for configuration

## Checklist

## Further comments

This has been on my TODO list for a long time, every time we add a new `New*` method I'm reminded of it. This moves  from multiple `New` methods that know what changes to set of configurations you need to pass to the constructor `With*` to reduce the impact of this I have also added a group of presets that have the default configs we have been using so far

### DIFF
--- a/internal/store/store.go
+++ b/internal/store/store.go
@@ -50,7 +50,9 @@ type Store struct {
 	service       string
 	baseURL       string
 	caCertificate string
-	skipVerify    string
+	skipVerify    bool
+	username      string
+	password      string
 	client        interface{}
 }
 
@@ -66,7 +68,22 @@ var defaultTransport = &http.Transport{
 	ExpectContinueTimeout: expectContinueTimeout,
 }
 
-func customCATransport(ca []byte) http.RoundTripper {
+var skipVerifyTransport = &http.Transport{
+	ResponseHeaderTimeout: responseHeaderTimeout,
+	TLSHandshakeTimeout:   tlsHandshakeTimeout,
+	DialContext: (&net.Dialer{
+		Timeout:   timeout,
+		KeepAlive: keepAlive,
+	}).DialContext,
+	MaxIdleConns:          maxIdleConns,
+	MaxIdleConnsPerHost:   maxIdleConnsPerHost,
+	Proxy:                 http.ProxyFromEnvironment,
+	IdleConnTimeout:       idleConnTimeout,
+	ExpectContinueTimeout: expectContinueTimeout,
+	TLSClientConfig:       &tls.Config{InsecureSkipVerify: true}, //nolint:gosec // this is optional for some users,
+}
+
+func customCATransport(ca []byte) *http.Transport {
 	caCertPool := x509.NewCertPool()
 	caCertPool.AppendCertsFromPEM(ca)
 	tlsClientConfig := &tls.Config{ //nolint:gosec // we let users set custom certificates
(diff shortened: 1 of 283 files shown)


internal/store/store.go
```diff
@@ -89,235 +106,246 @@ func customCATransport(ca []byte) http.RoundTripper {
...
 	}
-	s.baseURL = atlas.CloudURL
+	options = append(options, NetworkPresets(c))
+	return Options(options...)
+}
 
-	if configURL := c.OpsManagerURL(); configURL != "" {
-		s.baseURL = configURL
+// New
+func New(opts ...Option) (*Store, error) {
```
gssbzn, line 297: main change

internal/store/store.go
```diff
@@ -89,235 +106,246 @@ func customCATransport(ca []byte) http.RoundTripper {
...
+}
+
+type Config interface {
+	CredentialsGetter
+	TransportConfigGetter
+	Service() string
+	OpsManagerURL() string
+}
 
+func PublicAuthenticatedPreset(c Config) Option {
```
gssbzn, line 260: if you need an authenticated store for the public API

internal/store/store.go
```diff
@@ -89,235 +106,246 @@ func customCATransport(ca []byte) http.RoundTripper {
...
+		options = append(options, WithBaseURL(configURL), WithPublicPathBaseURL())
 	}
+	options = append(options, NetworkPresets(c))
+	return Options(options...)
+}
 
-	client, err := defaultClient(c)
-	if err != nil {
-		return nil, err
+func PublicUnauthenticatedPreset(c Config) Option {
```
gssbzn, line 269: if you need a store with no authentication for the public API

internal/store/store.go
```diff
@@ -89,235 +106,246 @@ func customCATransport(ca []byte) http.RoundTripper {
...
+	if configURL := c.OpsManagerURL(); configURL != "" {
+		options = append(options, WithBaseURL(configURL), WithPublicPathBaseURL())
 	}
+	options = append(options, NetworkPresets(c))
+	return Options(options...)
+}
 
-	if err2 := s.setAtlasClient(client); err2 != nil {
-		return nil, err2
+func PrivateAuthenticatedPreset(c Config) Option {
```
gssbzn, line 278: Authenticated private API

internal/store/store.go
```diff
@@ -89,235 +106,268 @@ func customCATransport(ca []byte) http.RoundTripper {
...
-
-	return client, nil
 }
 
-// New get the appropriate client for the profile/service selected
-func New(c Config) (*Store, error) {
-	s := new(Store)
-	s.service = c.Service()
+// Option configures a Store.
+type Option func(s *Store) error
```
andreaangiolillo, line 139: [q] where is the implementation of this?

internal/store/store.go
```diff
@@ -89,235 +106,268 @@ func customCATransport(ca []byte) http.RoundTripper {
...
 	return t.Client()
 }
 
-func defaultClient(c Config) (*http.Client, error) {
-	client := http.DefaultClient
-	if caCertificate := c.OpsManagerCACertificate(); caCertificate != "" {
-		dat, err := ioutil.ReadFile(caCertificate)
+func (s *Store) transport() (*http.Transport, error) {
+	switch {
+	case s.caCertificate != "":
```
andreaangiolillo, line 125: [q] In which scenario `s.caCertificate` would be not empty? are we using `caCertificate` somewhere?

internal/store/store.go
```diff
@@ -89,235 +106,268 @@ func customCATransport(ca []byte) http.RoundTripper {
...
-	return s, err
-}
-
-func (s *Store) setAtlasClient(client *http.Client) error {
-	opts := make([]atlas.ClientOpt, 0)
-	if s.baseURL != "" {
-		opts = append(opts, atlas.SetBaseURL(s.baseURL))
+	switch store.service {
+	case config.CloudService:
+		err = withAtlasClient(client)(store)
```
andreaangiolillo, line 340: [nit] I personally don't like the name `withAtlasClient`, maybe `AtlasClientOption` would be better

internal/store/store.go
```diff
@@ -89,235 +106,268 @@ func customCATransport(ca []byte) http.RoundTripper {
...
-		s.caCertificate = caCertificate
-	}
-	if skipVerify := c.OpsManagerSkipVerify(); skipVerify != yes {
-		s.skipVerify = skipVerify
-	}
-	client, err := authenticatedClient(c)
-	if err != nil {
-		return nil, err
+// Options turns a list of Option instances into an Option.
+func Options(opts ...Option) Option {
```
andreaangiolillo, line 142: [q] Could we add a comment that explains what an Option represents?

internal/store/store.go
andreaangiolillo: oh I see, This is one of its implementation

```
func withAtlasClient(client *http.Client) Option {
	return func(s *Store) error {
		opts := []atlas.ClientOpt{atlas.SetUserAgent(userAgent)}
		if s.baseURL != "" {
			opts = append(opts, atlas.SetBaseURL(s.baseURL))
		}
		c, err := atlas.New(client, opts...)
		if err != nil {
			return err
		}
		s.client = c
		return nil
	}
}
```

andreaangiolillo: [nit] maybe moving this close to where we define `Option` would make it easier to find

internal/store/store.go
```diff
@@ -89,235 +106,268 @@ func customCATransport(ca []byte) http.RoundTripper {
...
+		return nil
 	}
+}
 
-	client, err := defaultClient(c)
-	if err != nil {
-		return nil, err
+func withAtlasClient(client *http.Client) Option {
+	return func(s *Store) error {
+		opts := []atlas.ClientOpt{atlas.SetUserAgent(userAgent)}
```
andreaangiolillo, line 215: [q] why we need `userAgent`?

internal/store/store.go
andreaangiolillo: I see why we are doing this but I also think that this way is much less intuitive.

andreaangiolillo (approved):
I have just some questions to understand completely these changes. LGTM once my questions are answered. Thanks for all these changes!

internal/store/store.go
gssbzn: they are as closue as they can be, you have `func Options(opts ...Option) Option ` just after this, parmas and return type
and then after `Options` you have `func Service(service string) Option` again return type

internal/store/store.go
gssbzn: be default is not empty, you have to set the cli to use the custom ca

internal/store/store.go
gssbzn: why do we need the variable or why do we set it?

the variable is there so we generate it once instead of every time we create a client, also it's shared with the ops manager client

internal/store/store.go
gssbzn: why not? the reason behind the `with` is that you want a `Store` `with` an atlas client smae for all the `with` around

internal/store/store.go
gssbzn: I can agree, it may take a while to get use to but this makes the `New` method now more maintainable and testable as we can now test the options without depending on a `Config`

This refactor is nothing more than a [dependency inversion](https://en.wikipedia.org/wiki/Dependency_inversion_principle) `New` no longer depends on `Config` which since the beginning was a bad call from my side, and I knew it but it kept things simple but as we added more `New` methods that choice started to show why it was a bad option, since `New` no longer depends on `Config` it's what makes it easier to test and maintain, we don't need `PrivateAuthenticatedPreset` I added the presets to keep usage simple but they can go any time and `New` will still work

internal/store/store.go
andreaangiolillo: sorry If I wasn’t clear, I am not sure what the userAgent is used for. Why do we need to set a userAgent? and what the userAgent does?

gssbzn:
@andreaangiolillo RFAL

internal/store/store.go
gssbzn: So Atlas knows who/what made the HTTP call, https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/User-Agent

andreaangiolillo (approved):
LGTM! Thanks for all the clarifications!

PR #675, opened 2021-05-04
feat: improve go fmt calls

gssbzn:
## Proposed changes

Only format changed go on `make fmt`

## Checklist

## Further comments

The `make fmt` command has become extremely slow and this has been bothering me for a while, the `find` function we used to use it's the bottle neck, so changing it to behave in a similar way as we handle the pre-commit and only format changed files

Also drive by fix to the pre commit to not format mocks and re add changed files


andreaangiolillo (approved):
LGTM

PR #678, opened 2021-05-17
Add support for Atlas LDAP UserToDNMapping flags

ehershey:
## Proposed changes

This adds three new flags in support of this feedback idea I submitted:
https://feedback.mongodb.com/forums/930808-mongodb-cli/suggestions/43443876-support-setting-usertodnmapping-in-ldap-config

_Jira ticket:_ None

## Checklist

- ~~[] I have updated [e2e/E2E-TESTS.md](e2e/E2E-TESTS.md) (if a new command or e2e test has been added)~~

## Further comments

I didn't see an obvious way to add validation that only one of the substitution or query flags are included but the backend returns this error if both or neither are submitted. I'm open to ideas for how/where to add validation if it would be appropriate.
```
Error: PATCH https://cloud.mongodb.com/api/atlas/v1.0/groups/xxxxx/userSecurity: 400 (request "INVALID_ATTRIBUTE") Invalid attribute userToDNMapping specified.
``` 

I tested this against a few live clusters where I needed to set their full configuration and wanted to avoid copying and pasting in the GUI.


internal/cli/atlas/security/ldap/save.go
```diff
@@ -96,6 +100,9 @@ func SaveBuilder() *cobra.Command {
 	cmd.Flags().StringVar(&opts.bindPassword, flag.BindPassword, "", usage.BindPassword)
 	cmd.Flags().StringVar(&opts.caCertificate, flag.CaCertificate, "", usage.CaCertificate)
 	cmd.Flags().StringVar(&opts.authzQueryTemplate, flag.AuthzQueryTemplate, "", usage.AuthzQueryTemplate)
+	cmd.Flags().StringVar(&opts.mappingMatch, flag.MappingMatch, "", usage.MappingMatch)
+	cmd.Flags().StringVar(&opts.mappingLdapQuery, flag.MappingLdapQuery, "", usage.MappingLdapQuery)
+	cmd.Flags().StringVar(&opts.mappingSubstitution, flag.MappingSubstitution, "", usage.MappingSubstitution)
```
andreaangiolillo, line 105: Please add these flags to this test https://github.com/ehershey/mongocli/blob/572f9736c1b462595b9dea25b28347f59d9ac975/internal/cli/atlas/security/ldap/save_test.go#L57

andreaangiolillo: We also have the LDAP e2e test that has to be updated https://github.com/ehershey/mongocli/blob/572f9736c1b462595b9dea25b28347f59d9ac975/e2e/atlas/ldap_test.go#L120

andreaangiolillo: This [comment](https://github.com/ehershey/mongocli/blob/572f9736c1b462595b9dea25b28347f59d9ac975/internal/cli/atlas/security/ldap/save.go#L79) should be updated

internal/cli/atlas/security/ldap/save.go
```diff
@@ -64,6 +67,7 @@ func (opts *SaveOpts) newLDAPConfiguration() *atlas.LDAPConfiguration {
 			AuthorizationEnabled:  opts.authorizationEnabled,
 			Hostname:              opts.hostname,
 			Port:                  opts.port,
+			UserToDNMapping:       []*atlas.UserToDNMapping{{Match: opts.mappingMatch, LDAPQuery: opts.mappingLdapQuery, Substitution: opts.mappingSubstitution}},
```
andreaangiolillo, line 70: Are you sure that we are not setting `UserToDNMapping` when `opts.mappingMatch`, `opts.mappingLdapQuery` and `opts.mappingSubstitution` are empty?

Our LDAP e2e test is broken https://evergreen.mongodb.com/task/mongocli_master_e2e_atlas_clusters_atlas_ldap_e2e_patch_3dc498b12ecd12496d2dd71d8e152a5f411680da_60a2ad5d1e2d170ad5a81575_21_05_17_17_52_30

andreaangiolillo (changes requested):
Thanks for opening this PR and contributing to MongoDB CLI! 🎉  💯 

I have some comments that I think should be addressed.

Regarding the validation, I think you can add something in the PreRunE as we are doing [here](https://github.com/ehershey/mongocli/blob/572f9736c1b462595b9dea25b28347f59d9ac975/internal/cli/atlas/clusters/create.go#L202). Another way would be to create a new function and then call it inside `opts.PreRunE`

```
return opts.PreRunE(
                                 opts.ValidateFlags,
				opts.ValidateProjectID,
				opts.initStore,
				opts.InitOutput(cmd.OutOrStdout(), saveTemplate) 
```

andreaangiolillo (changes requested):
Thanks for opening this PR and contributing to MongoDB CLI! 🎉  💯 

I have some comments that I think should be addressed.

Regarding the validation, I think you can add something in the PreRunE as we are doing [here](https://github.com/ehershey/mongocli/blob/572f9736c1b462595b9dea25b28347f59d9ac975/internal/cli/atlas/clusters/create.go#L202). Another way would be to create a new function and then call it inside `opts.PreRunE`

```
return opts.PreRunE(
                                 opts.ValidateFlags,
				opts.ValidateProjectID,
				opts.initStore,
				opts.InitOutput(cmd.OutOrStdout(), saveTemplate) 
```

internal/cli/atlas/security/ldap/save.go
gssbzn: In particular this error, https://evergreen.mongodb.com/test_log/mongocli_master_e2e_atlas_clusters_atlas_ldap_e2e_patch_3dc498b12ecd12496d2dd71d8e152a5f411680da_60a2ad5d1e2d170ad5a81575_21_05_17_17_52_30/0/github.com_mongodb_mongocli_e2e_atlas.TestLDAP_Save

gssbzn:
Hi @ehershey I would like to do a mongocli release this week an include this improvement let me know if you need any help here to get tests green again

Also about this 

> I didn't see an obvious way to add validation that only one of the substitution or query flags are included

If only one of 
```go
mappingMatch          string
mappingLdapQuery      string
mappingSubstitution   string
```

can be set it would be nice to validate at the client side,

you can define something like 

```go
func (opts *SaveOpts) validate() error {
	// a && (b || c) || (b && c): check if at least two are true
	if opts.mappingMatch != "" && (opts.mappingLdapQuery != "" || opts.mappingSubstitution  != "") || (opts.mappingLdapQuery != "" && opts.mappingSubstitution  != "") {
		return errors.New("can't supply more than one userToDNMapping")
	}
}
``` 

and add it to the list of validations that run in the command

```go
	return opts.PreRunE(
	opts.ValidateProjectID,
	opts.initStore,
	opts.validate,
	opts.InitOutput(cmd.OutOrStdout(), saveTemplate))
},
```

internal/cli/atlas/security/ldap/save.go
gssbzn, line 70: `UserToDNMapping` is an array, you can only define one value here, do you think other users would agree with this?

This may also be the problem with our test you are now passing an empty element in the array when no flag is set so adding to the array may need to be conditioned

Also since is an array I wonder if it may then make sense to let users pass multiple values

ehershey:
Okay! I’ll try to circle back around to this today.

internal/cli/atlas/security/ldap/save.go
ehershey: @gssbzn that's a good point. Based on [the docs|https://docs.mongodb.com/manual/reference/configuration-options/#mongodb-setting-security.ldap.userToDNMapping], having multiple values seems like a reasonably well supported scenario. I've never used multiple values. I do think other users may need them. 

I think it would make sense to move this out of the "atlas security ldap save" command and into a separate new "atlas security ldap addMapping" command (or just "add" to match the "atlas security ldap delete" command? Or even better, replace "delete" with "deleteMapping" that takes an argument of the match string of the entire entry to delete).

I can't work on that in the near term though. I think supporting one mapping is better than none.

internal/cli/atlas/security/ldap/save.go
ehershey: Done, done and done. For the e2e test do you think we should add anything fancier like different combination of new parameters? Or a save call with all of them empty?

internal/cli/atlas/security/ldap/save.go
ehershey: Ah, thanks! I believe I fixed this. I tested different combinations with the fix:
1) Passing no new flags (works - didn't work before)
2) Passing all new flags (local validation fails)
3) MappingMatch with neither query nor substitution (local validation fails)
4) MappingMatch with Query (validates fine and saves)
5) MappingMatch with Substitution (validates fine and saves)

ehershey:
Thanks a lot for the reviews and tips @andreaangiolillo and @gssbzn! I believe I addressed all comments in the PR.

gssbzn (approved):
Changes look good, I'm going to wait for the ldap tests to pass before merging this 
Thanks for your contributions @ehershey 🚀

internal/cli/atlas/security/ldap/save.go
andreaangiolillo: I think the more tests we have the better but I am not going to block this PR for this

andreaangiolillo (approved):
LGTM, thanks for addressing all my comments

PR #682, opened 2021-05-20
Custom docs template cleanup

melissamahoney-mongodb:
This PR updates the custom docs template to resolve the following errors in the docs [build](https://workerpool-boxgs.mongodbstitch.com/pages/job.html?jobId=60a5690af1f6937450ed07f5):
```
WARNING(reference/mongocli-atlas-privateEndpoints-azure.txt:49ish): Title underline too short.
ERROR(reference/mongocli-config.txt:18ish): Unexpected indentation.
```
This PR:
- Makes the underline for "Related Commands" the full length of the heading
- Removes indentation in the long summary of the config command, which isn't valid in RST


internal/cli/config/config.go
```diff
@@ -248,7 +248,7 @@ func Builder() *cobra.Command {
 		Short: "Configure a profile to store access settings for your MongoDB deployment.",
 		Long: `Configure settings in a user profile.
 All settings are optional. You can specify settings individually by running: 
-  $ mongocli config set --help
```
gssbzn, line 251: [q] any reason for this change? it's an unwritten rule on our examples to separate text from code to add some indentation, you'll find this pattern around all our descriptions so if this change is needed we may want to do it all over the place and in a separate PR

melissamahoney-mongodb: The reason is because this creates a build error in the generated RST. RST is extremely sensitive to indentation. Its reported about halfway down in the [build log](https://workerpool-boxgs.mongodbstitch.com/pages/job.html?jobId=60a5690af1f6937450ed07f5), and here in VSCode:
image

This is the only file that threw an error, but to be safe I searched through the rest of the repo and this is the only example I could find where indentation is used. I don't think we'd need to change this in other places, unless I'm looking in the wrong place?

gssbzn: no, it looks like is a thing with `Description` and this is the only description with a command instead of an `Example`

PR #688, opened 2021-05-24
feat: replace golint with revive

gssbzn:
## Proposed changes

`golint` has been deprecated and the recommended replacement is `revive`

## Checklist

### DIFF
--- a/internal/cli/atlas/clusters/create.go
+++ b/internal/cli/atlas/clusters/create.go
@@ -154,7 +154,7 @@ func (opts *CreateOpts) providerName() string {
 
 func (opts *CreateOpts) newReplicationSpec() atlas.ReplicationSpec {
 	var (
-		readOnlyNodes int64 = 0
+		readOnlyNodes int64
 		Priority      int64 = 7
 	)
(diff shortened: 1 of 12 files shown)


internal/cli/atlas/clusters/create.go
```diff
@@ -154,7 +154,7 @@ func (opts *CreateOpts) providerName() string {
 
 func (opts *CreateOpts) newReplicationSpec() atlas.ReplicationSpec {
 	var (
-		readOnlyNodes int64 = 0
+		readOnlyNodes int64
```
gssbzn, line 157: default is zero

threebee (approved):
LGTM!

internal/cli/atlas/clusters/create.go
threebee: ah yes, I review similar change last week -> `var slaveDelay float64 = 0`. I'll keep it in mind going forward.

gssbzn: this is just the new linter is picky about it, I would not mind

PR #689, opened 2021-05-24
[Quickstart] Added validation to password question

andreaangiolillo:
## Checklist

## Further comments

**Error:**
```
[Set up your database access details]
? Database User Username [Press Enter to use 'Quickstart-1621862864']: Quickstart-1621862864
? Database User Password [Press Enter to use an auto-generated password 'VVdJTGZdUklX']: password
Error: POST https://cloud-dev.mongodb.com/api/atlas/v1.0/groups/5e4e593f70dfbf1010295836/databaseUsers: 400 (request "COMMON_PASSWORD") The password provided is too weak, as it can be found in most commonly used password lists.
```

**Fix:**
```
[Set up your database access details]
? Database User Username [Press Enter to use 'Quickstart-1621866516']: Quickstart-1621866516
X Sorry, your reply was invalid: the password provided is too weak, as it can be found in most commonly used password lists: password
? Database User Password [Press Enter to use an auto-generated password 'RWV1RVlnVEto']: password!21212!!!

```


internal/validate/common_password.go
```diff
@@ -0,0 +1,413 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package validate
+
+var commonPasswords = map[string]bool{
```
andreaangiolillo, line 17: here I am using the same list used by the endpoint. I am not happy about this list given that we need to keep it updated with the one in the endpoint. I am happy to consider another way to validate this

internal/validate/validate.go
```diff
@@ -158,3 +158,18 @@ func DBUsername(val interface{}) error {
 
 	return fmt.Errorf("%w: %s", ErrInvalidDBUsername, name)
 }
+
+var ErrWeakPassoword = errors.New("the password provided is too weak, as it can be found in most commonly used password lists")
+
+func WeakPassword(val interface{}) error {
```
gssbzn, line 164: [q] tests?

internal/validate/validate.go
```diff
@@ -158,3 +158,18 @@ func DBUsername(val interface{}) error {
 
 	return fmt.Errorf("%w: %s", ErrInvalidDBUsername, name)
 }
+
+var ErrWeakPassoword = errors.New("the password provided is too weak, as it can be found in most commonly used password lists")
```
gssbzn, line 162: [nit] this is two long, for sentinel errors you want them short
I would make it (but feel free to improve on)
```go
var ErrWeakPassoword = errors.New("invalid password")

func WeakPassword(val interface{}) error {
...
if commonPasswords[password] {
	return fmt.Errorf("%w: password is too common", ErrWeakPassoword)
}
```
don't return the password when you asked it masked just to expose it in an error

internal/validate/common_password.go
threebee: Which endpoint are you talking about? Could you please point out the endpoint?

internal/validate/validate.go
```diff
@@ -158,3 +158,18 @@ func DBUsername(val interface{}) error {
...
+
+func WeakPassword(val interface{}) error {
+	password, ok := val.(string)
+	if !ok {
+		return fmt.Errorf("%w: %v", ErrWeakPassoword, val)
+	}
+
+	if commonPasswords[password] {
+		return fmt.Errorf("%w: %v", ErrWeakPassoword, val)
+	}
```
gssbzn, line 172: what about checking length? 8 is standard but now days is considered insecure (can be cracked in a couple hours) 10 may be a good value

internal/validate/common_password.go
threebee: Instead of looking up for weak password, could we add a password policy? Something like atleast 2 uppercase, 2 lowercase, 1 number and 1 special character?

internal/validate/common_password.go
```diff
@@ -0,0 +1,413 @@
...
+	"xavier":      true,
+	"xxxxxx":      true,
+	"xxxxxxxx":    true,
+	"yamaha":      true,
+	"yankee":      true,
+	"yankees":     true,
+	"yellow":      true,
+	"zxcvbn":      true,
+	"zxcvbnm":     true,
+	"zzzzzz":      true}
```
gssbzn, line 413:
```suggestion
	"zzzzzz":      true,
}
```

gssbzn (changes requested):
good add here but main thing before getting it merge, tests

internal/validate/common_password.go
gssbzn: could we make that a `private` endpoint?

gssbzn: same as we did with with current IP

gssbzn: > Instead of looking up for weak password, could we add a password policy?

Check for known common passwords is a good strategy, password policies are actually a bad practice now days, length and uniqueness (across services) is all that matters

internal/validate/common_password.go
```diff
@@ -0,0 +1,413 @@
...
+	"action":      true,
+	"albert":      true,
+	"alberto":     true,
+	"alexis":      true,
+	"alejandra":   true,
+	"alejandro":   true,
+	"amanda":      true,
+	"amateur":     true,
+	"america":     true,
+	"andrea":      true,
```
gssbzn, line 52: @andreaangiolillo 👀

internal/validate/common_password.go
andreaangiolillo: yes, I was thinking about the private endpoint but I wasn’t sure it was worth it

andreaangiolillo: @threebee this is the endpoint

`Error: POST https://cloud-dev.mongodb.com/api/atlas/v1.0/groups/5e4e593f70dfbf1010295836/databaseUsers: 400 (request "COMMON_PASSWORD") The password provided is too weak, as it can be found in most commonly used password lists.`

gssbzn: > yes, I was thinking about the private endpoint but I wasn’t sure it was worth it

I think it is but I would not block this PR on that change as we may want to discuss it further with other teams

internal/validate/common_password.go
andreaangiolillo: ahahaha I didn’t notice that xD

internal/validate/validate.go
andreaangiolillo: added!

internal/validate/validate.go
```diff
@@ -158,3 +160,23 @@ func DBUsername(val interface{}) error {
...
+
+func WeakPassword(val interface{}) error {
+	password, ok := val.(string)
+	if !ok {
+		return ErrWeakPassword
+	}
+
+	if commonPasswords[password] {
+		return ErrWeakPassword
+	}
```
gssbzn, line 175: something I just realised the map is all lowercase you may want to do this instead to be safe
```suggestion
	if commonPasswords[strings.ToLower(password)] {
		return ErrWeakPassword
	}
```

internal/validate/common_password.go
andreaangiolillo: I am going to create a new ticket to discuss this private endpoint once I merge this

internal/validate/validate.go
```diff
@@ -158,3 +160,23 @@ func DBUsername(val interface{}) error {
...
+
+	if len(password) < minPasswordLength {
```
gssbzn, line 177: check length before, some of those common password are shorter and may fail here first

internal/validate/validate.go
```diff
@@ -158,3 +160,23 @@ func DBUsername(val interface{}) error {
...
+		return ErrShortPassword
```
gssbzn, line 178: based on previous comment on formatting
```suggestion
		return fmt.Erorf("%w min: %d", ErrShortPassword, minPasswordLength)
```

gssbzn (commented):
almost there

gssbzn (approved):
LGTM :shipit:

PR #691, opened 2021-05-26
feat: use Atlas default major version API for quickstart

gssbzn:
## Proposed changes

Came to fix a typo but a couple of other small fixes showed up

## Checklist

## Further comments

- We have a function to get the current major version from atlas, using that instead of a hardcoded value
- From user feedback we know some users may want to set the tier for the cluster, we won't ask it but a flag is now available for it
- Fix a typo

### DIFF
--- a/internal/cli/atlas/quickstart/quick_start.go
+++ b/internal/cli/atlas/quickstart/quick_start.go
@@ -72,9 +72,9 @@ Loading sample data into your cluster... [It's safe to 'Ctrl + C']
 
 const (
 	replicaSet        = "REPLICASET"
-	mdbVersion        = "4.4"
 	shards            = 1
-	tier              = "M2"
+	atlasM2           = "M2"
+	atlasM5           = "M5"
 	tenant            = "TENANT"
 	members           = 3
 	zoneName          = "Zone 1"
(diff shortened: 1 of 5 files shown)


internal/cli/atlas/quickstart/quick_start.go
```diff
@@ -72,9 +72,9 @@ Loading sample data into your cluster... [It's safe to 'Ctrl + C']
 
 const (
 	replicaSet        = "REPLICASET"
-	mdbVersion        = "4.4"
 	shards            = 1
-	tier              = "M2"
+	atlasM2           = "M2"
+	atlasM5           = "M5"
```
andreaangiolillo, line 77: [q] why are we adding M5?

andreaangiolillo (approved):
LGTM

internal/cli/atlas/quickstart/quick_start.go
```diff
@@ -295,12 +297,26 @@ func (opts *Opts) newReplicationSpec() atlas.ReplicationSpec {
...
+		InstanceSizeName:    opts.tier,
+		ProviderName:        providerName,
 		RegionName:          opts.Region,
-		BackingProviderName: opts.Provider,
+		BackingProviderName: backingProviderName,
+	}
+}
+
+func (opts *Opts) providerName() string {
+	if opts.tier == atlasM2 || opts.tier == atlasM5 {
```
gssbzn, line 316: @andreaangiolillo this is why M5

PR #700, opened 2021-06-03
Update docs output folder to /command instead of /reference

melissamahoney-mongodb:
This PR renames the auto-generated docs output directory from "/reference" to "/command". On the Docs side, we decided that we should still have a separate "/reference" directory for any miscellaneous reference material that we might need to provide in the future.

This also means that the auto-generated command reference pages will have a nice semantic URL like `https://docs.mongodb.com/mongocli/v1.16/command/mongocli-atlas-clusters-create/`

[Docs Staged Example](https://docs-mongodbcom-staging.corp.mongodb.com/mongocli/docsworker-xlarge/autogen-ref-docs/command/mongocli/)

@gssbzn I did my best but please let me know if I missed anything :)

### DIFF
--- a/build/ci/evergreen.yml
+++ b/build/ci/evergreen.yml
@@ -993,16 +993,16 @@ tasks:
       - func: "generate docs"
       - command: archive.targz_pack
         params:
-          target: src/github.com/mongodb/mongocli/docs/reference.tgz
+          target: src/github.com/mongodb/mongocli/docs/command.tgz
           source_dir: src/github.com/mongodb/mongocli/docs
           include:
-            - "reference/*.txt"
+            - "command/*.txt"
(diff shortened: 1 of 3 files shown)


gssbzn:
If the files from https://spruce.mongodb.com/task/mongocli_master_generate_docs_generate_docs_patch_5f68b36b415ed5ea293713e612e65998ff5f1a62_60b902c0d1fe0769a09b97ac_21_06_03_16_26_41/logs?execution=0 LGTY then it LGTM

gssbzn (approved):
You need to merge master before merging but if it's a clean merge (and it should) you don't need another review

gssbzn (approved):

PR #711, opened 2021-06-21
tests: stricter fails for data lake e2e

gssbzn:
## Proposed changes

Data lake api had a bug on creation and the cli would still try to run all other test which would fail anyway adding extra noise

The bug on the API has been fixed but these changes are still worth keeping

## Checklist


e2e/atlas/data_lake_test.go
```diff
@@ -30,7 +30,8 @@ import (
 
 func TestDataLakes(t *testing.T) {
 	cliPath, err := e2e.Bin()
-	require.NoError(t, err)
+	r := require.New(t)
+	r.NoError(err)
```
threebee, line 34: [q] Why this is better than `equire.NoError(t, err)`?

threebee (approved):
LGTM, I have added a question for my own learning.

e2e/atlas/data_lake_test.go
gssbzn: it's a recommendation the lib gives to people  when using `assert.Something` or `require.Something` multiple times, I haven't gone too deep into the advantages but one will cache the `t` helper so it may reduce some memory usage

PR #712, opened 2021-06-21
task: update GOLANGCI_VERSION to v1.41.1

gssbzn:
## Proposed changes

Update our linter version

## Checklist

## Further comments

Looks like magic number got stricter


colm-quinn (approved):
LGTM, but also a general question. There's a lot of setting default values here, should we think about putting these in one place?

I think the answer is no as each file is meant to be authoritative for it's settings in go (can't find a reference to prove/disprove) but just want to check.

gssbzn:
there's two values, one in the github action and one in our makefile, I guess if we manage to start running our github actions locally we could get rid of the makefile value one

there's this but I have never had the chance to test it here, https://github.com/nektos/act

Feel free to open a ticket for it

PR #764, opened 2021-07-14
Added doc annotation dbusers

andreaangiolillo:
## Checklist

### DIFF
--- a/internal/cli/atlas/dbusers/certs/list.go
+++ b/internal/cli/atlas/dbusers/certs/list.go
@@ -61,7 +61,7 @@ func ListBuilder() *cobra.Command {
 		Args:    require.ExactArgs(1),
 		Annotations: map[string]string{
 			"args":         "username",
-			"usernameDesc": "Username of the database user for which you require certificates.",
+			"usernameDesc": "Username of the database user to get certificates for.",
 		},
 		PreRunE: func(cmd *cobra.Command, args []string) error {
 			opts.username = args[0]
--- a/internal/cli/atlas/dbusers/certs/list.go
+++ b/internal/cli/atlas/dbusers/certs/list.go
@@ -61,7 +61,7 @@ func ListBuilder() *cobra.Command {
 		Args:    require.ExactArgs(1),
 		Annotations: map[string]string{
 			"args":         "username",
-			"usernameDesc": "Username of the database user to get certificates for.",
+			"usernameDesc": "Username of the database user for which you require certificates.",
 		},
 		PreRunE: func(cmd *cobra.Command, args []string) error {
 			opts.username = args[0]
--- a/internal/cli/atlas/dbusers/certs/list.go
+++ b/internal/cli/atlas/dbusers/certs/list.go
@@ -61,7 +61,7 @@ func ListBuilder() *cobra.Command {
 		Args:    require.ExactArgs(1),
 		Annotations: map[string]string{
 			"args":         "username",
-			"usernameDesc": "Username of the database user to get certificates for.",
+			"usernameDesc": "Username of the database user for whom you want to list Atlas-managed certificates.",
 		},
 		PreRunE: func(cmd *cobra.Command, args []string) error {
 			opts.username = args[0]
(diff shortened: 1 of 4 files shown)


internal/cli/atlas/dbusers/certs/list.go
```diff
@@ -59,6 +59,10 @@ func ListBuilder() *cobra.Command {
 		Aliases: []string{"ls"},
 		Short:   "List of all Atlas-managed, unexpired certificates for a database user.",
 		Args:    require.ExactArgs(1),
+		Annotations: map[string]string{
+			"args":         "username",
+			"usernameDesc": "Username of the database user to get certificates for.",
```
colm-quinn, line 64: [nit] Ending in a for doesn't read well here.

```suggestion
			"usernameDesc": "Username of the database user for which you require certificates.",
```

internal/cli/atlas/dbusers/delete.go
```diff
@@ -52,6 +52,10 @@ func DeleteBuilder() *cobra.Command {
 		Aliases: []string{"rm"},
 		Short:   "Delete a database user for your project.",
 		Args:    require.ExactArgs(1),
+		Annotations: map[string]string{
+			"args":         "username",
+			"usernameDesc": "Username that this resource deletes from the MongoDB database.",
```
colm-quinn, line 57: [q] is `that this resource` how we refer to the mongocli calls usually? I'd be more inclined to say:
```suggestion
			"usernameDesc": "Username to be deleted from the MongoDB database.",
```

Same with the other cases, unless resource is a standard way we phrase usually.

colm-quinn (commented):
Small nits, let me know what you want to do with those and we can LGTM or iterate.

internal/cli/atlas/dbusers/delete.go
andreaangiolillo: I have used the same description that we have in the documentation of the endpoint for all these parameters
See https://docs.atlas.mongodb.com/reference/api/database-users-delete-a-user/
Screenshot 2021-07-14 at 12 22 52

colm-quinn: sgtm, thanks

colm-quinn (approved):
LGTM

internal/cli/atlas/dbusers/delete.go
melissamahoney-mongodb: I know that's how we say it in the API, and I know that this is a wrapper around the API, but I think Colm's right here: I wouldn't call this a "resource". It might be better to reference the old handwritten reference docs when looking for copy (if the handwritten doc exists). You can find it in [version 16](https://docs.mongodb.com/mongocli/v1.16/reference/atlas/dbuser-delete/) of the mongocli docs -- thought admittedly those aren't always great either. But that's why I'm here!

```suggestion
			"usernameDesc": "Username to delete from the MongoDB database.",
```

internal/cli/atlas/dbusers/describe.go
```diff
@@ -60,6 +60,10 @@ func DescribeBuilder() *cobra.Command {
 		Short:   "Return a single Atlas database user for your project.",
 		Args:    require.ExactArgs(1),
 		Aliases: []string{"get"},
+		Annotations: map[string]string{
+			"args":         "username",
+			"usernameDesc": "Username that this resource retrieves from the MongoDB database.",
```
melissamahoney-mongodb, line 65:
```suggestion
			"usernameDesc": "Username to retrieve from the MongoDB database.",
```

internal/cli/atlas/dbusers/update.go
```diff
@@ -81,6 +81,10 @@ func UpdateBuilder() *cobra.Command {
   Update scopes for a user
   $ mongocli atlas dbuser update <username> --scope resourceName:resourceType --projectId <projectId>`,
 		Args: require.ExactArgs(1),
+		Annotations: map[string]string{
+			"args":         "username",
+			"usernameDesc": "Username that this resource updates in the MongoDB database.",
```
melissamahoney-mongodb, line 86:
```suggestion
			"usernameDesc": "Username to update in the MongoDB database.",
```

melissamahoney-mongodb (changes requested):

melissamahoney-mongodb (approved):

PR #766, opened 2021-07-15
CLOUDP-95318: Add cluster e2e test for FedRamp

andreaangiolillo:
## Proposed changes

_Jira ticket:_ CLOUDP-95318

Evergreen Task: 

- [atlas_gov_clusters_e2e [E2E Atlas Cluster Tests]](https://evergreen.mongodb.com/task/mongocli_master_e2e_atlas_clusters_atlas_gov_clusters_e2e_patch_b877c01f2fbef088a8590b444b9a4c63490f6fc4_60f000a5d1fe0720a2bc043d_21_07_15_09_32_22)
- [ atlas_clusters_e2e [E2E Atlas Cluster Tests]](https://evergreen.mongodb.com/task/mongocli_master_e2e_atlas_clusters_atlas_clusters_e2e_patch_b877c01f2fbef088a8590b444b9a4c63490f6fc4_60f000a5d1fe0720a2bc043d_21_07_15_09_32_22)

## Checklist


Makefile
```diff
@@ -18,7 +18,7 @@ UNIT_TAGS?=unit
 INTEGRATION_TAGS?=integration
 E2E_TAGS?=e2e
 E2E_BINARY?=../../bin/${BINARY_NAME}
-E2E_TIMEOUT?=20m
+E2E_TIMEOUT?=30m
```
andreaangiolillo, line 21: This is now matching the timeout that we use for the evergreen tasks

gssbzn: you mean evergreen as a whole? consider that you want the test to fail before evergreen shuts it down so you can get test results

e2e/atlas/clusters_test.go
```diff
@@ -36,18 +42,24 @@ func TestClustersFlags(t *testing.T) {
...
 		cmd := exec.Command(cliPath,
 			atlasEntity,
 			clustersEntity,
 			"create",
 			clusterName,
-			"--region=US_EAST_1",
+			"--region="+region,
 			"--members=3",
-			"--tier=M10",
+			"--tier=M30",
```
colm-quinn, line 59: [q] Is this (and the disk space increase) needed for all tests or just Gov?

e2e/atlas/clusters_test.go
```diff
@@ -23,11 +23,17 @@ import (
...
+	"github.com/mongodb/mongocli/internal/config"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 	"go.mongodb.org/atlas/mongodbatlas"
 )
 
+const (
+	usEast1    = "US_EAST_1"
+	usGovEast1 = "US_GOV_EAST_1"
+)
```
gssbzn, line 35: can we use, ListCloudProviderRegions instead to avoid hardcoding this?

e2e/atlas/clusters_test.go
```diff
@@ -36,18 +42,24 @@ func TestClustersFlags(t *testing.T) {
 	clusterName, err := RandClusterName()
 	req.NoError(err)
 
+	region := usEast1
+	service := os.Getenv("MCLI_SERVICE")
+	if service == config.CloudGovService {
+		region = usGovEast1
+	}
```
gssbzn, line 49: see my previous comment

e2e/atlas/clusters_test.go
```diff
@@ -174,7 +186,7 @@ func TestClustersFlags(t *testing.T) {
 			clustersEntity,
 			"update",
 			clusterName,
-			"--diskSizeGB=20",
+			"--diskSizeGB=40",
```
colm-quinn, line 189: [nit] I'd move these sizes to defines like `DefaultSize` `LargeSize` etc so we can reuse easier. Not blocking.

colm-quinn (dismissed):
LGTM, some questions and nits, but none are blocking.

e2e/atlas/clusters_test.go
gssbzn: to follow colm's suggestion, see https://pkg.go.dev/go.mongodb.org/atlas@v0.10.1/mongodbatlas#pkg-variables

e2e/atlas/clusters_test.go
```diff
@@ -208,13 +220,19 @@ func TestClustersFile(t *testing.T) {
 	clusterFileName, err := RandClusterName()
 	req.NoError(err)
 
+	clusterFile := "create_cluster_test.json"
+	service := os.Getenv("MCLI_SERVICE")
+	if service == config.CloudGovService {
```
gssbzn, line 225:
```suggestion
	if service := os.Getenv("MCLI_SERVICE"); service == config.CloudGovService {
```

Makefile
andreaangiolillo: Yes, [evergreen.yml](https://github.com/mongodb/mongocli/blob/f238209efda0fd98b5bad6535da6746db0173e8a/build/ci/evergreen.yml#L15). I am going to decrease this to 25.

gssbzn (commented):
I'd like to make this a bit more predictable rather than having `if`s in the middle of the test, either using some of the endpoinsts we have added to know available regions per provider or the cheap solution, I rather have two files, clusters atlas and cluster atlas gov than those ifs they can creep rather quickly as the envs evolve

internal/cli/atlas/clusters/availableregions/available_regions.go
```diff
@@ -0,0 +1,33 @@
...
+	"github.com/spf13/cobra"
+)
+
+func Builder() *cobra.Command {
+	const use = "availableRegions"
+	cmd := &cobra.Command{
+		Use:     use,
+		Aliases: cli.GenerateAliases(use),
+		Short:   "Manage available regions for your project.",
+		Hidden:  true,
```
andreaangiolillo, line 28: This is hidden

e2e/atlas/helper_test.go
```diff
@@ -128,6 +129,40 @@ func deployCluster() (string, error) {
 	return clusterName, nil
 }
 
+func newAvailableRegion(projectID, tier, provider string) (string, error) {
```
gssbzn, line 132: [q] why pass the project ID? shouldn't it use the env as well?

e2e/atlas/helper_test.go
```diff
@@ -128,6 +129,40 @@ func deployCluster() (string, error) {
...
+	err = json.Unmarshal(resp, &cloudProviders)
+	if err != nil {
+		return "", err
+	}
+
+	if len(cloudProviders.Results) == 0 || len(cloudProviders.Results[0].InstanceSizes) == 0 {
+		return "", errors.New("no regions available")
+	}
+
+	return cloudProviders.Results[0].InstanceSizes[0].AvailableRegions[0].Name, nil
```
gssbzn, line 163: could we `return InstanceSizes[0].Name, InstanceSizes[0].AvailableRegions[0].Name, nil`, like return both and avoid to hard code the instance size in the test?

e2e/atlas/helper_test.go
andreaangiolillo: yes, this is a leftover from my previous approach with the store

e2e/atlas/helper_test.go
andreaangiolillo: mh.. I don't understand why. The purpose of this function is to return the first region available, I do not see why it should also return the `InstanceSize` which is given as input with `tier`

andreaangiolillo:
@gssbzn ready for review

e2e/atlas/helper_test.go
gssbzn: trying to see if we can get default tier, so is the tier flag mandatory? this is not a blocker just like previous comments in this pr trying to reduce the amount of values that are hardcoded

gssbzn (approved):
LGTM thanks for reducing complexity on the tests, the less `if` the better

e2e/atlas/helper_test.go
andreaangiolillo: no, the tier is not mandatory. If we don't provide it, we get the regions for all instance sizes. Are you suggesting not passing the tier to this function and return `return InstanceSizes[0].Name, InstanceSizes[0].AvailableRegions[0].Name, nil`? (If yes) I believe that this strategy may work since all instance sizes have the same regions, but this may change in the future.

I am going to merge this PR and open a new one if there are follow-ups here.

gssbzn: > Are you suggesting not passing the tier to this function and return return InstanceSizes[0].Name, InstanceSizes[0].AvailableRegions[0].Name, nil?

Yes that's what I say, it be better in the long wrong to just have a this function instead of hardcoded values

andreaangiolillo: I will open a PR to address this comment.

PR #767, opened 2021-07-19
CLOUDP-93374: Define a rule that explains when we should add flags in mongocli

andreaangiolillo:
## Proposed changes

_Jira ticket:_ CLOUDP-93374

[CONTRIBUTING.md](https://github.com/mongodb/mongocli/blob/CLOUDP-93374/CONTRIBUTING.md)

## Checklist

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -104,8 +104,8 @@ Flags are a way to modify the command, also may be called "options". Flags alway
 
 - `--flagName value`: this type of flag passes the value to the command. Examples: `--projectId 5efda6aea3f2ed2e7dd6ce05`
 - `--booleanFlag`: this flag represents a boolean and it sets the related variable to true when the flag is used, false otherwise.  Example: `--force`
-- `--flagName value1,value2,..,valueN`: you will also find flags that accept a list of values. This type of flag can be very useful to represent data structures as `--role roleName1@db,roleName2@db` or `--privilege action@dbName.collection,action2@dbName.collection`. 
-As shown in the examples, the standard format used to represent data structures consists of splitting the first value with the second one by at sign `@` and the second value with the third one by a full stop `.`. 
+- `--flagName value1,value2,..,valueN`: you will also find flags that accept a list of values. This type of flag can be very useful to represent data structures as `--role roleName1@db,roleName2@db`, `--privilege action@dbName.collection,action2@dbName.collection` or `--key field:type`. 
+As shown in the examples, the standard format used to represent data structures consists of splitting the first value with the second one by at sign `@` or colon `:` and the second value with the third one by a full stop `.`. 
 We recommend using configuration files for complex data structures that require more than three values (see [mongocli atlas cluster create](https://github.com/mongodb/mongocli/blob/f2e6d661a3eb2cfcf9baab5f9e0b1c0f872b8c14/internal/cli/atlas/clusters/create.go#L235)).
 
 ### Third Party Dependencies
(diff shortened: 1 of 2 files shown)


CONTRIBUTING.md
```diff
@@ -97,6 +97,17 @@ For that reason, command arguments tend to match the path and query params of th
 with the last param being a required argument and the rest handled via flag options. 
 For commands that create or modify complex data structures, the use of configuration files is preferred over flag options.
 
+#### How to define flags:
+Flags are a way to modify the command, also may be called "options".  Flags always have a long version with two dashes (--state) but often also have a shortcut with one dash and one letter (-s).
```
gssbzn, line 101:
```suggestion
Flags are a way to modify the command, also may be called "options". Flags always have a long version with two dashes (--state) but may also have a shortcut with one dash and one letter (-s).
```

CONTRIBUTING.md
```diff
@@ -97,6 +97,17 @@ For that reason, command arguments tend to match the path and query params of th
...
+
+`mongocli` uses the following types of flags:
+
+- `--flagName value`: this type of flag passes the value to the command. Examples: `--projectId 5efda6aea3f2ed2e7dd6ce05`
+- `--booleanFlag`: this flag represents a boolean and it sets the related variable to true when the flag is used, false otherwise.  Example: `--force`
+- `--flagName value1,value2,..,valueN`: you will also find flags that accept a list of values. This type of flag can be very useful to represent data structures as `--role roleName1@db,roleName2@db` or `--privilege action@dbName.collection,action2@dbName.collection`.
```
gssbzn, line 107: [nit] we don't always use `@`, roles and privilege just happen to have a standard format with `@`
meaning this role/privilege at this collection/db, we usually use `:` to split structs as well

internal/cli/atlas/customdbroles/create.go
```diff
@@ -72,7 +72,7 @@ func (opts *CreateOpts) validate() error {
 	return nil
 }
 
-// mongocli atlas dbrole(s) create <roleName> --privilege action[@dbName.collection] --inheritedRole role@db.
+// mongocli atlas dbrole(s) create <roleName> --privilege action@dbName.collection --inheritedRole role@db.
```
gssbzn, line 75: the reason of `[]` is to denote it's optional

CONTRIBUTING.md
andreaangiolillo: Shouldn’t we decide on only one format and stick with it?

internal/cli/atlas/customdbroles/create.go
andreaangiolillo: ah okay, I thought It should be `[action@dbName.collection]` in that case but I see how the user can provide only `action`

CONTRIBUTING.md
gssbzn: > Shouldn’t we decide on only one format and stick with it?

Maybe? can you do an inventory and see what have we been doing? `@` for denoting "at", makes sense but not sure what else we have have out there and if it would. going for only `@` would also limit the usage to only things that have that kind of relationship

andreaangiolillo: We have only the following commands that use a similar structure:

- `atlas|om dbuser create/update --role roleName@dbName --scope resourceName:resourceType`
- `atlas dbrole(s) create/update --privilege action@dbName.collection`
- `om cluster index create --key field:type`

Personally, I am fine with either approach (`:` or `@`) for splitting the first and the second values but `@` required less changes

andreaangiolillo: @gssbzn

gssbzn: that `field:type` would be weird with an `@` like you say Ii'm ok to limit to those two and decide when it makes sense to use one or the other

gssbzn (approved):
LGTM but that splitting I think should be flexible to allow `@` and `:`

CONTRIBUTING.md
andreaangiolillo: updated

gssbzn (approved):
LGTM thanks for the improvements on the docs

PR #788, opened 2021-08-16
CLOUDP-93453: Document changes that include user facing copy should include docs as reviewers

fmenezes:
## Proposed changes

_Jira ticket:_ CLOUDP-93453

## Checklist

## Further comments

Keep docs/command updated on code base


andreaangiolillo (commented):
I think I missing something but I thought that this ticket was about adding the doc team as code owner of https://github.com/mongodb/mongocli/tree/master/internal/usage.

fmenezes:
> I think I am missing something but I thought that this ticket was about adding the doc team as code owner of https://github.com/mongodb/mongocli/tree/master/internal/usage.

Look at @gssbzn point in the comment section
> If we can move the docs repo to be part of the cli repo then we can automate this better having to run doc generation always and making the docs team owner of their folder

andreaangiolillo:
I have completely missed those two lines. Thanks!

.github/workflows/docs.yml
```diff
@@ -0,0 +1,21 @@
+---
```
gssbzn, line 1: [not blocking] evergreen would've probably been a better place for this

.githooks/pre-commit
```diff
@@ -22,3 +22,6 @@ for FILE in ${STAGED_EVG_FILES}
 do
     evergreen validate "${FILE}"
 done
+
+make gen-docs > /dev/null
+git add docs
```
gssbzn, line 27: [nit] I think it makes sense to only run this inside `if [[ -n "${STAGED_GO_FILES}" ]]; then`

.github/workflows/docs.yml
gssbzn: [side note and blocking] let's remove the evergreen task that generates and attaches docs to the task

gssbzn (commented):
Thanks for doing this, one not so small change, let's remove the evergreen task that used to do this

gssbzn:
@fmenezes point to @andreaangiolillo, let's also add docs as owner of that folder, that's all copy

fmenezes:
> let's also add docs as owner of that folder, that's all copy

@gssbzn isn't this enough https://github.com/mongodb/mongocli/pull/788/files#diff-3d36a1bf06148bc6ba1ce2ed3d19de32ea708d955fed212c0d27c536f0bd4da7R2 ?

gssbzn:
> @gssbzn isn't this enough https://github.com/mongodb/mongocli/pull/788/files#diff-3d36a1bf06148bc6ba1ce2ed3d19de32ea708d955fed212c0d27c536f0bd4da7R2 ?

Indirectly, yes, I guess we can leave it out for now, just and edge case of changing the go file and not having the pre commit hooks but I guess the gh action should cover that

build/ci/evergreen.yml
```diff
@@ -1767,15 +1719,6 @@ buildvariants:
       go_bin: "/opt/golang/go1.16/bin"
     tasks:
       - name: release
-  - name: release_generate_docs
```
gssbzn, line 1770: since this no longer runs you need to update this file, https://github.com/mongodb/mongocli/blob/master/build/package/send-slack-notification.sh and remove the reference to the generated docs

gssbzn (approved):
LGTM, thanks for all those changes

melissamahoney-mongodb (commented):
I'd like to approve this but I just want to make sure I understand what this PR does:
* Generates the docs and adds them to docs/command
* Adds docs-cloud-team as an automatic reviewer for PRs to docs/*
Is that the case?

fmenezes:
correct @melissamahoney-mongodb, it also removes the generation of `command.tgz` from evergreen, as a step to use the submodules in `docs-mongocli` instead

melissamahoney-mongodb (approved):
Thanks, LGTM!

PR #795, opened 2021-08-19
CLOUDP-98099: Add support for defaultRWconcern

gssbzn:
## Proposed changes

_Jira ticket:_ CLOUDP-98099

## Checklist


fmenezes (dismissed):
LGTM, left a question, if is not relevant merge away

internal/convert/process_config.go
```diff
@@ -257,6 +288,9 @@ func newMongosProcessConfig(p *opsmngr.Process) *ProcessConfig {
 	if p.Args26.Security != nil {
 		pc.Security = p.Args26.Security
 	}
+	if p.Args26.Security != nil {
+		pc.Security = p.Args26.Security
```
fmenezes, line 292: side note, Args26 is not the catchiest name is it? 😅

internal/convert/process_config.go
```diff
@@ -232,6 +248,21 @@ func newMongosProcessConfig(p *opsmngr.Process) *ProcessConfig {
 		Name:                        p.Name,
 		SetParameter:                p.Args26.SetParameter,
 	}
+	if p.DefaultRWConcern != nil {
```
fmenezes, line 251: [q] is this correct? should it be the opposite?

gssbzn: in what sense? this converts from the ops manager struct to the cli struct, and checks if ops manager has a value then set the cli values from it

internal/convert/process_config.go
gssbzn: no, and this looks like copy/paste, need to remove this block

fmenezes (dismissed):

fmenezes (approved):
Restamp

PR #797, opened 2021-08-20
task: clean up config/const

gssbzn:
## Proposed changes

consolidate into a single file

## Checklist


fmenezes (dismissed):
LGTM

internal/config/profile.go
```diff
@@ -141,6 +169,10 @@ func (p *Profile) Service() string {
 	return settings[service]
 }
 
+func IsCloud() bool {
```
fmenezes, line 172: [q] Where are you planning on using this func?

internal/config/profile.go
```diff
@@ -17,6 +17,7 @@ package config
 import (
```
fmenezes, line 17: since you are touching this file what do you think about replacing `configHome()` with `os.UserConfigDir()`?

internal/config/profile.go
gssbzn: While testing some cloud gov things I spotted a couple of places where we don't account for it and may be bundleing into ops manager/cloud manager, search for `config.Service() != cloud`

internal/config/profile.go
gssbzn: configHome uses `os.UserHomeDir()` but it has extra logic to account for `XDG_CONFIG_HOME` (https://specifications.freedesktop.org/basedir-spec/latest/ar01s03.html)

gssbzn: we can move it to this file as well if you want, at some point it made sense to keep them separate

gssbzn: I have moved the function here

fmenezes: I think I follow now, you want `os.UserHomeDir()` for anyone not in unix/posix

fmenezes: Because if you follow the code it does something very similar in here https://cs.opensource.google/go/go/+/master:src/os/file.go;l=444-457?q=userconfigdir&ss=go%2Fgo

gssbzn: also in case `XDG_CONFIG_HOME` is not set by some *nix not following the freedesktop spec

fmenezes (approved):
LGTM, all points are questions not blockers

internal/config/profile.go
gssbzn: oh sorry, you meant, [UserConfigDir](https://pkg.go.dev/os@go1.17#UserConfigDir) and I've been talking of [UserHomeDir](https://pkg.go.dev/os@go1.17#UserHomeDir)

We can't make the switch at this point, `UserConfigDir` was added in 1.13, we already had this logic in place (we started in 1.12) the current behaviour of `UserConfigDir` is different from ours so changing it would break macOS and Windows users

PR #801, opened 2021-08-23
rfc: use embed for templates

gssbzn:
## Proposed changes

RFC: start using `embed` for output templates instead of plain strings

### DIFF
--- a/internal/cli/alerts/list.tmpl
+++ b/internal/cli/alerts/list.tmpl
@@ -1,0 +1,2 @@
+{{- /*gotype: go.mongodb.org/atlas/mongodbatlas.AlertsResponse*/ -}}ID	TYPE	STATUS{{range .Results}}
+{{.ID}}	{{.EventTypeName}}	{{.Status}}{{end}}
--- a/internal/cli/alerts/list.go
+++ b/internal/cli/alerts/list.go
@@ -15,6 +15,8 @@
 package alerts
 
 import (
+	_ "embed"
+
 	"github.com/mongodb/mongocli/internal/cli"
 	"github.com/mongodb/mongocli/internal/cli/require"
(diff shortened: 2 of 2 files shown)


internal/cli/alerts/list.tmpl
```diff
@@ -0,0 +1,2 @@
+{{- /*gotype: go.mongodb.org/atlas/mongodbatlas.AlertsResponse*/ -}}ID	TYPE	STATUS{{range .Results}}
```
gssbzn, line 1: type hints add autocomplete to idea

fmenezes:
@gssbzn I think `embed` it's a great feature and it works well with templates, I think particularly with projects embedding HTML and large JSONs it makes perfect sense.

That being said this particular example which is a small template (2 lines) I think it might bit a bit much. If the template was a bit bigger (let's say 5 to 10 lines at least) that would make more sense to me.

I think there are some files in our tests that could benefit from this feature. But then embedding for testing might be unnecessary.

gssbzn:
@fmenezes I can agree is a simple example but at the same time is the most common in the code base. One of the things I like of extracting to a file is that we could potentially lint them and catch any potential copy/paste issue (I'm also a bit biased to the feature as we can then have "view" files which I like the idea of and was something I wanted since we introduced the "simplified" output)

Anyway this is to see how we all feel about maybe moving this way for "views" we don't have to make a decision soon

andreaangiolillo:
I like the idea of having "view" files but I think that creating a new file for just 2-3 lines doesn’t seem a great trade-off.

gssbzn:
I'm going to close this for now as this was more for showing, I still think we should do this, specially as things like this https://github.com/mongodb/mongocli/pull/825#discussion_r704308191 tend to happen, so this will make sense if we can find/build a linter for the templates

PR #810, opened 2021-08-30
gofmt: gofmt with go 1.17

gssbzn:
## Proposed changes

I accidentally hit format project in IDEA and well here we are

## Checklist

### DIFF
--- a/internal/cli/alerts/acknowledge_test.go
+++ b/internal/cli/alerts/acknowledge_test.go
@@ -12,6 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+//go:build unit
 // +build unit
 
 package alerts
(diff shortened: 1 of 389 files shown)


MihaiBojin:
@gssbzn there are multiple PRs in-flight.

To merge this, I suggest reaching out to the authors and having them apply the same formatting rules, so that they don't run into unnecessary merge issues.

fmenezes (approved):
LGTM, merge conflicts should be easy to fix (essentially this is adding one comment to every file)

gssbzn:
>To merge this, I suggest reaching out to the authors and having them apply the same formatting rules, so that they don't run into unnecessary merge issues.

Totally my bad and I should have added this to the description, go1.17 introduced `//go:build` as a newer way for `// +build unit` there should not be merge conflicts, go.1.16 will ignore it/not add it, go1.17 will translate old tags to the newer when formatting

PR #818, opened 2021-09-01
CLOUDP-82836: Validate OM version requests

fmenezes:
## Proposed changes

This is verifying the version of Ops Manager before making requests to API Access List.

_Jira ticket:_ CLOUDP-82836

## Checklist

## Further comments

⚠️ this change depends on https://github.com/mongodb/go-client-mongodb-atlas/pull/248 & https://github.com/mongodb/go-client-mongodb-ops-manager/pull/122 ⚠️


fmenezes:
Unfortunately I couldn't run iam tests on OM, I will open a separated ticket for it.

internal/cli/iam/organizations/apikeys/accesslists/access_lists.go
```diff
@@ -36,3 +43,68 @@ func Builder() *cobra.Command {
...
+	if err != nil {
+		return false, err
+	}
+	versionParts := strings.Split(version.Version, ".")
+
+	const maxVersionParts = 3
+
+	if len(versionParts) > maxVersionParts {
+		versionParts = versionParts[0:3] // ops manager versions are not semantic this is converting it to x.y.z
+	}
```
colm-quinn, line 64: [np]: If we were doing testing with an rc, I think this would break (new version would keep the `-4cX` - a trim end to remove rc might be good here.

I'm really not sure how likely this scenario is however.

internal/cli/iam/organizations/apikeys/accesslists/create.go
```diff
@@ -71,12 +71,26 @@ func (opts *CreateOpts) Run() error {
...
 	}
 
-	return opts.Print(r)
+	var result *atlas.AccessListAPIKeys
+	if useWhitelist {
+		r, e := opts.store.CreateOrganizationAPIKeyWhitelist(opts.ConfigOrgID(), opts.apyKey, fromAccessListAPIKeysReqToWhitelistAPIKeysReq(req))
+		if e != nil {
+			return e
+		}
+		result = fromWhitelistAPIKeysToAccessListAPIKeys(r)
```
colm-quinn, line 85: [nit] I don't know if we try and remove `if-else` structures in go, but would read cleaner to me as an 
```
if {
    .. do something
    ...return result
  }
  .. do something else
  ... return result
}
```

internal/cli/iam/organizations/apikeys/accesslists/create_test.go
```diff
@@ -45,15 +50,46 @@ func TestCreate_Run(t *testing.T) {
 		t.Fatalf("Run() unexpected error: %v", err)
 	}
 
-	mockStore.
-		EXPECT().
-		CreateOrganizationAPIKeyAccessList(createOpts.OrgID, createOpts.apyKey, r).
-		Return(expected, nil).
-		Times(1)
+	t.Run("OM 5.0", func(t *testing.T) {
```
colm-quinn, line 53: [np] I'd add the rc test as well, if we're doing it.

internal/cli/iam/organizations/apikeys/accesslists/access_lists.go
fmenezes: just tested and `semver.NewVersion("5.0.0-rc0")` parses correcly, no errors

internal/cli/iam/organizations/apikeys/accesslists/create.go
fmenezes: the reason is the way it is it's because we join the branch later to `return opts.Print(result)` I could duplicate the line if you think it would read better.

internal/cli/iam/organizations/apikeys/accesslists/access_lists.go
colm-quinn: @gssbzn Will remember this better, last release it was `5.0` as we made sure it was in the first rc, however after we decided we don't need to block the first rc on the version manifest.

If the semver is working and getting the magor version, I think we're good, as it will be correct for the final rc

internal/cli/iam/organizations/apikeys/accesslists/access_lists.go
```diff
@@ -15,8 +15,15 @@
...
 
 import (
+	"strings"
+
+	"github.com/Masterminds/semver/v3"
 	"github.com/mongodb/mongocli/internal/cli"
+	"github.com/mongodb/mongocli/internal/config"
+	"github.com/mongodb/mongocli/internal/store"
 	"github.com/spf13/cobra"
+
```
gssbzn, line 25:
```suggestion
```

internal/cli/iam/organizations/apikeys/accesslists/access_lists.go
```diff
@@ -36,3 +42,68 @@ func Builder() *cobra.Command {
 
 	return cmd
 }
+
+// shouldUseWhitelist returns true when it is service Ops Manager and all versions below 5.0
+// returns false when Atlas, Cloud Manager or Ops Manager 5+.
+func shouldUseWhitelist(s store.ServiceVersionGetter) (bool, error) {
```
gssbzn, line 48: [nit] can we reverse the name please? 

```suggestion
// shouldUseAccessList returns true when it is service Ops Manager and all versions below 5.0
// returns false when Atlas, Cloud Manager or Ops Manager 5+.
func shouldUseAccessList(s store.ServiceVersionGetter) (bool, error) {
```

internal/cli/iam/organizations/apikeys/accesslists/access_lists.go
```diff
@@ -36,3 +42,68 @@ func Builder() *cobra.Command {
...
+	version, err := s.GetServiceVersion()
+	if err != nil {
+		return false, err
+	}
+	versionParts := strings.Split(version.Version, ".")
+
+	const maxVersionParts = 3
+
+	if len(versionParts) > maxVersionParts {
+		versionParts = versionParts[0:3] // ops manager versions are not semantic this is converting it to x.y.z
```
gssbzn, line 62:
```suggestion
		versionParts = versionParts[0:maxVersionParts] // ops manager versions are not semantic this is converting it to x.y.z
```

internal/cli/iam/organizations/apikeys/accesslists/create.go
colm-quinn: Yeah that's how I would have written it, with the `return opts.Print(result)` in both cases. It's just a nut/style thing - your call if you prefer.

internal/cli/iam/organizations/apikeys/accesslists/access_lists.go
gssbzn: I'm not sure I follow @colm-quinn's question here, version manifest is for server versions, this is for ops manager version, we need to drop anything after the last dot(.) so @fmenezes is right here that this should work

internal/cli/iam/organizations/apikeys/accesslists/create.go
gssbzn: a mix of @colm-quinn's point and my own, `opts.Print` doesn't care about type do you need to make result typed and add to the `if` and do you need `fromWhitelistAPIKeysToAccessListAPIKeys`?

internal/cli/iam/organizations/apikeys/accesslists/create_test.go
```diff
@@ -45,15 +50,66 @@ func TestCreate_Run(t *testing.T) {
...
+			CreateOrganizationAPIKeyAccessList(createOpts.OrgID, createOpts.apyKey, r).
+			Return(expected, nil).
+			Times(1)
+
+		if err = createOpts.Run(); err != nil {
+			t.Fatalf("Run() unexpected error: %v", err)
+		}
+	})
+
+	t.Run("OM 4.4", func(t *testing.T) {
```
gssbzn, line 93: let's add atlas and cloud manager tests to be extra sure please

internal/store/api_keys_access_list.go
```diff
@@ -91,72 +101,43 @@ func (s *Store) OrganizationAPIKeyAccessLists(orgID, apiKeyID string, opts *atla
...
-	out := &atlas.AccessListAPIKeys{
-		TotalCount: in.TotalCount,
-		Links:      in.Links,
-	}
-
-	results := make([]*atlas.AccessListAPIKey, len(in.Results))
-	for i, element := range in.Results {
-		results[i] = fromWhitelistAPIKeyToAccessListAPIKey(element)
+// CreateOrganizationAPIKeyWhitelist encapsulates the logic to manage different cloud providers.
+func (s *Store) CreateOrganizationAPIKeyWhitelist(orgID, apiKeyID string, opts []*atlas.WhitelistAPIKeysReq) (*atlas.WhitelistAPIKeys, error) {
```
gssbzn, line 113: can we add a 
```suggestion
// CreateOrganizationAPIKeyWhitelist encapsulates the logic to manage different cloud providers.
//
// Deprecated: kept until 4.4 reaches EOL
func (s *Store) CreateOrganizationAPIKeyWhitelist(orgID, apiKeyID string, opts []*atlas.WhitelistAPIKeysReq) (*atlas.WhitelistAPIKeys, error) {
```

or something like it ?

internal/store/service_version.go
```diff
@@ -0,0 +1,44 @@
...
+
+	"github.com/mongodb/mongocli/internal/config"
+	atlas "go.mongodb.org/atlas/mongodbatlas"
+	"go.mongodb.org/ops-manager/opsmngr"
+)
+
+//go:generate mockgen -destination=../mocks/mock_service_version.go -package=mocks github.com/mongodb/mongocli/internal/store ServiceVersionGetter
+
+type ServiceVersionGetter interface {
+	GetServiceVersion() (*opsmngr.ServiceVersion, error)
```
gssbzn, line 29: 👀 
```suggestion
	ServiceVersion() (*opsmngr.ServiceVersion, error)
```

internal/cli/iam/organizations/apikeys/accesslists/access_lists.go
```diff
@@ -36,3 +42,68 @@ func Builder() *cobra.Command {
...
 }
+
+// shouldUseWhitelist returns true when it is service Ops Manager and all versions below 5.0
+// returns false when Atlas, Cloud Manager or Ops Manager 5+.
+func shouldUseWhitelist(s store.ServiceVersionGetter) (bool, error) {
+	if config.Service() != config.OpsManagerService {
+		return false, nil
+	}
+
+	version, err := s.GetServiceVersion()
```
gssbzn, line 53: how do you feel about having all this from `s.GetServiceVersion()` to `sv, err := semver.NewVersion(newVersion)` as a separate method?

internal/cli/iam/organizations/apikeys/accesslists/access_lists.go
colm-quinn: I might be misreading, the original split is: 
`versionParts := strings.Split(version.Version, ".")` ,giving `[5,0,0-rc0]` 

In that case, which means that the `-rc0` text isn't dropped by: `newVersion := strings.Join(versionParts, ".")`

gssbzn: yes, and that's valid, the problem is `5,0,0-rc0.something` .something is not valid and what we want to drop

colm-quinn: Ah got it, thanks!

fmenezes:
@gssbzn @colm-quinn this should be ready for another look

internal/cli/iam/organizations/apikeys/accesslists/create_test.go
```diff
@@ -29,67 +30,65 @@ import (
...
-			t.Fatalf("Run() unexpected error: %v", err)
-		}
-	})
+	runCreateTest(t, config.CloudService, "", true)
+	runCreateTest(t, config.CloudGovService, "", true)
+	runCreateTest(t, config.CloudManagerService, "", true)
+	runCreateTest(t, config.OpsManagerService, "5.0.0.100.20210101T0000Z", true)
+	runCreateTest(t, config.OpsManagerService, "5.0.0-rc0.100.20210101T0000Z", true)
+	runCreateTest(t, config.OpsManagerService, "4.4.0.100.20210101T0000Z", false)
+	runCreateTest(t, config.OpsManagerService, "4.2.0.100.20210101T0000Z", false)
```
colm-quinn, line 91: 💯

colm-quinn (approved):
LGTM from my side, thanks for the fixes and discussions!

PR #826, opened 2021-09-07
test: remove tests for deprecated command

gssbzn:
## Proposed changes

Remove tests for deprecated command, this commands have been deprecated for a while and they are starting to flake so better to remove them

## Checklist


andreaangiolillo (dismissed):
I don't agree with this decision as I think that if we provide this functionality to our customers we should test it (if possible). In my opinion, if we decide to remove the test we should also remote the command.
No Blocking, this is only my opinion.

colm-quinn (approved):
LGTM, also q, is there an entry that needs to be removed/changed in `E2e-TESTS.md` as a result of this?

gssbzn:
> is there an entry that needs to be removed/changed in E2e-TESTS.md as a result of this?

good call, I deleted the entries as we don't even document these commands any more, we should plan some time to start deleting some of the deprecated stuff

andreaangiolillo (approved):
LGTM

PR #839, opened 2021-09-10
CLOUDP-99850: mongocli run iam tests for Ops Manager

andreaangiolillo:
## Proposed changes

_Jira ticket:_ CLOUDP-99850

[ops_manager_4_4_iam_e2e](https://evergreen.mongodb.com/task/mongocli_master_e2e_ops_manager_44_ops_manager_4_4_iam_e2e_patch_e0481661d5a15c20ffc919b2bc90ea686bda8d28_613f1f1be3c3312a36b95b19_21_09_13_09_51_25)

[ops_manager_5_0_iam_e2e](https://evergreen.mongodb.com/task/mongocli_master_e2e_ops_manager_50_ops_manager_5_0_iam_e2e_patch_e0481661d5a15c20ffc919b2bc90ea686bda8d28_613f1f1be3c3312a36b95b19_21_09_13_09_51_25)

## Checklist


e2e/iam/atlas_users_test.go
```diff
@@ -11,6 +11,7 @@
 // WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 // See the License for the specific language governing permissions and
 // limitations under the License.
+//go:build e2e || iam
 // +build e2e iam
```
colm-quinn, line 15: [q] Is thee second line still needed here?

colm-quinn (approved):
LGTM if tests. are passing. Small question, not sure if it's relevant.

build/ci/evergreen.yml
```diff
@@ -1008,6 +1008,26 @@ tasks:
...
+      - func: "build"
+      - func: "deploy spawn host"
+      - func: ssh-ready
+      - func: "install ops manager"
+        vars:
+          ARCHIVE: ${ops_manager_4_4_archive}
+      - func: "set-up ops manager"
+      - func: "e2e test"
+        vars:
+          E2E_TAGS: iam_mms_44
```
gssbzn, line 1029: [q] do we need `iam_mms_44` and `iam_mms_50`? any reason to not use the existing `iam`

andreaangiolillo: some iam commands are not available on 4.4 or 5.0

andreaangiolillo: or we cannot test it with ops manager

gssbzn: it will also be interesting to try to rely on an existing test that deploys ops manager already as that's the part that takes longer on these tests and since they can't be flaky with concurrent tests no need to split them so much

e2e/iam/atlas_users_test.go
andreaangiolillo: no, it is not. I am going to open a follow-up PR to update the build tags to all our e2e tests, included the ones in this PR

build/ci/evergreen.yml
gssbzn: > some iam commands are not available on 4.4 or 5.0

can we make it then so that it's a combination `iam,om5` and the tests where is not available have `!om5` that way is more obvious they are not supported and not an overlook

andreaangiolillo: Done 
> can we make it then so that it's a combination iam,om5 and the tests where is not available have !om5 that way is more obvious they are not supported and not an overlook

I don't have a strong opinion, I believe is easy to see that we are running iam test against mms if we keep them separated. In addition, tests takes only ~12m
> it will also be interesting to try to rely on an existing test that deploys ops manager already as that's the part that takes longer on these tests and since they can't be flaky with concurrent tests no need to split them so much

gssbzn:
tests are broken https://spruce.mongodb.com/task/mongocli_master_e2e_generic_atlas_generic_e2e_patch_e0481661d5a15c20ffc919b2bc90ea686bda8d28_613f5915e3c3312a36b9a3f5_21_09_13_13_58_46/logs?execution=1&page=0&sortBy=STATUS&sortDir=DESC

build/ci/evergreen.yml
gssbzn: > tests takes only ~12m

the whole test takes 12m, where about 10 of those is installing ops manager, not a blocker right now but those minutes add up

gssbzn (approved):
LGTM thanks for those changes

andreaangiolillo:
evergreen retry

PR #854, opened 2021-09-14
fixed tests failures

andreaangiolillo:
## Proposed changes

Closes #[issue number]

## Checklist


andreaangiolillo:
Evergreen patch with all tests passing expect ldap: https://evergreen.mongodb.com/version/6140bc200ae6063208e792ff

e2e/iam/org_api_key_access_list_test.go
```diff
@@ -11,7 +11,7 @@
 // WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 // See the License for the specific language governing permissions and
 // limitations under the License.
-//go:build e2e || (iam && atlas) || (iam && om50) || (!om44 && !opsmanager)
+//go:build (e2e || iam || (iam && om50)) && !om44 && !opsmanager
```
gssbzn, line 14: could we drop `om50` given this should cover eventually om60?
```suggestion
//go:build (e2e || iam) && !om44 && !opsmanager
```

gssbzn: also what's the use of `!opsmanager` wouldn't that mean never for ops manager or something like it? that's how I see it

gssbzn (commented):
couple of questions that may apply to all test before LGTMing

e2e/iam/org_api_key_access_list_test.go
andreaangiolillo: Do you want to keep om44 and drop only om50? if yes, I was thinking to use opsmanager for 50. is this fine? are you proposing something different?

andreaangiolillo: `ops_manager_4_4_generic_e2e` use `opsmanager` as a tag so` !opsmanager` was just preventing running this test as part of that evergreen task. I believe I can remove it with this new build formula

andreaangiolillo: Updated

gssbzn: original comment is about going disallow list only rather than allow and disallow at the same time, if the test calls `iam` it runs all of iam, `iam,om44` runs all iam except for those with `!om44` indicating not supported, and so on so we don't need to revisit the tags when a new version of OM comes next year

The use of `opsmanager` is honest question, not sure how it's been used here

build/ci/evergreen.yml
```diff
@@ -1145,7 +1145,7 @@ tasks:
       - func: "set-up ops manager"
       - func: "e2e test"
         vars:
-          E2E_TAGS: iam,om50
```
gssbzn, line 1148: what would happen next year for ops manager 6.0? will ops manager 6.0 become `opsmanager`?

andreaangiolillo: yes, this was my idea. Then we would drop om44 and use om50

gssbzn: why not use om50 now and save some refactor later? I don't see why using `opsmanager` to indicate latest, is not intuitive

andreaangiolillo:
All green 🌲 🌲 🌲  https://evergreen.mongodb.com/version/6141bb393e8e8601b2248963

e2e/iam/org_api_key_access_list_test.go
andreaangiolillo: with the latest changes, I am using `opsmanger` as the latest version, in this case 5.0. In the future, `opsmanager` would represent 6.0

andreaangiolillo: with the latest changes, I am using `opsmanger` as the latest version, in this case 5.0. In the future, `opsmanager` would represent 6.0

build/ci/evergreen.yml
andreaangiolillo: I honestly removed it only because of this comment
> could we drop om50 given this should cover eventually om60?

I will update `opsmanager` with `om50`.

e2e/iam/org_api_key_access_list_test.go
gssbzn: what will happen when a feature is `om60` only, you'll need to add a `!om50` to that test meaning you now need to change the test for ops manger 5.0 to be `om50` instead of `opsmanager`, this is what I'm trying to prevent next year and why I'm saying go disallow list, by default all test run, unless not allowed to run

e2e/cloud_manager/agent_api_keys_test.go
```diff
@@ -11,7 +11,7 @@
 // WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 // See the License for the specific language governing permissions and
 // limitations under the License.
-//go:build e2e || (cloudmanager && remote && replica) || (opsmanager && remote && replica)
+//go:build e2e || (cloudmanager && remote && replica) || (om44 && remote && replica) || (om50 && remote && replica)
```
andreaangiolillo, line 14: I am not sure if we want something like this 
```suggestion
//go:build e2e ||  remote && replica
```
or we want to keep it explicit who is running this test

gssbzn (dismissed):
LGTM some small comments to simplify the bool logic but not blocking

gssbzn (dismissed):
LGTM

andreaangiolillo:
@gssbzn I need a restamp. I have updates some flags. Please have a look at `updated tags` (ops manager iam was failing) and `updated version manifest` (version manifest tests wasn’t run as part of the generic tag)

gssbzn (approved):

PR #865, opened 2021-09-16
CLOUDP-100613: When no profile file exists the version checker fails

fmenezes:
## Proposed changes

When no profile exists `~/.config/mongocli.toml` commands might fail (for instance `mongocli config ls`) this is fixing it by moving the code responsible to find the version outside of the store and also ignoring errors if it fails to print the update alert banner (we should not block a command from being successful even if stderr is somehow defective).

_Jira ticket:_ CLOUDP-100613

## Checklist


gssbzn (approved):
Lgtm

PR #949, opened 2022-01-26
Static code analysis with gokart

MihaiBojin:
## Proposed changes

Added a static analysis (security) check: https://github.com/praetorian-inc/gokart

## Checklist

- [ ] I have updated [e2e/E2E-TESTS.md](e2e/E2E-TESTS.md) (if an e2e test has been added)


.tool-versions
```diff
@@ -0,0 +1 @@
+golang 1.17
```
MihaiBojin, line 1: Fix the go version.  According to https://github.com/kennyp/asdf-golang#version-selection `asdf` should pick up the version from `go.mod`, but it didn't work for me.  If this is a `me` problem, please let me know and I'll update the docs / remove this file. Thanks!

Makefile
```diff
@@ -29,14 +29,19 @@ setupgolangcilint:  ## Install golangci-lint
 	@echo "==> Installing golangci-lint..."
 	curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s $(GOLANGCI_VERSION)
 
+.PHONY: setupgokart
+setupgokart:  ## Install gokart
```
MihaiBojin, line 33: install the tool

Makefile
```diff
@@ -60,13 +65,18 @@ lint: ## Run linter
...
+	gokart scan . -v -g
+
 .PHONY: fix-lint
 fix-lint: ## Fix linting errors
 	@echo "==> Fixing lint errors"
 	golangci-lint run --fix --timeout 5m
 
 .PHONY: check
-check: test fix-lint ## Run tests and linters
+check: scan test fix-lint ## Run tests and linters
```
MihaiBojin, line 79: run the scanner

.tool-versions
gssbzn: works for me, we used to have this file but because of that same doc you're linking we removed it, make youre your goenv is properly set tho

Makefile
gssbzn: let's please flow a red -> green -> refactor flow, meaning tests before linting

gssbzn:
couple of questions, how is this different from gosec https://github.com/mongodb/mongocli/blob/245f8531886aa4bd2035e25ef33f8e8c2c7f2e5b/.golangci.yml#L73

Could we replace gosec with gokart?

And finally should this run on our CIs?

.tool-versions
MihaiBojin: Without this file, asdf-golang doesn't seem to correctly pick up the necessary go version and instead relies on the shimed/system-installed go version.

```
> asdf install
golang 1.16 is already installed

> asdf list golang
  1.16
  1.17

> cat go.mod | head -3
module github.com/mongodb/mongocli

go 1.17
```

MihaiBojin:
I don't see a reason to remove `gosec` (more security is better / you never know what one might catch than the other might not).

gssbzn:
> I don't see a reason to remove gosec

Reading their [announcement](https://www.praetorian.com/blog/introducing-gokart/) they sell themselves as a "better" `gosec` so I think we can disable it to save some cpu cycles on linting but no strong opinion there

build/ci/evergreen.yml
```diff
@@ -234,6 +242,14 @@ functions:
       params:
         working_dir: src/github.com/mongodb/mongocli
         command: make setupgolangcilint
+  "install gokart":
+    - command: subprocess.exec
+      type: setup
+      params:
+        working_dir: src/github.com/mongodb/mongocli
```
gssbzn, line 249:
```suggestion
      params:
        add_to_path:
          - ${go_bin}
        working_dir: src/github.com/mongodb/mongocli
```

Makefile
```diff
@@ -29,14 +29,19 @@ setupgolangcilint:  ## Install golangci-lint
 	@echo "==> Installing golangci-lint..."
 	curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s $(GOLANGCI_VERSION)
 
+.PHONY: setupgokart
+setupgokart:  ## Install gokart
+	@echo "==> Installing gokart..."
+	go install github.com/praetorian-inc/gokart@latest
```
gssbzn, line 35: This PR risks becoming more complex than probably you planned to but, installing latest could break evg by updating the linter to a newer version than what we may be running locally, that said go has a "feature" where doing go install with a pined version will pollute the `go.mod` file, we used to have this issue and we hacked it around with some `pushd`/`popd`, we can leave latest for now until it becomes a problem

MihaiBojin: We seem to be doing it in other places too, e.g.: `check-licenses.sh`, which is why I went with `latest` and didn't fix the version in place.

gssbzn: check licenses is more deterministic, it checks the file includes a license header but like I mention for now we can live with latest, I may give a try adding gokart as an official lint to golangci lint

MihaiBojin: I found out that there isn't a way to ignore in `gokart`. That makes the tool less useful since we can't make it fail in CI (or locally).  One option would be to remove the use of `os.Environ()` in `mongosh.go`, but we may have a legitimate use for passing the env (now or in the future), making it difficult to work with.

MihaiBojin:
Since there hasn't been an update in go-kart which enables ignoring certain files/lines, I'm going to close this PR and we can revisit adding this tool in the future.

PR #951, opened 2022-01-26
task: regenerate mocks

gssbzn:
## Proposed changes

While generating a new mock a mock for live migration as also generated, this means we had a fake livemigration not generated by the tooling, regenerating all mocks just in case

## Checklist


colm-quinn (approved):
LGTM. Out of curiousity, what causes the incorrect mock to be generated in the first place (how can we avoid it happening again in the future?)

gssbzn:
> what causes the incorrect mock to be generated in the first place (how can we avoid it happening again in the future?)

We add something like this (actual culprit here) to interfaces we want to mock
```go
//go:generate mockgen -destination=../mocks/mock_live_migration_validations.go -package=mocks github.com/mongodb/mongocli/internal/store LiveMigrationValidationsCreator,LiveMigrationCutoverCreator,LiveMigrationValidationsDescriber
```

My guess, somebody changed the output file on that command and forgot to regenerate the mocks since no test will fail as long as the mocks are the same

to not happen again maybe add a CI check to generate mocks and fail if any `diff`

PR #959, opened 2022-01-31
CLOUDP-111260: Allow mongocli to use tokens for authentication

gssbzn:
## Proposed changes

_Jira ticket:_ CLOUDP-111260

## Checklist

## Further comments

```bash
./bin/mongocli auth login --profile dev

First, copy your one-time code: PW79-F9DX

Next, sign with your browser and enter the code.

Or go to https://account.mongodb.com/account/connect

Your code will expire after 10 minutes.
Successfully logged in.
Press Enter to continue your profile configuration
? Choose a default organization: 1 (5e7cdd07d59e2b4b392bcb9c)
? Choose a default project: 1 (61aa6fc08397d91def2ff012)
? Default Output Format:
? Default MongoDB Shell Path: /usr/local/bin/mongosh

To use this profile, you must set the flag [-P dev] for every command.
You can use [mongocli config set] to change these settings at a later time.
```


internal/cli/auth/login.go
```diff
@@ -139,7 +206,8 @@ func LoginBuilder() *cobra.Command {
 	cmd.Flags().BoolVar(&opts.isGov, "gov", false, "Log in to Atlas for Government.")
 	cmd.Flags().BoolVar(&opts.isCloudManager, "cm", false, "Log in to Cloud Manager.")
 	cmd.Flags().BoolVar(&opts.noBrowser, "noBrowser", false, "Don't try to open a browser session.")
-
+	cmd.Flags().BoolVar(&opts.noProfile, "loginOnly", false, "Skip profile configuration.")
```
blva, line 209: [nit] why not using similar names in `noProfile` and `loginOnly`?

internal/cli/auth/login.go
```diff
@@ -110,10 +117,67 @@ Your code will expire after %.0f minutes.
...
+	var projectID string
+	if err != nil || len(pSlice) == 0 {
+		return errors.New("no projects")
+	}
+
+	p := prompt.NewProjectSelect(pSlice)
+	if err := survey.AskOne(p, &projectID); err != nil {
+		return err
+	}
+	opts.ProjectID = pMap[projectID]
```
tibulca, line 180: most of the code here can be found also in [config command](https://github.com/mongodb/mongocli/blob/master/internal/cli/config/config.go)
For consistency (especially for the future, if we'll change any of these 2 commands) is it possible to share the common logic (I'd say config.go seems like a more suitable place), or we're trying to keep the commands as isolated as possible?

internal/cli/auth/login.go
```diff
@@ -110,10 +117,67 @@ Your code will expire after %.0f minutes.
...
+		return err
+	}
+
+	opts.SetUpProject()
+	opts.SetUpOrg()
+	if err := survey.Ask(opts.DefaultQuestions(), opts); err != nil {
+		return err
+	}
+	opts.SetUpOutput()
+	opts.SetUpMongoSHPath()
```
blva, line 142: could this be a function to make simplify `Run`?

internal/cli/auth/login.go
```diff
@@ -110,10 +117,67 @@ Your code will expire after %.0f minutes.
...
 	if err := opts.config.Save(); err != nil {
 		return err
 	}
-	_, _ = fmt.Fprintf(opts.OutWriter, "Successfully logged in\n")
+	_, _ = fmt.Fprintf(opts.OutWriter, "\nYour profile is now configured.\n")
```
blva, line 146: [q] I'm wondering if given that we are printing so many things here in this package, do we want to maybe create a function that simplifies this output with a void function like `print( "\nYour profile is now configured.\n")` 

and 
```
(opts *loginOpts) print(txt string) {
	_, _ = fmt.Fprintf(opts.OutWriter, txt)
}
```

or is this not very go-ish?

blva (commented):
added minor comments, checking tests

internal/cli/auth/login.go
gssbzn: not sure if not goish but I gave it a try based on your comment, I decided against it as linters and the IDE loose the ability to flag when you make a formatting error

I intentionally removed one of the params and I no longer get warnings for it

Screenshot 2022-02-02 at 12 14 23

also given the varadic nature of printf wrapping it adds a bit of memory overhead to pack and unpack the args

internal/cli/auth/login.go
gssbzn: having a hard time to come up with a name for this but we could move the oauth part to a function which probably archives the same goal

internal/cli/auth/login.go
gssbzn: there's a main difference is that for config we have a fallback case where we don't do a select but let users input the id directly, this happens if they don't provide API keys which is allowed and would not happen for login, the main part that's shared `opts.Projects()` is being extracted as par of this PR

internal/cli/auth/login.go
gssbzn: it started as `noProfile` but then `loginOnly` felt as a better name, renamed the prop now

tibulca (approved):

internal/cli/auth/login.go
blva: nice check! fair enough

blva (approved):
LGTM

PR #963, opened 2022-02-02
CLOUDP-106123: [mongocli ] Implement mongocli auth logout

gssbzn:
## Proposed changes

_Jira ticket:_ CLOUDP-106123

## Checklist

## Further comments

```bash
./bin/mongocli auth logout --profile test
? Are you sure you want to log out No

./bin/mongocli auth logout --profile test
? Are you sure you want to log out Yes
Successfully logged out

./bin/mongocli auth logout --profile test
Error: not logged in
```


internal/cli/delete_opts.go
```diff
@@ -54,6 +54,8 @@ func (opts *DeleteOpts) Delete(d interface{}, a ...string) error {
 
 	var err error
 	switch f := d.(type) {
+	case func() error:
```
gssbzn, line 57: adding support for a no "entry" case

internal/cli/auth/logout.go
```diff
@@ -0,0 +1,95 @@
...
+
+	return opts.Delete(opts.config.Delete)
+}
+
+func LogoutBuilder() *cobra.Command {
+	opts := &logoutOpts{
+		DeleteOpts: cli.NewDeleteOpts("Successfully logged out\n", " "),
+	}
+
+	cmd := &cobra.Command{
```
blva, line 69: Do we need to add annotations to commands with no args? I've seen something like `		Annotations: map[string]string{}, ` for such scenario.

internal/cli/auth/logout.go
```diff
@@ -0,0 +1,95 @@
...
+	if _, err := opts.flow.Revoke(ctx, config.RefreshToken(), "refresh_token"); err != nil {
+		return err
+	}
+
+	return opts.Delete(opts.config.Delete)
+}
+
+func LogoutBuilder() *cobra.Command {
+	opts := &logoutOpts{
+		DeleteOpts: cli.NewDeleteOpts("Successfully logged out\n", " "),
```
blva, line 66: is the space intentional here in `" "`?

blva (commented):
Minor comments

internal/cli/auth/logout.go
gssbzn: I don't think we should there's no benefit to adding an empty map

internal/cli/auth/logout.go
gssbzn: yes, if I don't there's a fallback message we don't want 

I was actually tempted to not use our deleter struct and just copy/paste what's needed so I'm happy to take that path

internal/cli/auth/logout.go
```diff
@@ -0,0 +1,95 @@
...
+		PreRunE: func(cmd *cobra.Command, args []string) error {
+			opts.OutWriter = cmd.OutOrStdout()
+			opts.config = config.Default()
+			return opts.initFlow()
+		},
+		RunE: func(cmd *cobra.Command, args []string) error {
+			if config.RefreshToken() == "" {
+				return errors.New("not logged in")
+			}
+			if err := opts.PromptWithMessage("Are you sure you want to log out"); err != nil {
```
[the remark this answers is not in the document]
gssbzn: added

blva (approved):

PR #968, opened 2022-02-04
Fix M0 e2e tests failure

andreaangiolillo:
## Proposed changes

[atlas_clusters_m0_e2e](https://evergreen.mongodb.com/task/mongocli_master_e2e_atlas_clusters_atlas_clusters_m0_e2e_4b4ad609c9a229f1890d0fdced14842bce75744c_22_02_03_15_36_01) is failing on evergreen. This PR fixes the issue.

## Checklist

### DIFF
--- a/e2e/atlas/helper_test.go
+++ b/e2e/atlas/helper_test.go
@@ -73,7 +73,6 @@ const (
 	tierM30                      = "M30"
 	diskSizeGB40                 = "40"
 	diskSizeGB30                 = "30"
-	mcliOpsManagerURL            = "https://cloud-dev.mongodb.com/"
 )
 
 // Cluster settings.
@@ -246,10 +245,11 @@ func RandProjectNameWithPrefix(prefix string) (string, error) {
 func MongoDBMajorVersion() (string, error) {
 	atlasClient := mongodbatlas.NewClient(nil)
 	atlasURL := os.Getenv("MCLI_OPS_MANAGER_URL")
-	if atlasURL != mcliOpsManagerURL {
-		return "", fmt.Errorf("the env variable %s was %s but it must be %s", "MCLI_OPS_MANAGER_URL", atlasURL, mcliOpsManagerURL)
+	baseURL, err := url.Parse(atlasURL)
+	if err != nil {
+		return "", err
 	}
-	atlasClient.BaseURL, _ = url.Parse(atlasURL)
+	atlasClient.BaseURL = baseURL
 	version, _, err := atlasClient.DefaultMongoDBMajorVersion.Get(context.Background())
 	if err != nil {
 		return "", err
--- a/e2e/atlas/helper_test.go
+++ b/e2e/atlas/helper_test.go
@@ -23,6 +23,8 @@ import (
 	"os/exec"
 	"testing"
 
+	"golang.org/x/net/context"
+
 	"github.com/mongodb/mongocli/e2e"
 	"github.com/stretchr/testify/assert"
 	"go.mongodb.org/atlas/mongodbatlas"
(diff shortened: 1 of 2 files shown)


gssbzn (changes requested):
this failing because the new default is going to be 5.0  this may require a better fix than anew env

andreaangiolillo:
> this may require a better fix than anew env

@gssbzn I have initially tried to set `--mdbVersion` in the create command but it is not available for M0. 
We could update`e2eMDBVer` to `5.0` instead of creating a new env. I don't see other options. Am I missing something? Thanks!

gssbzn:
the best thing to do would be to use default version endpoint so we don't need to manually update this all the time

https://github.com/mongodb/go-client-mongodb-atlas/blob/master/mongodbatlas/default_mongodb_major_version.go

andreaangiolillo:
@gssbzn Are we okay with adding a new hidden command to call this private endpoint? os should I call the endpoint with server request in the `helper_test.go`?

gssbzn:
> are we okay with adding a new hidden command to call this private endpoint? os should I call the endpoint with a server request in the helper_test.go?

Which ever approach you think is the best

e2e/atlas/helper_test.go
```diff
@@ -240,6 +242,17 @@ func RandProjectNameWithPrefix(prefix string) (string, error) {
 	return prefixedName, nil
 }
 
+func MongoDBMajorVersion() (string, error) {
+	atlasClient := mongodbatlas.NewClient(nil)
+	atlasClient.BaseURL, _ = url.Parse("https://cloud-dev.mongodb.com/")
```
andreaangiolillo, line 247: the endpoint returns 4.4 in production

andreaangiolillo, line 247: the endpoint returns 4.4 in production

gssbzn: please use `MCLI_OPS_MANAGER_URL` for this, don't hard code it

e2e/atlas/helper_test.go
```diff
@@ -240,6 +242,18 @@ func RandProjectNameWithPrefix(prefix string) (string, error) {
 	return prefixedName, nil
 }
 
+func MongoDBMajorVersion() (string, error) {
+	atlasClient := mongodbatlas.NewClient(nil)
+	atlasURL := os.Getenv("MCLI_OPS_MANAGER_URL")
+	atlasClient.BaseURL, _ = url.Parse(atlasURL)
```
gssbzn, line 248: you should not ignore this error in case `MCLI_OPS_MANAGER_URL` is not set or has a wrong value

e2e/atlas/helper_test.go
```diff
@@ -240,6 +242,18 @@ func RandProjectNameWithPrefix(prefix string) (string, error) {
...
+	version, _, err := atlasClient.DefaultMongoDBMajorVersion.Get(context.TODO())
```
gssbzn, line 249: let's use background instead of TODO
```suggestion
	version, _, err := atlasClient.DefaultMongoDBMajorVersion.Get(context.Background())
```

gssbzn (commented):
Almost there, two small nits

e2e/atlas/helper_test.go
andreaangiolillo: sorry I have missed "wrong value"

gssbzn: ex: `MCLI_OPS_MANAGER_URL=wrongvalue make e2e-test`

We know it's properly set on evergreen but evergreen is not the only way to run tests

e2e/atlas/helper_test.go
```diff
@@ -71,6 +73,7 @@ const (
 	tierM30                      = "M30"
 	diskSizeGB40                 = "40"
 	diskSizeGB30                 = "30"
+	mcliOpsManagerURL            = "https://cloud-dev.mongodb.com/"
```
gssbzn, line 76: Please don't
```suggestion
```

e2e/atlas/helper_test.go
```diff
@@ -240,6 +243,21 @@ func RandProjectNameWithPrefix(prefix string) (string, error) {
 	return prefixedName, nil
 }
 
+func MongoDBMajorVersion() (string, error) {
+	atlasClient := mongodbatlas.NewClient(nil)
+	atlasURL := os.Getenv("MCLI_OPS_MANAGER_URL")
+	if atlasURL != mcliOpsManagerURL {
```
gssbzn, line 249: [nit] what if I actually want to run the tests against production or my local instance, the main point of checking the URL is parsable is to catch human error/typos, not to forbid the use of any other url that's not dev

gssbzn (approved):
LGTM thanks for all the changes and keeping our tests green

PR #977, opened 2022-02-10
CLOUDP-112236: Update profile.go

andreaangiolillo:
## Proposed changes

_Jira ticket:_ CLOUDP-112236

## Checklist

## Further comments

**Description:**

- Update `profile.go` to use the correct config file. We will store the config file in` ~/.config/atlascli/config.toml.` 
- To improve the transition from mongocli to atlascli if a mongocli config exists and no atlas config exists we generate the atlascli config file from the mongocli one. We will print a message to stderr when doing this operation

**With mongocli.toml** 
```
./bin/atlas cluster ls
AtlasCLI has found an existing MongoCLI configuration file, copying its content to: /Users/andrea.angiolillo/.config/atlascli/config.toml
AtlasCLI has copied your MongoCLI configuration to: /Users/andrea.angiolillo/.config/atlascli/config.toml

ID                         NAME                    MDB VER   STATE
620a21fcc60a18018675c3bc   Quickstart-1644831225   5.0.6     IDLE
```

**No mongocli.toml**
```
./bin/atlas cluster list
Error: missing credentials

To set credentials, run: atlas config
```


internal/config/profile.go
```diff
@@ -573,7 +579,65 @@ func (p *Profile) Save() error {
...
+		return false
+	}
+	defer out.Close()
+
+	_, err = io.Copy(out, in)
+	if err != nil {
+		return false
+	}
+
+	_, _ = fmt.Fprintf(os.Stderr, `we have used %s to generate %s
```
andreaangiolillo, line 614: I don't like this message. Open to suggestions

blva (dismissed):
LGTM, 2 suggestions but no strong opinions

internal/config/profile.go
```diff
@@ -151,7 +154,7 @@ func Exists(name string) bool {
 }
 
 func newProfile() *Profile {
-	configDir, err := configHome()
+	configDir, err := configHome(ToolName)
 	if err != nil {
 		log.Fatal(err)
 	}
```
andreaangiolillo, line 160: @gssbzn I am not sure I understood what you proposed. I am already injecting the config dir based on the `ToolName`in `newProfile`. I am not sure what `newAtlasProfile` should do.

gssbzn: what you are currently doing is setting the configDir based on the global state of `ToolName` you are injecting to `p` yes but `newProfile` has no concept of injection, new profile should not have an if to decide  what home dir to use
it's better to keep `newProfile` as is, and create a new method, `newAtlasProfile` and none of them have an if based on global state

andreaangiolillo: we still have to use the global state of `ToolName` to decide which newProfile function to call, right? Where should I add this logic? I see that we are using `var p = newProfile()` to create a new profile now

internal/config/profile.go
```diff
@@ -573,7 +579,65 @@ func (p *Profile) Save() error {
 	return viper.WriteConfigAs(p.Filename())
 }
 
-func configHome() (string, error) {
+func (p *Profile) copyMongoCLIConfig() bool {
```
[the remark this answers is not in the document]
andreaangiolillo: discussing the second suggestion in the other open comment.

internal/config/profile.go
```diff
@@ -573,7 +569,31 @@ func (p *Profile) Save() error {
 	return viper.WriteConfigAs(p.Filename())
 }
 
-func configHome() (string, error) {
+// ReadMongoCLIConfig retrieves mongoCLI configuration.
+func ReadMongoCLIConfig() string { return p.readMongoCLIConfig() }
+func (p *Profile) readMongoCLIConfig() string {
```
gssbzn, line 574: `ReadMongoCLIConfig` names hint at loading the file into memory, which this method doesn't seem to be doing. Your only use of `p` is to get the file system wrapper, `ReadMongoCLIConfig` and `readMongoCLIConfig` are not used is this package, does this need to be in this package?

cmd/atlas/atlas.go
```diff
@@ -53,6 +62,74 @@ func initConfig() {
 	}
 }
 
+// createConfigFromMongoCLIConfig creates atlasCLI config file from mongoCLI config file.
+func createConfigFromMongoCLIConfig(err error) bool {
+	// search mongoCLI config only if atlasCLI config doesn't exist
+	var e viper.ConfigFileNotFoundError
+	if !errors.As(err, &e) {
+		return false
+	}
```
gssbzn, line 71: don't call `createConfigFromMongoCLIConfig` instead of passing an error

cmd/atlas/atlas.go
```diff
@@ -53,6 +62,74 @@ func initConfig() {
...
+
+	_, _ = fmt.Fprintf(os.Stderr, `we have used %s to generate %s
+
+`, mongoCLIConfigPath, atlasCLIConfigPath)
+
+	return true
+}
+
+// copyConfig copies config in oldConfigPath to the correct config location.
+func copyConfig(oldConfigPath string) string {
```
gssbzn, line 91: why is this method swallowing all the errors?

internal/config/profile.go
gssbzn: `var p = newProfile()` was meant to simulate a singleton for convenience, this introduced some problems on testing and why we have slowly moved to inject `config.Default()` into commands instead of using `config.SetX` or `config.GetX` we should move away from those global setters and global state.

Prefer a specific method, ie `config.LoadAtlas`, `config.SaveAtlas` and the only place to probably compromise is `config.Default` to have an `if` to keep it backward compatible

gssbzn (commented):
better but not there yet

cmd/atlas/atlas.go
andreaangiolillo: if we get an error in copying the mongocli config we don't print it but we follows the "regular" flow: if the error in `config.Load()` is `ConfigFileNotFoundError`, go-ahead. Otherwise, print the error

cmd/atlas/atlas.go
```diff
@@ -53,6 +62,74 @@ func initConfig() {
...
+	atlasCLIConfigPath := copyConfig(mongoCLIConfigPath)
+	if atlasCLIConfigPath == "" {
+		return false
+	}
+
+	_, _ = fmt.Fprintf(os.Stderr, `we have used %s to generate %s
+
+`, mongoCLIConfigPath, atlasCLIConfigPath)
+
+	return true
```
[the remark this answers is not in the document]
andreaangiolillo: they are used in the `initConfig()` here 

```
		// we use mongocli.toml to generate atlasCLI config
		if !createConfigFromMongoCLIConfig(err) {
			printError(err)
		} else if err := config.Load(); err != nil {
			printError(err)
		}
	}
```

if `createConfigFromMongoCLIConfig` == true -> try to load again the config
if `createConfigFromMongoCLIConfig` == false -> print the error if err == `ConfigFileNotFoundError` as it was before

cmd/atlas/atlas.go
andreaangiolillo: updated

internal/config/profile.go
andreaangiolillo: updated

cmd/atlas/atlas.go
```diff
@@ -40,8 +44,14 @@ func Execute(ctx context.Context) {
 
 // initConfig reads in config file and ENV variables if set.
 func initConfig() {
-	if err := config.Load(); err != nil {
-		log.Fatalf("Error loading config: %v", err)
+	if err := config.LoadAtlasCLIConfig(); err != nil {
+		// we use mongocli.toml to generate atlasCLI config
+		var e viper.ConfigFileNotFoundError
+		if !errors.As(err, &e) || !createConfigFromMongoCLIConfig() { // search mongoCLI config only if atlasCLI config doesn't exist
+			printError(err)
```
gssbzn, line 51: double casting of err into `viper.ConfigFileNotFoundError`

internal/config/profile.go
```diff
@@ -527,11 +550,20 @@ func (p *Profile) Rename(newProfileName string) error {
...
+}
+
+func LoadMongoCLIConfig() error { return Default().LoadMongoCLIConfig(true) }
+func (p *Profile) LoadMongoCLIConfig(readEnvironmentVars bool) error {
 	viper.SetConfigName(ToolName)
+	return p.load(readEnvironmentVars)
+}
+
+func (p *Profile) load(readEnvironmentVars bool) error {
+	viper.SetConfigType(configType)
```
gssbzn, line 566: just a reminder that `viper.SetEnvPrefix(EnvPrefix)` needs to change, not needed for this PR

internal/config/profile.go
```diff
@@ -585,3 +611,18 @@ func configHome() (string, error) {
 
 	return fmt.Sprintf("%s/.config", home), nil
 }
+
+// AtlasCLIConfigHome retrieves configHome path based used by atlasCLI.
+func AtlasCLIConfigHome() (string, error) { return atlasCLIConfigHome() }
+func atlasCLIConfigHome() (string, error) {
```
gssbzn, line 617: do you need a public and private method that both do the same?

cmd/atlas/atlas.go
andreaangiolillo: it is wanted: In the `if` I check if the error is not `ConfigFileNotFoundError`: in this case, I have to print the error. If the error is `ConfigFileNotFoundError` (config not found), I need to run the `createConfigFromMongoCLIConfig`. This function may not be successful, in that case I have to check again if the error was  `ConfigFileNotFoundError` because we don't print error in that case

internal/config/profile.go
andreaangiolillo: yes, I have this ticket https://jira.mongodb.org/browse/CLOUDP-112237 for that

internal/config/profile.go
andreaangiolillo: no, I will update this

cmd/atlas/atlas.go
andreaangiolillo: this check was first inside `createConfigFromMongoCLIConfig` but it was moved here to address this https://github.com/mongodb/mongocli/pull/977#discussion_r804593264

internal/config/profile.go
andreaangiolillo: Updated

cmd/atlas/atlas.go
```diff
@@ -40,8 +44,14 @@ func Execute(ctx context.Context) {
 
 // initConfig reads in config file and ENV variables if set.
 func initConfig() {
-	if err := config.Load(); err != nil {
-		log.Fatalf("Error loading config: %v", err)
+	if err := config.LoadAtlasCLIConfig(); err != nil {
```
gssbzn, line 47: I can understand this back and forth may start becoming frustrating but a lot of this changes are making the logic more complicated than it needs to be which also makes reviewing this PR harder than a normal PR, main point logic is going all over the place so here's a high level of what I'd like to see

You don't need to  try to load the config to then try to copy an old config from the cli, there's no direct dependency and you can decouple this.

To do so `cobra.OnInitialize` takes a list of functions, make copying an old config the first thing you do, an example
```go
source, err := os.Open(src)
if err != nil {
  return // file doesn't esist or can't be processed
  // there's also errors.Is(err, os.ErrNotExist) if you want to actually check for not exists
}
defer source.Close()

destination, err := os.Create(dst)
if err != nil {
  return 0, err
}
defer destination.Close()

if _, err := io.Copy(destination, source); err != nil {
// good idea to try to log the error here
}
```
then `initConfig` can stay as it was, the old file was copied on a previous step and there's no tangling between the operations

andreaangiolillo: Updated

cmd/atlas/atlas.go
```diff
@@ -53,9 +55,73 @@ func initConfig() {
...
+	if mongoCLIConfigPath == "" {
+		return
+	}
+
+	in, err := os.Open(mongoCLIConfigPath)
+	if err != nil {
+		return
+	}
+	defer in.Close()
+
```
gssbzn, line 82: let's add  `_, _ = fmt.Fprintf` here, soemthing like
> Atlas CLI found an existing mongocli configuration, copying its content to: ${atlas path}

cmd/atlas/atlas.go
```diff
@@ -53,9 +55,73 @@ func initConfig() {
...
+	out, err := os.Create(atlasConfigPath)
+	if err != nil {
+		return
+	}
+	defer out.Close()
+
+	_, err = io.Copy(out, in)
+	if err != nil {
+		log.Printf("we cannot create %s from your %s: %v", atlasConfigPath, mongoCLIConfigPath, err)
```
gssbzn, line 99:
```suggestion
		log.Printf("There was an error generating %s: %v", atlasConfigPath, mongoCLIConfigPath, err)
```

cmd/atlas/atlas.go
```diff
@@ -53,9 +55,73 @@ func initConfig() {
...
+	}
+
+	_, _ = fmt.Fprintf(os.Stderr, `we have used %s to generate %s
+
+`, mongoCLIConfigPath, atlasConfigPath)
```
gssbzn, line 104: I think We can delete this, or change it to be more generic 

> Atlas CLI has copied your MongoCLI config to: %s

gssbzn (approved):
Thanks for all the work here, minor comments tho not blocking, would be good to review the copy on some of the messages

cmd/atlas/atlas.go
```diff
@@ -53,9 +55,73 @@ func initConfig() {
...
+	configDir, err := config.MongoCLIConfigHome()
+	if err != nil {
+		return ""
+	}
+
+	configPath := fmt.Sprintf("%s/mongocli.toml", configDir)
+	_, err = os.Stat(configPath)
+	if err != nil {
+		return ""
+	}
```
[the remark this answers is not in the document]
andreaangiolillo: I've missed this comment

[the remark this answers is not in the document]
andreaangiolillo: Updated

cmd/atlas/atlas.go
andreaangiolillo: updated

cmd/atlas/atlas.go
andreaangiolillo: Updated

cmd/atlas/atlas.go
andreaangiolillo: Updated

gssbzn (approved):
LGTM

PR #1015, opened 2022-03-03
CLOUDP-114248: Store all config files with mongocli folder

blva:
## Proposed changes

_Jira ticket:_ [CLOUDP-114366](https://jira.mongodb.org/browse/CLOUDP-114366)

## Checklist

## Further comments

* While loading the config, if the file doesn't exist, check the old path` ~/.config `and load the file copying to the new path `~/.config/mongocli/config.toml`. 
* ~Note: Still trying to figure out linux test failure - but sending for early review~
* ~Note: some evergreen files need to be updated to address this, but I want to run an e2e test without the change to see this solution working.~

### DIFF
--- a/build/ci/evergreen.yml
+++ b/build/ci/evergreen.yml
@@ -344,7 +344,7 @@ tasks:
           script: |
             set -Eeou pipefail
 
-            cat <<EOF > config.toml
+            cat <<EOF > mongocli/config.toml
             [e2e]
               org_id = "5e429e7706822c6eac4d5971"
               public_api_key = "AAUMGJXA"
--- a/build/ci/evergreen.yml
+++ b/build/ci/evergreen.yml
@@ -343,8 +343,8 @@ tasks:
         params:
           script: |
             set -Eeou pipefail
-
-            cat <<EOF > mongocli/config.toml
+            mkdir mongocli &&  cd mongocli
+            cat <<EOF > config.toml
             [e2e]
               org_id = "5e429e7706822c6eac4d5971"
               public_api_key = "AAUMGJXA"
--- a/build/ci/evergreen.yml
+++ b/build/ci/evergreen.yml
@@ -344,7 +344,7 @@ tasks:
           script: |
             set -Eeou pipefail
 
-            cat <<EOF > mongocli.toml
+            cat <<EOF > config.toml
             [e2e]
               org_id = "5e429e7706822c6eac4d5971"
               public_api_key = "AAUMGJXA"
--- a/build/ci/evergreen.yml
+++ b/build/ci/evergreen.yml
@@ -343,8 +343,8 @@ tasks:
         params:
           script: |
             set -Eeou pipefail
-            mkdir mongocli &&  cd mongocli
-            cat <<EOF > config.toml
+            mkdir mongocli
+            cat <<EOF > mongocli/config.toml
             [e2e]
               org_id = "5e429e7706822c6eac4d5971"
               public_api_key = "AAUMGJXA"
--- a/build/ci/set-up-cloud-manager.sh
+++ b/build/ci/set-up-cloud-manager.sh
@@ -18,6 +18,7 @@ set -euo pipefail
 
 export MCLI_SERVICE="${cloud_manager_service:?}"
 
+mkdir "${XDG_CONFIG_HOME}/mongocli"
 cat <<EOF > "${XDG_CONFIG_HOME}/mongocli/config.toml"
 skip_update_check = true
 [default]
--- a/build/ci/set-up-cloud-manager.sh
+++ b/build/ci/set-up-cloud-manager.sh
@@ -18,7 +18,7 @@ set -euo pipefail
 
 export MCLI_SERVICE="${cloud_manager_service:?}"
 
-cat <<EOF > "${XDG_CONFIG_HOME}/mongocli.toml"
+cat <<EOF > "${XDG_CONFIG_HOME}/config.toml"
 skip_update_check = true
 [default]
   service = "${MCLI_SERVICE}"
--- a/build/ci/set-up-cloud-manager.sh
+++ b/build/ci/set-up-cloud-manager.sh
@@ -18,7 +18,7 @@ set -euo pipefail
 
 export MCLI_SERVICE="${cloud_manager_service:?}"
 
-cat <<EOF > "${XDG_CONFIG_HOME}/config.toml"
+cat <<EOF > "${XDG_CONFIG_HOME}/mongocli/config.toml"
 skip_update_check = true
 [default]
   service = "${MCLI_SERVICE}"
--- a/build/ci/set-up-ops-manager.sh
+++ b/build/ci/set-up-ops-manager.sh
@@ -75,6 +75,7 @@ echo "create project"
 AGENT_API_KEY=$(./bin/mongocli iam projects create myProj -o="go-template={{.AgentAPIKey}}")
 MCLI_PROJECT_ID=$(./bin/mongocli iam project list -o="go-template={{ (index .Results 0).ID }}")
 
+mkdir "${XDG_CONFIG_HOME}/mongocli"
 cat <<EOF > "${XDG_CONFIG_HOME}/mongocli/config.toml"
 skip_update_check = true
 [default]
--- a/build/ci/set-up-ops-manager.sh
+++ b/build/ci/set-up-ops-manager.sh
@@ -75,7 +75,7 @@ echo "create project"
 AGENT_API_KEY=$(./bin/mongocli iam projects create myProj -o="go-template={{.AgentAPIKey}}")
 MCLI_PROJECT_ID=$(./bin/mongocli iam project list -o="go-template={{ (index .Results 0).ID }}")
 
-cat <<EOF > "${XDG_CONFIG_HOME}/config.toml"
+cat <<EOF > "${XDG_CONFIG_HOME}/mongocli/config.toml"
 skip_update_check = true
 [default]
   ops_manager_url = "${MCLI_OPS_MANAGER_URL}"
--- a/build/ci/set-up-ops-manager.sh
+++ b/build/ci/set-up-ops-manager.sh
@@ -75,7 +75,7 @@ echo "create project"
 AGENT_API_KEY=$(./bin/mongocli iam projects create myProj -o="go-template={{.AgentAPIKey}}")
 MCLI_PROJECT_ID=$(./bin/mongocli iam project list -o="go-template={{ (index .Results 0).ID }}")
 
-cat <<EOF > "${XDG_CONFIG_HOME}/mongocli.toml"
+cat <<EOF > "${XDG_CONFIG_HOME}/config.toml"
 skip_update_check = true
 [default]
   ops_manager_url = "${MCLI_OPS_MANAGER_URL}"
--- a/cmd/atlas/atlas.go
+++ b/cmd/atlas/atlas.go
@@ -55,7 +55,7 @@ func initConfig() {
 	}
 }
 
-// createConfigFromMongoCLIConfig creates the atlasCLI config file from the mongocli.toml.
+// createConfigFromMongoCLIConfig creates the atlasCLI config file from the mongocli config file.
 func createConfigFromMongoCLIConfig() {
 	atlasConfigHomePath, err := config.AtlasCLIConfigHome()
 	if err != nil {
--- a/cmd/atlas/atlas.go
+++ b/cmd/atlas/atlas.go
@@ -65,9 +65,9 @@ func createConfigFromMongoCLIConfig() {
 	atlasConfigPath := fmt.Sprintf("%s/%s", atlasConfigHomePath, "config.toml")
 	f, err := os.Open(atlasConfigPath) // if config.toml is already there, exit
 	if err == nil {
+		f.Close()
 		return
 	}
-	defer f.Close()
 
 	in, err := mongoCLIConfigFile()
 	if err != nil {
@@ -102,41 +102,27 @@ func createConfigFromMongoCLIConfig() {
 `, atlasConfigPath)
 }
 
-func mongoCLIConfigPaths() (configPath, oldConfigPath string) {
-	configPath = ""
-	oldConfigPath = ""
+func mongoCLIConfigFile() (in *os.File, err error) {
+	configPath := ""
+	configDir := ""
 
-	configDir, err := config.MongoCLIConfigHome()
-	if err == nil {
+	if configDir, err = config.MongoCLIConfigHome(); err == nil {
 		configPath = fmt.Sprintf("%s/config.toml", configDir)
 	}
 
-	oldConfigDir, err := config.OldMongoCLIConfigHome()
-	if err == nil {
-		oldConfigPath = fmt.Sprintf("%s/mongocli.toml", oldConfigDir)
-	}
-
-	return configPath, oldConfigPath
-}
-
-func mongoCLIConfigFile() (*os.File, error) {
-	mongoCLIConfigPath, oldMongoCLIConfigPath := mongoCLIConfigPaths()
-	if mongoCLIConfigPath == "" && oldMongoCLIConfigPath == "" {
-		return nil, nil
-	}
-
-	// Newest file is available, ignore the old one
-	in, err := os.Open(mongoCLIConfigPath)
+	// Try to open new file
+	in, err = os.Open(configPath)
 	if err == nil {
 		return in, nil
 	}
 
-	in, err = os.Open(oldMongoCLIConfigPath)
-	if err == nil {
-		return in, nil
+	// Try finding old file since new file opening failed/did not exist
+	if configDir, err = config.OldMongoCLIConfigHome(); err == nil {
+		configPath = fmt.Sprintf("%s/mongocli.toml", configDir)
 	}
 
-	return nil, err
+	// Try to open old file
+	return os.Open(configPath)
 }
 
 func main() {
--- a/cmd/atlas/atlas.go
+++ b/cmd/atlas/atlas.go
@@ -99,6 +99,7 @@ func createConfigFromMongoCLIConfig() {
 	_, err = io.Copy(out, in)
 	if err != nil {
 		log.Printf("There was an error generating %s: %v", atlasConfigPath, err)
+		return
 	}
 
 	_, _ = fmt.Fprintf(os.Stderr, `AtlasCLI has copied your MongoCLI configuration to: %s
--- a/cmd/atlas/atlas.go
+++ b/cmd/atlas/atlas.go
@@ -69,7 +69,12 @@ func createConfigFromMongoCLIConfig() {
 		return
 	}
 
-	in, err := mongoCLIConfigFile()
+	path, err := mongoCLIConfigFilePath()
+	if err != nil {
+		return
+	}
+
+	in, err := os.Open(path)
 	if err != nil {
 		return
 	}
@@ -102,7 +107,7 @@ func createConfigFromMongoCLIConfig() {
 `, atlasConfigPath)
 }
 
-func mongoCLIConfigFile() (in *os.File, err error) {
+func mongoCLIConfigFilePath() (path string, err error) {
 	configPath := ""
 	configDir := ""
 
@@ -110,19 +115,19 @@ func mongoCLIConfigFile() (in *os.File, err error) {
 		configPath = fmt.Sprintf("%s/config.toml", configDir)
 	}
 
-	// Try to open new file
-	in, err = os.Open(configPath)
-	if err == nil {
-		return in, nil
+	// Check if file exists, if any error is detected try to get older file
+	if _, err = os.Stat(configPath); err == nil {
+		return configPath, nil
 	}
 
-	// Try finding old file since new file opening failed/did not exist
 	if configDir, err = config.OldMongoCLIConfigHome(); err == nil {
 		configPath = fmt.Sprintf("%s/mongocli.toml", configDir)
 	}
 
-	// Try to open old file
-	return os.Open(configPath)
+	if _, err = os.Stat(configPath); err == nil {
+		return configPath, nil
+	}
+	return "", err
 }
 
 func main() {
--- a/cmd/atlas/atlas.go
+++ b/cmd/atlas/atlas.go
@@ -69,12 +69,7 @@ func createConfigFromMongoCLIConfig() {
 	}
 	defer f.Close()
 
-	mongoCLIConfigPath := mongoCLIConfigPath()
-	if mongoCLIConfigPath == "" {
-		return
-	}
-
-	in, err := os.Open(mongoCLIConfigPath)
+	in, err := mongoCLIConfigFile()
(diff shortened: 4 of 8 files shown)


internal/config/profile.go
```diff
@@ -618,13 +618,38 @@ func (p *Profile) load(readEnvironmentVars bool, envPrefix string) error {
...
+				_ = p.copyOldConfig(readEnvironmentVars, envPrefix)
+			}
 			return nil
 		}
 		return err
 	}
 	return nil
 }
 
+func (p *Profile) copyOldConfig(readEnvironmentVars bool, envPrefix string) error {
```
gssbzn, line 631: [nit] let's not do this here, let's follow a similar approach to what was done for atlas cli, atlas cli will read if a mongocli config exists and copy it to the atlas cli config path, this check is done at run time when calling a command,
Also notice that logic may need to be updated as well to account for two possible paths

blva: Do you mean [this](https://github.com/mongodb/mongocli/blob/098509e33af5bba51b4429bcaae7961231c5c15e/cmd/atlas/atlas.go#L59)? 

I initially thought of following the same approach but overall I thought the logic would look a bit complex if I followed that approach.

Since we load the profile at the beginning of the command, we get an error if we update `MongoCLIConfigHome()` ->` Error: credentials not set`. 

Maybe you thought about something else but what I thought was that we would need to create another profile `oldProfile` and a `newProfile` and try loading `newProfile`, then try `oldProfile` if it doesn't exist until one of them exists. If oldProfile exists we return it and then the atlas command will remove the file, making oldProfile invalid and newProfile loadable. So my understanding was that in the end we would need changes in both files and more clutter to make it work. Again this was what I thought as a solution, please let me know if there is a better way I did not see.

internal/config/profile.go
```diff
@@ -641,19 +666,32 @@ func (p *Profile) Save() error {
...
+	}
+
+	home, err := os.UserHomeDir()
+	if err != nil {
+		return "", err
+	}
+
+	return fmt.Sprintf("%s/.config", home), nil
+}
+
```
andreaangiolillo, line 682: Could you add deprecatation comments saying that mongocli version prior v1.24.0 uses this path and that new versions use `MongoCLIConfigHome`? 

I'd also rename the function as `DeprecatedMongoCLIConfigHome` but this is just my preference

andreaangiolillo: > add a paragraph to its doc comment that begins with “Deprecated:” followed by some information about the deprecation.

https://go.dev/blog/godoc

internal/config/profile.go
andreaangiolillo: +1

gssbzn: >Do you mean [this](https://github.com/mongodb/mongocli/blob/098509e33af5bba51b4429bcaae7961231c5c15e/cmd/atlas/atlas.go#L59)?

yes

> Since we load the profile at the beginning of the command, we get an error if we update MongoCLIConfigHome() -> Error: credentials not set.

The logic above runs before initialziing the profile, it's actually nicely seen here
```
	cobra.OnInitialize(createConfigFromMongoCLIConfig, initConfig)
```
first copy then load config

> I thought was that we would need to create another profile oldProfile and a newProfile and try loading newProfile, then try oldProfile if it doesn't exist until one of them exists

Too many responsibilities for the profile to have to remember there's actually two files, Andrea and I already went down discussing the approach for the linked piece of code, you can revisit [here](https://github.com/mongodb/mongocli/pull/977#discussion_r804860498)

And just in case the linked discussion/approach keep the existing mongocli config, we don't need that part we can move it without keeping a back up. This is also why profile should not have a fallback logic, the entry point should always make sure the file is in the right place 

@andreaangiolillo you may want to expand your `+1` as it's not clear what idea are you supporting here

andreaangiolillo: I added my comment when Bianca's reply wasn’t there 🤷‍♂️ .

I agree that we should use the same logic that we are using with atlascli

blva: cool, thanks for the explanation! let me try to draft this

cmd/atlas/atlas.go
```diff
@@ -99,6 +99,7 @@ func createConfigFromMongoCLIConfig() {
 	_, err = io.Copy(out, in)
 	if err != nil {
 		log.Printf("There was an error generating %s: %v", atlasConfigPath, err)
+		return
```
blva, line 102: pls check if this makes sense

build/ci/set-up-cloud-manager.sh
```diff
@@ -18,7 +18,7 @@ set -euo pipefail
 
 export MCLI_SERVICE="${cloud_manager_service:?}"
 
-cat <<EOF > "${XDG_CONFIG_HOME}/mongocli.toml"
+cat <<EOF > "${XDG_CONFIG_HOME}/config.toml"
```
gssbzn, line 21: [q] what about the folder?

build/ci/evergreen.yml
```diff
@@ -344,7 +344,7 @@ tasks:
           script: |
             set -Eeou pipefail
 
-            cat <<EOF > mongocli.toml
+            cat <<EOF > config.toml
```
gssbzn, line 347: [q] what about the folder?

build/ci/set-up-ops-manager.sh
```diff
@@ -75,7 +75,7 @@ echo "create project"
 AGENT_API_KEY=$(./bin/mongocli iam projects create myProj -o="go-template={{.AgentAPIKey}}")
 MCLI_PROJECT_ID=$(./bin/mongocli iam project list -o="go-template={{ (index .Results 0).ID }}")
 
-cat <<EOF > "${XDG_CONFIG_HOME}/mongocli.toml"
+cat <<EOF > "${XDG_CONFIG_HOME}/config.toml"
```
gssbzn, line 78: [q] what about the folder?

internal/config/profile.go
```diff
@@ -534,7 +534,7 @@ func (p *Profile) Delete() error {
 }
 
 func (p *Profile) Filename() string {
-	return filepath.Join(p.configDir, ToolName+".toml")
+	return filepath.Join(p.configDir, "config.toml")
```
andreaangiolillo, line 537: FYI: we were storing the atlasCLI config in the path `.config/atlascli/atlascli.toml` instead of `.config/atlascli/config.toml`. This should fix the issue 💯

build/ci/set-up-ops-manager.sh
blva: It looks like when `XDG_CONFIG_HOME` is set for atlascli [we don't try to add the folder to the path ](https://github.com/mongodb/mongocli/blob/f319c546a8f66a1f6d2970da6e6127d758e180f0/internal/config/profile.go#L661). So I assumed that for mcli would be the same. Is that something we want to do for both?

cmd/atlas/atlas.go
```diff
@@ -99,21 +94,49 @@ func createConfigFromMongoCLIConfig() {
...
 
 	_, _ = fmt.Fprintf(os.Stderr, `AtlasCLI has copied your MongoCLI configuration to: %s
 
 `, atlasConfigPath)
 }
 
-func mongoCLIConfigPath() string {
+func mongoCLIConfigPaths() (configPath, oldConfigPath string) {
+	configPath = ""
+	oldConfigPath = ""
```
andreaangiolillo, line 107: [nit] you can remove these vars as they are already defined

cmd/atlas/atlas.go
```diff
@@ -99,21 +94,49 @@ func createConfigFromMongoCLIConfig() {
...
+
 	configDir, err := config.MongoCLIConfigHome()
-	if err != nil {
-		return ""
+	if err == nil {
+		configPath = fmt.Sprintf("%s/config.toml", configDir)
+	}
```
andreaangiolillo, line 112:
```go
	if configDir, err := config.MongoCLIConfigHome(); err == nil{
		configPath = fmt.Sprintf("%s/config.toml", configDir)
	}
```

cmd/atlas/atlas.go
```diff
@@ -99,21 +94,49 @@ func createConfigFromMongoCLIConfig() {
...
+
+	oldConfigDir, err := config.OldMongoCLIConfigHome()
+	if err == nil {
+		oldConfigPath = fmt.Sprintf("%s/mongocli.toml", oldConfigDir)
+	}
```
andreaangiolillo, line 117:
```suggestion
	if oldConfigDir, err := config.OldMongoCLIConfigHome(); err == nil {
		oldConfigPath = fmt.Sprintf("%s/mongocli.toml", oldConfigDir)
	}
```

cmd/mongocli/mongocli.go
```diff
@@ -38,6 +40,59 @@ func Execute(ctx context.Context) {
...
+	if err != nil {
+		return
+	}
+
+	mongoCLIConfigPath := fmt.Sprintf("%s/%s", mongoCLIConfigHome, "config.toml")
+	f, err := os.Open(mongoCLIConfigPath) // if config.toml is already there, exit
+	if err == nil {
+		return
+	}
+	defer f.Close()
```
gssbzn, line 54: you're leaking these file since you'll never close it

cmd/mongocli/mongocli.go
```diff
@@ -38,6 +40,59 @@ func Execute(ctx context.Context) {
 	}
 }
 
+func updateMongoCLIConfigPath() {
+	mongoCLIConfigHome, err := config.MongoCLIConfigHome()
+	if err != nil {
+		return
+	}
+
+	mongoCLIConfigPath := fmt.Sprintf("%s/%s", mongoCLIConfigHome, "config.toml")
```
gssbzn, line 49: I'm just realising this, but, does this work for windows or do we need to use `filepath.Join`?

cmd/mongocli/mongocli.go
```diff
@@ -38,6 +40,59 @@ func Execute(ctx context.Context) {
...
+	in, err := os.Open(oldMongoCLIConfigPath)
+	if err != nil {
+		return
+	}
+	defer in.Close()
+
+	_, _ = fmt.Fprintf(os.Stderr, `MongoCLI uses a new config path. Copying mongocli.toml content to: %s
+`, mongoCLIConfigPath)
+
+	_, err = os.Stat(oldMongoCLIConfigHome) // check if the dir is already there
```
andreaangiolillo, line 71: you are already trying to open the old config in line 62 `in, err := os.Open(oldMongoCLIConfigPath)`, if you reach this point the dir should be already there, right?

cmd/mongocli/mongocli.go
```diff
@@ -38,6 +40,59 @@ func Execute(ctx context.Context) {
...
+	if err != nil {
+		defaultPermissions := 0700
+		if err = os.Mkdir(mongoCLIConfigHome, os.FileMode(defaultPermissions)); err != nil {
+			return
+		}
```
andreaangiolillo, line 76: I am not sure why you are checking for the `oldMongoCLIConfigHome` and then create the `mongoCLIConfigHome` if the old dir is not there

build/ci/set-up-ops-manager.sh
gssbzn: oh I think this uncovers a bug on the atlas implementation when `XDG_CONFIG_HOME` is set the config path for both atlas and now mongocli should be `XDG_CONFIG_HOME/{mongocli|atlascli}/config.toml`

cmd/atlas/atlas.go
```diff
@@ -99,21 +94,49 @@ func createConfigFromMongoCLIConfig() {
...
+		return
 	}
 
 	_, _ = fmt.Fprintf(os.Stderr, `AtlasCLI has copied your MongoCLI configuration to: %s
 
 `, atlasConfigPath)
 }
 
-func mongoCLIConfigPath() string {
+func mongoCLIConfigPaths() (configPath, oldConfigPath string) {
```
gssbzn, line 105: is this method necessary? you're probably only gonna use one, can we just call them separately instead of a method to get both at the same time?

cmd/atlas/atlas.go
```diff
@@ -99,21 +94,49 @@ func createConfigFromMongoCLIConfig() {
...
+	}
+
+	in, err = os.Open(oldMongoCLIConfigPath)
+	if err == nil {
+		return in, nil
 	}
 
-	configPath := fmt.Sprintf("%s/mongocli.toml", configDir)
-	return configPath
+	return nil, err
```
gssbzn, line 139: `return os.Open(oldMongoCLIConfigPath)`

cmd/mongocli/mongocli.go
blva: my bad, this should be `mongoCLIConfigHome`, thank you!

build/ci/evergreen.yml
blva: fixing

cmd/atlas/atlas.go
blva: sure!

cmd/atlas/atlas.go
blva: I'll still need it since I'll move this to another method now. Please, LMK if you have another suggestion in the next rev

cmd/mongocli/mongocli.go
blva: I would assume it works since `MongoCLIConfigHome` has been using this already? How can I confirm if this works?

cmd/mongocli/mongocli.go
blva: yes, this was the wrong one as you pointed in your other comment, thx :)

cmd/mongocli/mongocli.go
andreaangiolillo: 1. spawn a windows host on evergreen
2. run the task package mongocli msi on evergreen
3. access your windows host with Microsoft Remote Desktop
4. install the msi created at step 2, on your windows machine

blva: Thanks! On it

internal/config/profile.go
```diff
@@ -641,25 +641,39 @@ func (p *Profile) Save() error {
...
+	}
+
+	return fmt.Sprintf("%s/.config", home), nil
+}
+
 // MongoCLIConfigHome retrieves configHome path based used by mongoCLI.
 func MongoCLIConfigHome() (string, error) {
 	if home := os.Getenv("XDG_CONFIG_HOME"); home != "" {
-		return home, nil
+		return fmt.Sprintf("%s/mongocli", home), nil
```
fmenezes, line 662: I don't know if that was agreed already on a spec but a more OS compatible way for the .config folder is [os.UserConfigDir](https://pkg.go.dev/os#UserConfigDir). Currently our code always defaults to `os.UserHomeDir() / .config` but the above mentioned code would try `$HOME/Library/Application Support` for macos and `%AppData%` for Windows which is probably what our customers would have expected.

Not blocking just mentioning here since we are changing folders for the CLI this might by the right moment to rethink this folder.

andreaangiolillo: I didn’t know about this. Thanks for sharing it! I am in favor of this change as we are already changing the location of the config. Tagging @gssbzn to see if we all agree on this change.

blva: sounds great, to see if I understood -  we would replace the whole logic within `MongoCLIConfigHome() `and use `os.UserConfigDir()`, furthermore we would no longer use `.config`. Just simply create a mongocli folder under `UserConfigDir`. Adding to this PR sounds reasonable.

cmd/mongocli/mongocli.go
blva: Ok, tested and we're creating the right folder structure if that's the concern here:

```
PS C:\Users\Administrator\AppData\Roaming> ls


    Directory: C:\Users\Administrator\AppData\Roaming


Mode                LastWriteTime         Length Name
----                -------------         ------ ----
d-----        1/31/2022   5:40 PM                ActiveState
d-----         3/7/2022   4:06 PM                Adobe
d---s-         3/7/2022   4:06 PM                Microsoft
d-----         3/7/2022   4:21 PM                mongocli
d-----        1/31/2022   5:57 PM                npm
d-----        1/31/2022   5:31 PM                NuGet
d-----        1/31/2022   5:51 PM                Visual Studio Setup

```
```
PS C:\Users\Administrator\AppData\Roaming> cd mongocli
PS C:\Users\Administrator\AppData\Roaming\mongocli> ls


    Directory: C:\Users\Administrator\AppData\Roaming\mongocli


Mode                LastWriteTime         Length Name
----                -------------         ------ ----
-a----         3/7/2022   4:21 PM             45 config.toml

```
```
PS C:\Users\Administrator\AppData\Roaming\mongocli> cat config.toml

[default]
  output = ""
  service = "cloud"
```

internal/config/profile.go
blva: I added the changes for this, LMK if we can proceed with this approach.

internal/config/profile.go
```diff
@@ -641,31 +641,42 @@ func (p *Profile) Save() error {
...
+	home, err := os.UserConfigDir()
+	if err == nil {
+		return fmt.Sprintf("%s/atlascli", home), nil
 	}
 
-	home, err := os.UserHomeDir()
-
-	if err != nil {
-		return "", err
+	home, err = os.UserHomeDir()
```
fmenezes, line 676: Looks like in `MongoCLIConfigHome` there is no fallback but in `AtlasCLIConfigHome` we do have a fallback to `UserHomeDir`, any reason for the difference?

I'd be inclined to keep things in sync.

fmenezes: Without a fallback

internal/config/profile.go
```diff
@@ -641,31 +641,42 @@ func (p *Profile) Save() error {
...
 }
 
-// MongoCLIConfigHome retrieves configHome path based used by mongoCLI.
-func MongoCLIConfigHome() (string, error) {
+// OldMongoCLIConfigHome retrieves configHome path based used by mongoCLI.
+// Deprecated: MongoCLI versions below v1.24.0 use this path.
+func OldMongoCLIConfigHome() (string, error) {
 	if home := os.Getenv("XDG_CONFIG_HOME"); home != "" {
-		return home, nil
+		return "", errors.New("not applicable")
```
fmenezes, line 648: [q] Should we really return an error here?

internal/config/profile.go
blva: no reason other than me me forgetting to remove the fallback, synced w/ you offline that is OK to remove for AtlasCLIConfigHome

internal/config/profile.go
blva: not anymore since we're now changing this as well. thanks! updating

cmd/atlas/atlas.go
```diff
@@ -99,21 +94,35 @@ func createConfigFromMongoCLIConfig() {
...
+
+	// Try finding old file since new file opening failed/did not exist
+	if configDir, err = config.OldMongoCLIConfigHome(); err == nil {
+		configPath = fmt.Sprintf("%s/mongocli.toml", configDir)
 	}
 
-	configPath := fmt.Sprintf("%s/mongocli.toml", configDir)
-	return configPath
+	// Try to open old file
+	return os.Open(configPath)
```
fmenezes, line 125: nitpick: my personal style is never to open files within functions like this rather to return the path and let the consumer open/close files, I personally believe writing code to open/close should always be together (that is one of the strengths of defer keyword).

If you are relying on open errors to fallback consider:
```go
if _, err := os.Stat("/path/to/whatever"); errors.Is(err, os.ErrNotExist) {
  // path/to/whatever does not exist
}
```

fmenezes (approved):

cmd/atlas/atlas.go
```diff
@@ -99,21 +99,35 @@ func createConfigFromMongoCLIConfig() {
...
+	}
+
+	if configDir, err = config.OldMongoCLIConfigHome(); err == nil {
+		configPath = fmt.Sprintf("%s/mongocli.toml", configDir)
+	}
+
+	if _, err = os.Stat(configPath); err == nil {
+		return configPath, nil
+	}
+	return "", err
```
gssbzn, line 130:
```suggestion
	return os.Stat(configPath)
```

cmd/atlas/atlas.go
```diff
@@ -99,21 +99,35 @@ func createConfigFromMongoCLIConfig() {
...
 	}
 
-	configPath := fmt.Sprintf("%s/mongocli.toml", configDir)
-	return configPath
+	// Check if file exists, if any error is detected try to get older file
+	if _, err = os.Stat(configPath); err == nil {
+		return configPath, nil
+	}
+
+	if configDir, err = config.OldMongoCLIConfigHome(); err == nil {
```
gssbzn, line 123:
```suggestion
	if configDir, err := config.OldMongoCLIConfigHome(); err == nil {
```

cmd/mongocli/mongocli.go
gssbzn: that looks like powershell, I may give this a try in `cmd` but let's not block the PR on this

cmd/mongocli/mongocli.go
```diff
@@ -38,6 +40,60 @@ func Execute(ctx context.Context) {
...
+	}
+
+	out, err := os.Create(mongoCLIConfigPath)
+	if err != nil {
+		return
+	}
+	defer out.Close()
+
+	_, err = io.Copy(out, in)
+	if err != nil {
```
gssbzn, line 87:
```suggestion
	if _, err = io.Copy(out, in); err != nil {
```

cmd/atlas/atlas.go
blva: this returns FileInfo object and we're returning file path here? 
```
type FileInfo interface {
	Name() string       // base name of the file
	Size() int64        // length in bytes for regular files; system-dependent for others
	Mode() FileMode     // file mode bits
	ModTime() time.Time // modification time
	IsDir() bool        // abbreviation for Mode().IsDir()
	Sys() interface{}   // underlying data source (can return nil)
}
```

gssbzn (dismissed):
LGTM thanks for all the work here, none of my comments are blocking and are mostly style

Let's not merge this until 1.23.1 is out as this is a big change for a bug fix version

cmd/atlas/atlas.go
gssbzn: true, my bad

andreaangiolillo (approved):
LGTM. Thanks for addressing all the comments! I have selected additional evergreen tasks to double-check that everything works as expected. Okay to merge this once we release 1.23.1 and all tasks are successful.

andreaangiolillo:
CM tests are failing because of one of the changes in this PR https://evergreen.mongodb.com/task_log_raw/mongocli_master_e2e_cloud_manager_remote_cloud_manager_deploy_replica_set_e2e_patch_2911a9e4aba35f76cfa2ab03053c9510c83f47d1_62264cc20305b9056d333730_22_03_07_18_19_47/0?type=T#L47

andreaangiolillo (approved):
LGTM if tests are successful. Thanks!

PR #1038, opened 2022-03-14
CLOUDP-114717: Validate Header

fmenezes:
## Proposed changes

Adds validation for the audit log header

_Jira ticket:_ CLOUDP-114717

## Checklist

### DIFF
--- a/internal/cli/opsmanager/logs/decrypt.go
+++ b/internal/cli/opsmanager/logs/decrypt.go
@@ -98,9 +98,7 @@ func DecryptBuilder() *cobra.Command {
 	cmd.Flags().StringVarP(&opts.Output, flag.Output, flag.OutputShort, "", usage.FormatOut)
 
 	_ = cmd.MarkFlagRequired(flag.File)
-
 	_ = cmd.MarkFlagFilename(flag.File)
-	_ = cmd.MarkFlagRequired(flag.File)
 	_ = cmd.MarkFlagFilename(flag.Out)
 	_ = cmd.MarkFlagFilename(flag.LocalKeyFile)
 	_ = cmd.MarkFlagFilename(flag.KMIPServerCAFile)
--- a/internal/cli/opsmanager/logs/decrypt.go
+++ b/internal/cli/opsmanager/logs/decrypt.go
@@ -79,7 +79,7 @@ func DecryptBuilder() *cobra.Command {
 		Use:   "decrypt",
 		Short: "Decrypts a log file with the provided local key file or KMIP files.",
 		Example: `
-  $ mongocli ops-manager decrypt --localKey filePath --file logPath --out resultPath`,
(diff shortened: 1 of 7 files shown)


internal/cli/opsmanager/logs/decrypt.go
```diff
@@ -100,6 +100,7 @@ func DecryptBuilder() *cobra.Command {
 	_ = cmd.MarkFlagRequired(flag.File)
 
 	_ = cmd.MarkFlagFilename(flag.File)
+	_ = cmd.MarkFlagRequired(flag.File)
```
blva, line 103: isn't this duplicated with line 100?

internal/decryption/header.go
```diff
@@ -41,17 +49,57 @@ func decodeHeader(logLine *AuditLogLine, opts KeyProviderOpts) (*HeaderRecord, e
...
+func validateHeaderFields(logLine *AuditLogLine) error {
+	if logLine.TS == nil {
+		return fmt.Errorf("missing timestamp")
+	}
+
+	if logLine.Version == nil {
+		return fmt.Errorf("missing version")
+	}
+
+	if logLine.CompressionMode == nil {
```
blva, line 74: [q] is there a list of compression modes we support? should we validate the mode as well here?

blva: also, the spec says that it will be either "zstd" or not set, does that mean it's OK to be nil?

blva (approved):
Approving since there are only non-blocker comments.

internal/cli/opsmanager/logs/decrypt.go
fmenezes: great catch

internal/decryption/header.go
fmenezes: The reason why it can be nil it is because we parse from a JSON and JSONs can miss keys. this check is to enforce presence of a value before using this field.

fmenezes: Added validation

blva (approved):

internal/cli/opsmanager/logs/decrypt.go
```diff
@@ -79,7 +79,7 @@ func DecryptBuilder() *cobra.Command {
 		Use:   "decrypt",
 		Short: "Decrypts a log file with the provided local key file or KMIP files.",
 		Example: `
-  $ mongocli ops-manager decrypt --localKey filePath --file logPath --out resultPath`,
+  $ mongocli ops-manager logs decrypt --localKeyFile filePath --file logPath --out resultPath`,
```
melissamahoney-mongodb, line 82: Might be better to make it obvious the options takes a path by actually giving a path in the example (but using the right extension, I just guessed here!):

`$ mongocli ops-manager logs decrypt --localKeyFile /path/to/keyFile.pem --file /path/to/logFile.log --out /path/to/output/logFile.log`

Minor and not blocking.

PR #1044, opened 2022-03-15
CLOUDP-112049: Support BSON files

fmenezes:
## Proposed changes

Supports BSON file to be decrypted

_Jira ticket:_ CLOUDP-112049

Closes #[issue number]

## Checklist

### DIFF
--- a/docs/mongocli/command/mongocli-cloud-manager-logs-decrypt.txt
+++ b/docs/mongocli/command/mongocli-cloud-manager-logs-decrypt.txt
@@ -83,5 +83,5 @@ Examples
 .. code-block::
 
 
-   $ mongocli ops-manager logs decrypt --localKeyFile filePath --file logPath --out resultPath
+   $ mongocli ops-manager logs decrypt --localKeyFile /path/to/file.key --file /path/to/file.bson --out /path/to/file.json
 
--- a/docs/mongocli/command/mongocli-cloud-manager-logs-decrypt.txt
+++ b/docs/mongocli/command/mongocli-cloud-manager-logs-decrypt.txt
@@ -83,5 +83,6 @@ Examples
 .. code-block::
 
 
-   $ mongocli ops-manager logs decrypt --localKeyFile /path/to/file.key --file /path/to/file.bson --out /path/to/file.json
+   $ mongocli ops-manager logs decrypt --localKeyFile /path/to/keyFile --file /path/to/logFile.bson --out /path/to/file.json
+   $ mongocli ops-manager logs decrypt --localKeyFile /path/to/keyFile --file /path/to/logFile.json --out /path/to/file.json
 
--- a/docs/mongocli/command/mongocli-ops-manager-logs-decrypt.txt
+++ b/docs/mongocli/command/mongocli-ops-manager-logs-decrypt.txt
@@ -83,5 +83,5 @@ Examples
 .. code-block::
 
 
-   $ mongocli ops-manager logs decrypt --localKeyFile filePath --file logPath --out resultPath
+   $ mongocli ops-manager logs decrypt --localKeyFile /path/to/file.key --file /path/to/file.bson --out /path/to/file.json
 
--- a/docs/mongocli/command/mongocli-ops-manager-logs-decrypt.txt
+++ b/docs/mongocli/command/mongocli-ops-manager-logs-decrypt.txt
@@ -83,5 +83,6 @@ Examples
 .. code-block::
 
 
-   $ mongocli ops-manager logs decrypt --localKeyFile /path/to/file.key --file /path/to/file.bson --out /path/to/file.json
+   $ mongocli ops-manager logs decrypt --localKeyFile /path/to/keyFile --file /path/to/logFile.bson --out /path/to/file.json
+   $ mongocli ops-manager logs decrypt --localKeyFile /path/to/keyFile --file /path/to/logFile.json --out /path/to/file.json
 
--- a/internal/cli/opsmanager/logs/decrypt_test.go
+++ b/internal/cli/opsmanager/logs/decrypt_test.go
@@ -42,11 +42,15 @@ func TestDecryptBuilder(t *testing.T) {
 
 func TestDecrypt_Run(t *testing.T) {
 	listOpts := &DecryptOpts{
-		inFileName: "test",
+		inFileName:       "test",
+		localKeyFileName: "localKey",
 	}
 	listOpts.Out = "decryptedAuditLog"
 	listOpts.Fs = afero.NewMemMapFs()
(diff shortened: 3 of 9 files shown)


internal/cli/opsmanager/logs/decrypt_test.go
```diff
@@ -42,11 +42,15 @@ func TestDecryptBuilder(t *testing.T) {
...
 func TestDecrypt_Run(t *testing.T) {
 	listOpts := &DecryptOpts{
-		inFileName: "test",
+		inFileName:       "test",
+		localKeyFileName: "localKey",
 	}
 	listOpts.Out = "decryptedAuditLog"
 	listOpts.Fs = afero.NewMemMapFs()
-	_, _ = listOpts.Fs.Create(listOpts.inFileName)
+	fileJSON := `{"ts":{"$date":{"$numberLong":"1644232049921"}},"version":"0.0","compressionMode":"zstd","keyStoreIdentifier":{"provider":"local","filename":"localKey"},"encryptedKey":{"$binary":{"base64":"+yjPCaKKE1M8fZmPGzGHkyfHYxaw34okpavsHzpd8iPVx2+JjOhXwXw5E2FdI5Rcb5JgmcPUFRPISh/7Si1R/g==","subType":"0"}},"MAC":"qE9fUsGK0EuRrrCRAQAAAAAAAAAAAAAA","auditRecordType":"header"}
```
fmenezes, line 50: This test is now testing 60% of all decrypt functionality

tibulca (approved):

internal/decryption/decryption.go
```diff
@@ -90,19 +91,86 @@ func Decrypt(logReader io.Reader, out io.Writer, opts KeyProviderOpts) error {
...
+		logLines, err = readAuditLogFileBSON(reader)
+	case JSON:
+		logLines, err = readAuditLogFileJSON(reader)
+	}
+	return auditLogFormat, logLines, err
+}
+
+func readAuditLogFileBSON(reader io.ReadSeeker) ([]*AuditLogLine, error) {
+	var logLines []*AuditLogLine
+	for {
```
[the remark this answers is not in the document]
fmenezes: the alternative would be like 
```
for raw, err := bson.NewFromIOReader(reader); err == nil || err != io.EOF {
    if err != nil {
       return nil, err
    }
    ....
}
```

I believe the break is still more readable, how about you?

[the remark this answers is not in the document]
blva: do we need to check err == nil?  this should do it, no? 
```
for raw, err := bson.NewFromIOReader(reader); err != io.EOF {
    if err != nil {
       return nil, err
    }
    ....
}
```

I don't have a strong preference, I just remember that I got a suggestion to change a for {} before. I'll leave my approval. Thanks for all the changes.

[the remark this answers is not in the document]
fmenezes: I've investigated a bit further although we don't need `err == nil || ...` we still need `raw, err := bson.NewFromIOReader(reader)` on every loop

so something like `for raw, err := bson.NewFromIOReader(reader); err != io.EOF {` would never work because `raw, err := bson.NewFromIOReader(reader)` is only initialised once.

Ideally for this to work we would need to write something like `for ; raw, err := bson.NewFromIOReader(reader); err != io.EOF; {` unfortunately that does not work either because the evaluation section contain a `;` so then it moves the `err != io.EOF` to be the incrementor section of the for instead so no luck this time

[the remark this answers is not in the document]
blva: TIL. Thanks for the details Filipe :D

internal/decryption/decryption.go
```diff
@@ -34,19 +33,19 @@ type KeyProviderOpts struct {
...
-	auditLogEncoding := newAuditLogEncoding(auditLogFormat)
 	output := buildOutput(out)
 	var decryptSection *DecryptSection
 
-	for idx, line := range logLines {
+	idx := 0
+	for ; logLineScanner.Scan(); idx++ {
 		lineNb := idx + 1
-		logLine, err := auditLogEncoding.Parse(line)
+		logLine, err := logLineScanner.AuditLogLine()
```
fmenezes, line 48: @tibulca @blva I had to move things around because I was not counting for the case when one line fails to parse and we continue to the next line

docs/mongocli/command/mongocli-cloud-manager-logs-decrypt.txt
```diff
@@ -83,5 +83,6 @@ Examples
 .. code-block::
 
 
-   $ mongocli ops-manager logs decrypt --localKeyFile filePath --file logPath --out resultPath
+   $ mongocli ops-manager logs decrypt --localKeyFile /path/to/keyFile --file /path/to/logFile.bson --out /path/to/file.json
```
colm-quinn, line 86: [q] Why the duplicated line?

docs/mongocli/command/mongocli-ops-manager-logs-decrypt.txt
```diff
@@ -83,5 +83,6 @@ Examples
 .. code-block::
 
 
-   $ mongocli ops-manager logs decrypt --localKeyFile filePath --file logPath --out resultPath
+   $ mongocli ops-manager logs decrypt --localKeyFile /path/to/keyFile --file /path/to/logFile.bson --out /path/to/file.json
```
colm-quinn, line 86: Same as above

docs/mongocli/command/mongocli-cloud-manager-logs-decrypt.txt
fmenezes: Not duplicated, different example json and bson log files

docs/mongocli/command/mongocli-ops-manager-logs-decrypt.txt
fmenezes: Not same, see previous comment

colm-quinn (approved):
LGTM!

internal/decryption/audit_log_line_scanner.go
```diff
@@ -0,0 +1,137 @@
...
+}
+
+func (s *bsonScanner) Scan() bool {
+	raw, err := bson.NewFromIOReader(s.r)
+	if err != nil {
+		if err == io.EOF {
+			return false
+		}
+		s.err = err
+		return false
```
tibulca, line 93: in both cases we want to return `false`? If a BSON document that is in the middle of the file is corrupted, can bson.NewFromIOReader skip it and move to next document?

If in both cases we should return false then this can be written like this:
```
if err != io.EOF {
	s.err = err
}
return false
```

tibulca (approved):

PR #1055, opened 2022-03-18
CLOUDP-115398: [Atlas CLI] Improve CLI login experience

andreaangiolillo:
## Proposed changes

_Jira ticket:_ CLOUDP-115398

## Checklist

## Further comments

In this PR we're including the following improvements to the login experience:

- Showing the message "you have already set the programmatic keys for this profile, run `atlas auth login --profile <profile_name>` to use your username and password on a new profile" when the user is trying to use auth login on a profile which has programmatic keys
- Enabling the flag `--loginOnly   Skip profile configuration.`

```
./bin/atlas auth login --help
Authenticate with MongoDB Atlas.

Usage:
  atlas auth login [flags]

Examples:
  To start the interactive setup:
  $ mongocli auth login


Flags:
      --gov         Log in to Atlas for Government.
  -h, --help        help for login
      --loginOnly   Skip profile configuration.
      --noBrowser   Don't try to open a browser session.

Global Flags:
  -P, --profile string   Profile to use from your configuration file.
```

```
/bin/atlas auth login --loginOnly
Error: you have already set the programmatic keys for this profile.

Run 'atlas auth login --profile <profile_name>' to use your username and password on a new profile
```


internal/cli/auth/login.go
```diff
@@ -88,6 +88,16 @@ func (opts *loginOpts) SetOAuthUpAccess() {
 }
 
 func (opts *loginOpts) Run(ctx context.Context) error {
+	// Check if user has already logged in
+	if s, err := opts.config.AccessTokenSubject(); err == nil && s != "" {
```
gssbzn, line 92: what will happen when the token expires?

andreaangiolillo: users will have to first run `auth logout` and then run `auth login` as I think they are doing right now

gssbzn (changes requested):
checking the token has a subject is not a good check, tokens expire which means users need to re-authenticate, you can check if the token has expired, we have logic for that but then we have a refresh token which sadly lacks expiration embedded and you can only verify it expired by trying to use it 
TL;DR preventing users to authenticate based on there being a token is no bueno and may lock out users with expired tokens

internal/cli/auth/login.go
andreaangiolillo: I see how it makes sense to check if the token is valid

gssbzn: right now I don't have to login, I just re-login

gssbzn: my recommendation to discuss with @dianchenghu and @boooczek is to ask users if they want to reauthenticate,
```bash
You are already authenticated as <>: do you want to re authenticate(Y/n): _
```

boooczek: I understand what we're discussing here is case when token expieres. Then when user types in "atlas login" the flow is as usual, i.e. they go through auth process in the browser.
@gssbzn why users would want to re-authenticate?

gssbzn: > why users would want to re-authenticate? Is it because their token might expire soon?
- I may have multiple atlas account (I do, personal and professional) and I may want to switch between them
- See my general comment on token expiration:  tokens expire which means users need to re-authenticate, you can check if the token has expired, we have logic for that but then we have a refresh token which sadly lacks expiration embedded and you can only verify it expired by trying to use it, there are two tokens one we know when it expires the second one we don't until we try to use it, we can't change how that second token works

internal/cli/auth/login.go
```diff
@@ -188,11 +210,13 @@ func LoginBuilder() *cobra.Command {
...
 	}
 
+	if config.ToolName == config.MongoCLI {
+		cmd.Flags().BoolVar(&opts.isCloudManager, "cm", false, "Log in to Cloud Manager.")
+	}
+
 	cmd.Flags().BoolVar(&opts.isGov, "gov", false, "Log in to Atlas for Government.")
-	cmd.Flags().BoolVar(&opts.isCloudManager, "cm", false, "Log in to Cloud Manager.")
 	cmd.Flags().BoolVar(&opts.noBrowser, "noBrowser", false, "Don't try to open a browser session.")
 	cmd.Flags().BoolVar(&opts.loginOnly, "loginOnly", false, "Skip profile configuration.")
```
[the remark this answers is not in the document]
andreaangiolillo: Leaving this comment open so that doc team can decide how to call this flag

internal/cli/root/atlas/builder.go
```diff
@@ -88,6 +89,14 @@ func Builder(profile *string) *cobra.Command {
 			"toc": "true",
 		},
 		PersistentPreRunE: func(cmd *cobra.Command, args []string) error {
+			if err := opts.InitFlow(); err != nil {
+				return err
+			}
+
+			if err := opts.RefreshAccessToken(cmd.Context()); err != nil {
+				return err
+			}
```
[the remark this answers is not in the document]
andreaangiolillo: makes sense. Updated

andreaangiolillo:
evergreen retry

internal/cli/root/atlas/builder.go
```diff
@@ -93,6 +94,13 @@ func Builder(profile *string) *cobra.Command {
...
 
 			if shouldCheckCredentials(cmd) {
+				if err := opts.InitFlow(); err != nil {
+					return err
+				}
+
+				if err := opts.RefreshAccessToken(cmd.Context()); err != nil {
+					return err
+				}
 				return validate.Credentials()
```
tibulca, line 104: not a strong opinion here (since I've seen both options in the past), but if refreshAccessToken fails do we want to fail the flow? Maybe the existing credentials are still valid. And if the credentials are not valid anymore the flow will be interrupted at by line `return validate.Credentials()`

tibulca: ignore this, checked the RefreshAccessMethod and if the token is valid it doesn't try to refresh it (I thought it will try to refresh it if x% of it's validity passed)
```
	if current.Valid() {
		return nil
	}
```

tibulca (approved):

internal/cli/auth/login.go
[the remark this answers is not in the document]
sarahsimpers: I do think skipConfig is a better name for it.

sarahsimpers (dismissed):
LGTM % comment on flag name, once this gets merged I will change the Atlas CLI docs to reflect this as the quickest getting started method. Thanks!

johansteyn (approved):
LGTM

tibulca (approved):

gssbzn (approved):
LGTM

PR #1061, opened 2022-03-21
CLOUDP-112093: Stdout and Error format

fmenezes:
## Proposed changes

This is to change error from plain string to an actual json format and also supports stdout when `--out` is omitted.

_Jira ticket:_ CLOUDP-112093

Closes #[issue number]

## Checklist

### DIFF
--- a/internal/cli/opsmanager/logs/decrypt.go
+++ b/internal/cli/opsmanager/logs/decrypt.go
@@ -16,6 +16,8 @@ package logs
 
 import (
 	"fmt"
+	"io"
+	"os"
 
 	"github.com/mongodb/mongocli/internal/cli"
 	"github.com/mongodb/mongocli/internal/decryption"
@@ -43,11 +45,17 @@ func (opts *DecryptOpts) initFiles() func() error {
 }
 
 func (opts *DecryptOpts) Run() error {
-	outWriter, err := opts.NewWriteCloser()
-	if err != nil {
-		return err
+	var outWriter io.WriteCloser
+	if len(opts.Out) > 0 {
+		var err error
+		outWriter, err = opts.NewWriteCloser()
+		if err != nil {
+			return err
+		}
+		defer outWriter.Close()
+	} else {
+		outWriter = os.Stdout
 	}
-	defer outWriter.Close()
 
 	inReader, err := opts.Fs.Open(opts.inFileName)
 	if err != nil {
@@ -62,11 +70,13 @@ func (opts *DecryptOpts) Run() error {
 	}
 
 	if err := decryption.Decrypt(inReader, outWriter, keyProviderOpts); err != nil {
-		_ = opts.OnError(outWriter)
-		return err
+		if len(opts.Out) > 0 {
+			_ = opts.OnError(outWriter)
+			return err
+		}
 	}
 
-	if opts.Out != "/dev/stdout" {
+	if len(opts.Out) > 0 {
 		fmt.Printf("Decrypt of %s to %s completed.\n", opts.inFileName, opts.Out)
 	}
 
--- a/internal/decryption/output.go
+++ b/internal/decryption/output.go
@@ -21,39 +21,87 @@ import (
 	"go.mongodb.org/mongo-driver/bson"
 )
 
-type AuditLogOutput struct {
-	Warningf  func(lineNb int, format string, a ...interface{}) error
-	Error     func(lineNb int, err error) error
-	Errorf    func(lineNb int, format string, a ...interface{}) error
-	LogRecord func(lineNb int, logRecord interface{}) error
+type AuditLogOutput interface {
+	Warningf(lineNb int, format string, a ...interface{}) error
+	Error(lineNb int, err error) error
+	Errorf(lineNb int, format string, a ...interface{}) error
+	LogRecord(lineNb int, logRecord interface{}) error
 }
 
-func buildOutput(out io.Writer) AuditLogOutput {
-	newLine := []byte{'\n'}
-	writeLine := func(value []byte) error {
-		if _, err := out.Write(value); err != nil {
-			return err
-		}
-		_, err := out.Write(newLine)
+type auditLogOutputImpl struct {
+	out io.Writer
+}
+
+type AuditLogErrorLevel string
+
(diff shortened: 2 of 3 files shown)


internal/cli/opsmanager/logs/decrypt.go
```diff
@@ -42,12 +44,23 @@ func (opts *DecryptOpts) initFiles() func() error {
 	}
 }
 
+func (opts *DecryptOpts) stdOutMode() bool {
+	return opts.Out == ""
```
tibulca, line 48:
```suggestion
	return opts.Out == "" || opts.Out == "/dev/stdout"
```
but this depends on the discussion about "/dev/stdout" (see below)

internal/cli/opsmanager/logs/decrypt.go
```diff
@@ -42,12 +44,23 @@ func (opts *DecryptOpts) initFiles() func() error {
...
+}
+
 func (opts *DecryptOpts) Run() error {
-	outWriter, err := opts.NewWriteCloser()
-	if err != nil {
-		return err
+	var outWriter io.WriteCloser
+
+	if opts.stdOutMode() {
```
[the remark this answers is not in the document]
fmenezes: I think `/dev/stdout` is not cross platform, have we tested it on windows?

internal/decryption/output.go
```diff
@@ -21,39 +21,87 @@ import (
...
+	AuditLogErrorLevelWarning AuditLogErrorLevel = "decryptionLogWarning"
+)
+
+type AuditLogError struct {
+	Level AuditLogErrorLevel
+	Line  int
+	Err   error
+}
+
+func (e AuditLogError) MarshalBSON() ([]byte, error) {
```
tibulca, line 48: is this used?

internal/decryption/output.go
```diff
@@ -21,39 +21,87 @@ import (
...
-		}
-		_, err := out.Write(newLine)
+type AuditLogOutput interface {
+	Warningf(lineNb int, format string, a ...interface{}) error
+	Error(lineNb int, err error) error
+	Errorf(lineNb int, format string, a ...interface{}) error
+	LogRecord(lineNb int, logRecord interface{}) error
+}
+
+type auditLogOutputImpl struct {
```
gssbzn, line 31: [nit] not very go idiomatic to have `Impl`, prefer to give them names related to what they do than by what they are, `AuditLogOutput` could be `AuditLogPrinter`  and `auditLogOutputImpl` an `auditLogIOWriter`(?)

Also do we really need this interface, prefer interfaces as inputs for functions and concretes as outputs and also why is a good idea to define the interface close to the function that takes the expected contract

internal/decryption/output.go
fmenezes: yes by the bson marshaler

internal/decryption/output.go
```diff
@@ -21,39 +21,87 @@ import (
...
+		bson.E{Key: "atype", Value: string(e.Level)},
+		bson.E{Key: "line", Value: e.Line},
+		bson.E{Key: "error", Value: e.Err.Error()},
+	}
+	return bson.Marshal(doc)
+}
+
+func (e AuditLogError) Error() string {
+	return e.Err.Error()
+}
```
gssbzn, line 59: [q] should we implement `func (e *AuditLogError) Is(target error) bool {`?

internal/decryption/output.go
fmenezes: https://github.com/mongodb/mongodb-atlas-cli/pull/1061/files#diff-f92e3e928b3bb2039c0c5949652bbfcff0a03b58998de03f746dee968b709da2R68

internal/cli/opsmanager/logs/decrypt.go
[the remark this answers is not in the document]
tibulca: in the os package I see Stdout used one line below defined like this: `Stdout = NewFile(uintptr(syscall.Stdout), "/dev/stdout")`
so it might be further in the framework where /dev/stdout is mapped to the correct os stdout.
Will try in a windows VM

internal/decryption/output.go
```diff
@@ -21,39 +21,87 @@ import (
...
+
+func NewAuditLogOutput(out io.Writer) AuditLogOutput {
+	return &auditLogOutputImpl{
+		out: out,
+	}
+}
```
gssbzn, line 65: [future improvement] `auditLogOutputImpl` has potential to grow in customisations, consider making this `new` functional 
```suggestion
func NewAuditLogOutput(opts ...func(AuditLogOutput)) AuditLogOutput {
	a := &auditLogOutputImpl{
		out: out,
	}
	for _, o := range opts {
		o(a)
	}
	return a	
}
```
see https://github.com/mongodb/go-client-mongodb-atlas/blob/116acf986f4738d6a54e4ebaca13071ac3f86326/mongodbatlas/mongodbatlas.go#L307-L316

gssbzn (approved):
LGTM

internal/cli/opsmanager/logs/decrypt.go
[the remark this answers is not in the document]
tibulca: in windows, for os.OpenFile("/dev/stdout",...) I got "open /dev/stdout: The system cannot find the path specified." 
while os.Stdout seems to work.
I guess we can leave the current implementation here and revisit the usage of /dev/stdout in the other places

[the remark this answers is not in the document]
fmenezes: @tibulca would you open a Jira for a follow up? If we have code that is broken on Windows we have to fix it.

internal/decryption/output.go
fmenezes: I don't think so according to https://pkg.go.dev/errors#Is we should only do it if we want to treat other know errors (like `os.FileNotExists`) the same as our custom error. Which is not the case.

internal/decryption/output.go
fmenezes: it is actually an interface https://pkg.go.dev/go.mongodb.org/mongo-driver/bson#Marshaler

PR #1071, opened 2022-03-22
CLOUDP-114862: [Atlas CLI / MongoCLI] Improve error handling when user is unathenticated

andreaangiolillo:
## Proposed changes

_Jira ticket:_ CLOUDP-114862

## Checklist

- [x] I have signed the [MongoDB CLA](https://www.mongodb.com/legal/contributor-agreement)
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] I have added any necessary documentation in document requirements section listed in [CONTRIBUTING.md](https://github.com/mongodb/mongocli/blob/master/CONTRIBUTING.md) (if appropriate)
- [x] I have addressed the @mongodb/docs-cloud-team comments (if appropriate)
- [x] I have updated [e2e/E2E-TESTS.md](https://github.com/mongodb/mongocli/blob/master/e2e/E2E-TESTS.md) (if an e2e test has been added)
- [x] I have run `make fmt` and formatted my code

## Further comments

This PR updates the missing credential error message to include `auth login`

**Before**: 
```
./bin/mongocli atlas cluster ls
Error: missing credentials

To set credentials, run: mongocli config
```

**After**:
```
./bin/atlas cluster ls
Error: this actions requires authentication

To login using your Atlas username and password, run: atlas auth login
To set credentials using API keys, run: atlas config init
```

```
./bin/mongocli atlas cluster ls
Error: this actions requires authentication

To login using your Atlas username and password, run: mongocli auth login
To set credentials using API keys, run: mongocli config
```

PR #1074, opened 2022-03-23
CLOUDP-116473: [CLI] Fix HomeBrew Test

andreaangiolillo:
## Proposed changes

_Jira ticket:_ CLOUDP-116473

## Checklist

## Further comments

[brew test is failing](https://evergreen.mongodb.com/task/mongocli_master_e2e_generic_brew_e2e_39b500c735a3545e77b18d3e410784f0a8dd631e_22_03_23_09_18_47) after merging https://github.com/mongodb/mongodb-atlas-cli/pull/1071.

**Proposed Changes:**

* fix the e2e test
* add the atlas cli brew test
* move the e2e test to the code health variant

**We will need to update the brew formula of both CLIs in their next release.**

### DIFF
--- a/build/ci/evergreen.yml
+++ b/build/ci/evergreen.yml
@@ -334,18 +334,6 @@ tasks:
     commands:
       - func: "clone"
       - func: "check license header"
-  - name: brew_e2e
-    tags: [ "code_health" ]
-    depends_on:
-      - name: compile
-        variant: "code_health"
-        patch_optional: true
-    commands:
-      - func: "clone"
-      - func: "install gotestsum"
-      - func: "e2e test"
-        vars:
-          E2E_TAGS: brew
   - name: config_e2e
     tags: ["e2e","generic"]
     depends_on:
@@ -376,6 +364,18 @@ tasks:
           MCLI_OPS_MANAGER_URL: ${mcli_ops_manager_url}
           MCLI_SERVICE: cloud
           E2E_TAGS: atlas,config
+  - name: brew_e2e
+    tags: ["e2e","generic"]
+    depends_on:
+      - name: compile
+        variant: "code_health"
+        patch_optional: true
+    commands:
+      - func: "clone"
+      - func: "install gotestsum"
+      - func: "e2e test"
+        vars:
+          E2E_TAGS: brew
   # If your e2e tests depend on a cluster running please consider setting it on its own task
   - name: atlas_generic_e2e
     tags: ["e2e","generic","atlas"]
--- a/build/ci/evergreen.yml
+++ b/build/ci/evergreen.yml
@@ -334,6 +334,18 @@ tasks:
     commands:
       - func: "clone"
       - func: "check license header"
+  - name: brew_e2e
+    tags: [ "code_health" ]
+    depends_on:
+      - name: compile
+        variant: "code_health"
+        patch_optional: true
+    commands:
+      - func: "clone"
+      - func: "install gotestsum"
+      - func: "e2e test"
+        vars:
+          E2E_TAGS: brew
   - name: config_e2e
     tags: ["e2e","generic"]
     depends_on:
@@ -364,18 +376,6 @@ tasks:
           MCLI_OPS_MANAGER_URL: ${mcli_ops_manager_url}
           MCLI_SERVICE: cloud
           E2E_TAGS: atlas,config
-  - name: brew_e2e
-    tags: ["e2e","generic"]
-    depends_on:
-      - name: compile
-        variant: "code_health"
-        patch_optional: true
-    commands:
-      - func: "clone"
-      - func: "install gotestsum"
-      - func: "e2e test"
-        vars:
-          E2E_TAGS: brew
   # If your e2e tests depend on a cluster running please consider setting it on its own task
   - name: atlas_generic_e2e
     tags: ["e2e","generic","atlas"]
--- a/internal/cli/root/atlas/builder.go
+++ b/internal/cli/root/atlas/builder.go
@@ -17,6 +17,7 @@ package atlas
 import (
 	"fmt"
 	"io"
+	"os"
 	"runtime"
 	"strings"
 	"time"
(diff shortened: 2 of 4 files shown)


andreaangiolillo:
evergreen retry

fmenezes:
@andreaangiolillo we will probably need to update homebrew when bumping next time https://github.com/Homebrew/linuxbrew-core/blob/251e0f56f238fe453329445841fc8e7b1af357f0/Formula/mongocli.rb#L34

build/ci/evergreen.yml
```diff
@@ -334,6 +334,18 @@ tasks:
...
+    depends_on:
+      - name: compile
+        variant: "code_health"
+        patch_optional: true
+    commands:
+      - func: "clone"
+      - func: "install gotestsum"
+      - func: "e2e test"
+        vars:
+          E2E_TAGS: brew
```
gssbzn, line 348: [nit] we should not move e2e to be part of the code health, if your end goal is to have e2e be mandatory that's a different conversation

internal/cli/root/atlas/builder.go
```diff
@@ -88,10 +88,6 @@ func Builder(profile *string) *cobra.Command {
 			"toc": "true",
 		},
 		PersistentPreRunE: func(cmd *cobra.Command, args []string) error {
-			if config.Service() == "" {
```
gssbzn, line 91: can you explain a bit more about this removal?

andreaangiolillo: When running `atlas config ls` without any configuration, the command was listing the default profile with `service = cloud`. it took me a while to understand why and the reason was this piece of code that was copied/pasted from the atlas command.

build/ci/evergreen.yml
andreaangiolillo: yes, the main reason was to make this test mandatory

gssbzn: let's rollback this and discuss with the team making e2e mandatory

internal/cli/root/atlas/builder.go
colm-quinn: [nit/idea] Regardless of the decision, should we output to stdout that we're assuming cloud if the value was not set? Would improve the UX and be make the flow more predictable.

The tradeoff is being more verbose/maybe users don't care.

build/ci/evergreen.yml
colm-quinn: As part of the discussion, mandatory for a code change vs mandatory for a release (only required in the PCT runs)

internal/cli/root/atlas/builder.go
gssbzn: this was kind of requested, we didn't use to assume and force people to do `MONGODB_ATLAS_SERVICE=cloud` and people hated it and asked if they were using an atlas command why they need to specify the service since  is a given by the name of the command 🤷

andreaangiolillo: > I expect that to be successful without having to explicitly set the service, and the reason behind setting the service here

@gssbzn why? given that the cli can be used with both atlas and atlas for gov, this expectation doesn’t seem correct

> [nit/idea] Regardless of the decision, should we output to stdout that we're assuming cloud if the value was not set? Would improve the UX and be make the flow more predictable
yes, I agree that if we decide to automatically assign the service, we should print a warning message

build/ci/evergreen.yml
andreaangiolillo: I have removed this change from the PR. I will open a discussion in the slack channel. I agree that we may just need to make this test mandatory for a release.

internal/cli/root/atlas/builder.go
andreaangiolillo: Sorry I haven’t seen your response in the thread.

andreaangiolillo: I have updated my changes. Now I am not setting the service only when running the config and completion commands. I am also printing a message when we set the service.


```
./bin/atlas project list
service not set. Atlas CLI used 'cloud'

Error: this action requires authentication

```

```
./bin/mongocli atlas cluster ls
service not set. MongoCLI used 'cloud'

Error: this action requires authentication

To log in using your Atlas username and password, run: mongocli auth login
To set credentials using API keys, run: mongocli config
```

internal/cli/root/atlas/builder.go
```diff
@@ -88,7 +88,8 @@ func Builder(profile *string) *cobra.Command {
...
+			if shouldSetService(cmd) {
+				_, _ = fmt.Printf("service not set. Atlas CLI used '%s'\n\n", config.CloudService)
```
gssbzn, line 92: this is a breaking change, tools that parse the output are going to break

andreaangiolillo: I don't see how to prevent this if we want to print a warning message 😕

internal/cli/root/atlas/builder.go
gssbzn: @andreaangiolillo this request predates atlas gov, also we are promising people a smooth transition from mongocli to atlas cli, not fulfilling this expectation is a breaking change.

Message should be review by product, `cloud` is a confusing terms from the time when originally mogndob cli would also include things like realm and not atlas only so people may not understand what we mean by `cloud` here

Also notice this warning message should go to stderr otherwise it's a breaking change and you'll break any tool parsing the output

internal/cli/root/atlas/builder.go
gssbzn: setting the default to cloud for atlas commands it's been there even before `atlas cli` no one has been surprised, I don't believe this warning adds any value but if product wants it then use sdterr instead of stdout

andreaangiolillo: @boooczek We need your input to decide if we want to print a message to stderr if the user did not provide the service as we're automatically setting the service = cloud. As Gustavo said, we have been always automatically setting service = cloud when service is not set without printing any message. If you agree to print this message, I also need your input to decide what we should show to the user. The current message is 

> service not set. MongoCLI used 'cloud'

but `cloud` may be confusing.

andreaangiolillo: Discussed with the product manager and they decided to not show the message as we have been always doing. @colm-quinn

gssbzn (approved):
LGTM

internal/cli/root/atlas/builder.go
colm-quinn: sgtm, thanks for the tag!

PR #1112, opened 2022-04-05
CLOUDP-118221: [CLI] We should not parse the access token when programmatic keys are available

andreaangiolillo:
## Proposed changes

_Jira ticket:_ CLOUDP-118221

## Checklist

## Further comments

This PR updates `WithAuthentication` to not parse the access token when programmatic keys are available

### DIFF
--- a/internal/store/store.go
+++ b/internal/store/store.go
@@ -219,11 +219,14 @@ func WithAuthentication(c CredentialsGetter) Option {
 	return func(s *Store) error {
 		s.username = c.PublicAPIKey()
 		s.password = c.PrivateAPIKey()
-		t, err := c.Token()
-		if err != nil {
-			return err
+
+		if s.username == "" && s.password == "" {
+			t, err := c.Token()
+			if err != nil {
+				return err
+			}
+			s.accessToken = t
 		}
-		s.accessToken = t
 		return nil
 	}
 }


internal/store/store.go
```diff
@@ -219,11 +219,14 @@ func WithAuthentication(c CredentialsGetter) Option {
...
-		t, err := c.Token()
-		if err != nil {
-			return err
+
+		if s.username == "" && s.password == "" {
+			t, err := c.Token()
+			if err != nil {
+				return err
+			}
+			s.accessToken = t
```
andreaangiolillo, line 228: we use the access token only when programmatic keys are not set

andreaangiolillo:
Moving this to WIP: I have discussed this proposal with @johansteyn and it may make sense to keep the access token even if the user is using the programmatic keys for telemetry. Waiting to hear back from the product team to see what to do here

fmenezes (dismissed):
Go code LGTM

internal/store/store.go
andreaangiolillo: yes, correct. I am just waiting to see if I should remove the logic that remove the auth tokens in the config commands

internal/cli/atlas/config/init.go
```diff
@@ -97,6 +99,20 @@ Enter [?] on any option to get help.
...
+func deleteAccessAndRefreshTokenFromProfile() {
+	accessToken := config.GetString(config.AccessTokenField)
+	refreshToken := config.GetString(config.RefreshTokenField)
+
+	if accessToken == "" && refreshToken == "" {
+		return
+	}
+
+	config.Set(config.AccessTokenField, "")
+	config.Set(config.RefreshTokenField, "")
```
gssbzn, line 111: it'd probably be better to implement a `Unset`

internal/store/store.go
andreaangiolillo: This is not the behavior that I see though. If I have an invalid access token and the programmatic key I get this error without this change. I believe `t, err := c.Token()` may the root cause here

```
./bin/atlas project ls -P default-access-token
Error: token contains an invalid number of segments
```

gssbzn: well that may be your own fault, `invalid number of segments` means you manually modified that token and the only case I would expect `c.Token()` to error, if you give it a random string and not a valid JWT, and by valid in the generic is a JWT token kind of way, you can verify your string in this page https://jwt.io/, if that page can parse then this function can as well

andreaangiolillo: I see. I don't think we should parse the access token if we are not using it though

gssbzn: and I don't think users should be manually editing their tokens, in general my comment in this particular change is not blocking, just saying there's not much gain but it doesn't hurt, deleting things without previous warning does hurt and a lot

gssbzn (approved):
LGTM

PR #1137, opened 2022-04-20
CLOUDP-118467: Remove mongosh questions

blva:
## Proposed changes

_Jira ticket:_ [CLOUDP-118467](https://jira.mongodb.org/browse/CLOUDP-118467)

## atlas config init
```shell
➜  mongocli git:(CLOUDP-118467) atlasdev config init
You are configuring a profile for atlas.

All values are optional and you can use environment variables (MONGODB_ATLAS_*) instead.

Enter [?] on any option to get help.

? Public API Key:
? Private API Key: [? for help]
? Default Project ID: 61dc598fd1f07c7b37ba425d
? Default Org ID: 5e429e7706822c6eac4d5970
? Default Output Format:

You can use [atlas config set] to change these settings at a later time.

```
## atlas auth login
```shell
➜  mongocli git:(master) ✗ atlasdev login

First, copy your one-time code: FBQJ-WTQM

Next, sign in with your browser and enter the code.

Or go to https://account-dev.mongodb.com/account/connect

Your code will expire after 10 minutes.
Successfully logged in as bianca.vianadeaguiar@mongodb.com.
Press Enter to continue your profile configuration
You have access to more than 500 organizations
? Do you want to enter the Org ID manually? No
Skipping default organization setting
You have access to more than 500 projects
? Do you want to enter the Project ID manually? No
Skipping default project setting
? Default Output Format:

You can use [atlas config set] to change these settings at a later time.
```

## Checklist

### DIFF
--- a/internal/cli/default_setter_opts.go
+++ b/internal/cli/default_setter_opts.go
@@ -226,14 +226,16 @@ func (opts *DefaultSetterOpts) SetUpOrg() {
 }
 
 func (opts *DefaultSetterOpts) SetUpMongoSHPath() {
-	if opts.IsCloud() {
-		defaultPath := config.MongoShellPath()
-		if defaultPath == "" {
-			defaultPath = mongosh.Path()
-		}
-
-		config.SetMongoShellPath(defaultPath)
+	if !opts.IsCloud() {
+		return
 	}
+
+	defaultPath := config.MongoShellPath()
+	if defaultPath == "" {
+		defaultPath = mongosh.Path()
+	}
+
+	config.SetMongoShellPath(defaultPath)
 }
 
 func (opts *DefaultSetterOpts) SetUpOutput() {
--- a/internal/cli/default_setter_opts.go
+++ b/internal/cli/default_setter_opts.go
@@ -226,8 +226,13 @@ func (opts *DefaultSetterOpts) SetUpOrg() {
 }
 
 func (opts *DefaultSetterOpts) SetUpMongoSHPath() {
-	if opts.MongoShellPath != "" {
-		config.SetMongoShellPath(opts.MongoShellPath)
+	if opts.IsCloud() {
+		defaultPath := config.MongoShellPath()
+		if defaultPath == "" {
+			defaultPath = mongosh.Path()
+		}
+
+		config.SetMongoShellPath(defaultPath)
 	}
 }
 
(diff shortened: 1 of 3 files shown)


internal/cli/default_setter_opts.go
```diff
@@ -226,8 +226,13 @@ func (opts *DefaultSetterOpts) SetUpOrg() {
 }
 
 func (opts *DefaultSetterOpts) SetUpMongoSHPath() {
-	if opts.MongoShellPath != "" {
-		config.SetMongoShellPath(opts.MongoShellPath)
+	if opts.IsCloud() {
+		defaultPath := config.MongoShellPath()
+		if defaultPath == "" {
+			defaultPath = mongosh.Path()
```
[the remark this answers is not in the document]
blva: mongosh.Path will return an empty string if it doesn't find it
```go
func Path() string {
	if path, err := exec.LookPath(Bin()); err == nil {
		return path
	}

	return ""
}
```

internal/cli/default_setter_opts.go
```diff
@@ -274,21 +279,5 @@ func (opts *DefaultSetterOpts) DefaultQuestions() []*survey.Question {
...
-			Name: "mongoShellPath",
-			Prompt: &survey.Input{
-				Message: "Default MongoDB Shell Path:",
-				Help:    "MongoDB CLI will use the MongoDB shell version provided to allow you to access your deployments.",
-				Default: defaultPath,
-			},
-			Validate: validate.OptionalPath,
-		}
-		q = append(q, atlasQuestion)
-	}
```
colm-quinn, line 292: Does this mean we now never prompt for the path? What if it wasn't detected in the standard path, but it is installed for instance?

I'm wondering does this put us in a situation where if we can't detect the path, the user can't enter it.

blva: * `Does this mean we now never prompt for the path?` Yes
* `What if it wasn't detected in the standard path, but it is installed for instance?` It would be updated during login/atlas config commands as we call `SetUpMongoSHPath()` as part of these commands
* `I'm wondering does this put us in a situation where if we can't detect the path, the user can't enter it.` that is a great point, we would be relying on finding the path or the customer can edit the configuration file directly. One thing I can suggest would be to ask the path if it's not detected for one of the commands (maybe config init?) but still it would get automatically updated if we find a new path...

blva: One more thing I missed to explain is that for quickstart we ask 1) `"Do you want to provide the path to your MongoDB Shell binary?"` and if so, then we offer them the option to add the mongo path, I did not change this part as it is hidden under another question.

colm-quinn: >"Do you want to provide the path to your MongoDB Shell binary?"

Ah ok, at least they are getting prompted to set the path. 
[idea] How about we show the detected path if they say no?

blva: sure! we can do that, thank you

blva: actually, we only ask this if the MongoDB shell path is not present. In that sense, I think if they say no there is no way for us to say a path because we already checked config and searched for a path. wdyt?

colm-quinn: Yes, the path for yes allows a path to be entered.
The path for no sounds slightly more subtle. In that case, since we don't have a shell, we probably finish creating the cluster as usual but not try to call the shell to connect to the cluster, instead displaying the connection string.

colm-quinn (approved):
LGTM, thanks for those changes and discussion!

PR #1138, opened 2022-04-21
CLOUDP-120225: [AtlasCLI] we should not show credential error when running "atlas login"

andreaangiolillo:
## Proposed changes

_Jira ticket:_ CLOUDP-120225

## Checklist

### DIFF
--- a/internal/cli/root/atlas/builder.go
+++ b/internal/cli/root/atlas/builder.go
@@ -218,6 +218,7 @@ func shouldCheckCredentials(cmd *cobra.Command) bool {
 		fmt.Sprintf("%s %s", atlas, "completion"), // completion commands do not require credentials
 		fmt.Sprintf("%s %s", atlas, "config"),     // user wants to set credentials
 		fmt.Sprintf("%s %s", atlas, "auth"),       // user wants to set credentials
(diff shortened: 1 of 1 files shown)


internal/cli/root/atlas/builder.go
```diff
@@ -218,6 +218,7 @@ func shouldCheckCredentials(cmd *cobra.Command) bool {
 		fmt.Sprintf("%s %s", atlas, "completion"), // completion commands do not require credentials
 		fmt.Sprintf("%s %s", atlas, "config"),     // user wants to set credentials
 		fmt.Sprintf("%s %s", atlas, "auth"),       // user wants to set credentials
+		fmt.Sprintf("%s %s", atlas, "login"),      // user wants to set credentials
```
gssbzn, line 221: thanks for the fix can you check if mongocli has the same problem please?

gssbzn: also I think you should be able to do this in the search by name since `"login"` is a terminal command

andreaangiolillo: no, it doesn’t. We are not validating the credentials at the build level but only in the atlas command

gssbzn (approved):
LGTM

PR #1163, opened 2022-05-03
CLOUDP-120449: Telemetry - Track Command Duration and Result

fmenezes:
## Proposed changes

Track duration and result in CLI

_Jira ticket:_ CLOUDP-120449

Closes #[issue number]

## Checklist

### DIFF
--- a/internal/telemetry/telemetry.go
+++ b/internal/telemetry/telemetry.go
@@ -69,7 +69,7 @@ func TrackCommand(cmd *cobra.Command) {
 	track(cmd)
 }
 
-func generateEventData(cmd *cobra.Command) Event {
+func newEvent(cmd *cobra.Command) Event {
 	now := time.Now()
 	cmdPath := cmd.CommandPath()
 	command := strings.ReplaceAll(cmdPath, " ", "-")
@@ -96,7 +96,7 @@ func generateEventData(cmd *cobra.Command) Event {
 }
 
 func track(cmd *cobra.Command) {
-	event := generateEventData(cmd)
+	event := newEvent(cmd)
 
 	cacheDir, err := os.UserCacheDir()
 	if err != nil {
--- a/internal/telemetry/telemetry.go
+++ b/internal/telemetry/telemetry.go
@@ -78,6 +78,8 @@ func newEvent(cmd *cobra.Command) Event {
 	var duration time.Duration
 	if found {
 		duration = now.Sub(ctxValue.startTime)
+	} else {
+		logError(errors.New("telemetry context not found"))
 	}
 
 	var properties = map[string]interface{}{
--- a/internal/telemetry/telemetry.go
+++ b/internal/telemetry/telemetry.go
@@ -15,6 +15,7 @@
 package telemetry
 
 import (
+	"context"
 	"encoding/json"
 	"errors"
 	"os"
@@ -35,12 +36,33 @@ const (
 
 var fs = afero.NewOsFs()
 var maxCacheFileSize int64 = 100_000_000 // 100MB
+var contextKey = telemetryContextKey{}
 
 type Event struct {
 	Timestamp  string                 `json:"timestamp"`
 	Source     string                 `json:"source"`
 	Name       string                 `json:"name"`
 	Properties map[string]interface{} `json:"properties"`
+}
+
+type telemetryContextKey struct{}
+
+type telemetryContextValue struct {
+	startTime time.Time
+}
+
+func NewContext() context.Context {
+	return context.WithValue(context.Background(), contextKey, telemetryContextValue{
+		startTime: time.Now(),
+	})
+}
+
+func getValueFromContext(ctx context.Context) *telemetryContextValue {
+	value, ok := ctx.Value(contextKey).(telemetryContextValue)
+	if !ok {
+		return nil
+	}
+	return &value
 }
 
 func TrackCommand(cmd *cobra.Command) {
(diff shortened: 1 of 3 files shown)


internal/telemetry/telemetry.go
```diff
@@ -43,26 +45,59 @@ type Event struct {
...
+func newEvent(cmd *cobra.Command) Event {
 	now := time.Now()
 	cmdPath := cmd.CommandPath()
 	command := strings.ReplaceAll(cmdPath, " ", "-")
+
+	ctxValue, found := valueFromContext(cmd.Context())
+	var duration time.Duration
+	if found {
+		duration = now.Sub(ctxValue.startTime)
+	}
```
tibulca, line 81: if `found == false` the duration will default to 0 I think.
Maybe in this case we should return an error and the `track` function below to handle that error (log and return)?
(so the signature of this function will be `func newEvent(cmd *cobra.Command) (*Event, error)`)

internal/telemetry/telemetry.go
```diff
@@ -43,26 +45,59 @@ type Event struct {
...
+
+func NewContext() context.Context {
+	return context.WithValue(context.Background(), contextKey, telemetryContextValue{
+		startTime: time.Now(),
+	})
+}
+
+func valueFromContext(ctx context.Context) (telemetryContextValue, bool) {
+	value, ok := ctx.Value(contextKey).(telemetryContextValue)
+	return value, ok
```
tibulca, line 62:
```suggestion
	return ctx.Value(contextKey).(telemetryContextValue)
```

internal/telemetry/telemetry.go
fmenezes: Telemetry is meant to ignore errors completely so commands are never affected by telemetry, also I don't want to not track an event just because it did not use context correctly, chances are if we are receiving 0 duration context was not carried out to that specific command which would be about fixing it instead of not tracking.

internal/telemetry/telemetry.go
fmenezes: see my comment above, I don't think this is needed

johansteyn:
Do we need to store the start time in the context?

I'm fine to go with that if there are advantages to doing it that way (eg: if we need to access the start time in other parts of the app) . 

Otherwise, a simple "startTime" global variable in telemetry.go might suffice:
  https://github.com/mongodb/mongodb-atlas-cli/pull/1164

internal/telemetry/telemetry.go
tibulca: `track` function is only logging errors, not affecting the commands, that's what I was also thinking when sugested "and the track function below to handle that error"
Even if capturing the event is more important than, in this case, the correctness of its duration, we should not simply hide a possible mishandling of the context. At least we should log that as an error (like we do in the track function)

fmenezes:
> Do we need to store the start time in the context?
> 
> I'm fine to go with that if there are advantages to doing it that way (eg: if we need to access the start time in other parts of the app) .
> 
> Otherwise, a simple "startTime" global variable in telemetry.go might suffice: #1164

@johansteyn thanks for the suggestion, global variables are normally considered something to avoid (in many programming languages) unless their value is static (now is not). Also today I don't see how context would not be located (meaning 0 instead of actual value), the main reason to use context here is that it is created in the main function and it will be passed down to all commands and API calls. 

The official documentation says `Use context Values only for request-scoped data that transits processes and APIs, not for passing optional parameters to functions.` at https://pkg.go.dev/context#WithValue. Which I feel works well for something like telemetry.

This opens doors for us (in future) to track many different timings like API calls latency, serialization/deserialization across commands and APIs and etc.

Let me know what you think @johansteyn @tibulca.

internal/telemetry/telemetry.go
tibulca: image

internal/telemetry/telemetry.go
fmenezes: @tibulca I'm now calling logError, let me know what you think

tibulca (approved):

johansteyn:
> @johansteyn thanks for the suggestion, global variables are normally considered something to avoid (in many programming languages) unless their value is static (now is not). 

I agree about avoiding global variables, though I think it's more applicable to long-running multi-threaded apps. With the CLI being a short-lived single-threaded app I'm not too concerned about global variables. Besides, contextKey itself is also a global variable, along with fs and maxCacheFileSize, and with all of them being private it's probably fine.

> The official documentation says `Use context Values only for request-scoped data that transits processes and APIs, not for passing optional parameters to functions.` at https://pkg.go.dev/context#WithValue. Which I feel works well for something like telemetry.

Their use of the term "request-scoped" likely means they are referring to long-running multi-threaded apps.
BTW. while they state that the context key should not be of type string, their example does just that :)

Anyway, it's not a blocker using context, so I'm happy to approve (unless @tibulca wants to chime in first?)

johansteyn (approved):
LGTM

fmenezes:
> I agree about avoiding global variables, though I think it's more applicable to long-running multi-threaded apps.
for sure that is a bigger concerns rise on long running applications, but I think it is a good and concise approach to avoid even in short-lived processes

> Besides, contextKey itself is also a global variable, along with fs and maxCacheFileSize, and with all of them being private it's probably fine.
The difference there is they are static, again more of a concern for long-running processes

> BTW. while they state that the context key should not be of type string, their example does just that :)

Well it is actually `favContextKey` which is declared as `type favContextKey string`.

For go that is a totally valid custom type.

PR #1186, opened 2022-05-06
CLOUDP-120446: Telemetry - Track Events for All Errors

fmenezes:
## Proposed changes

Telemetry - Track Events for All Errors

_Jira ticket:_ CLOUDP-120446

Closes #[issue number]

## Checklist


johansteyn:
Filipe, what happens if the error message is very long, or spans multiple lines?
We may want to trim error messages if they are excessive.
I'll experiment to come up with examples...

internal/telemetry/telemetry.go
```diff
@@ -69,219 +40,28 @@ func TrackCommand(cmd *cobra.Command) {
...
-	if exists {
-		info, statError := fs.Stat(filename)
-		if statError != nil {
-			return nil, statError
-		}
-		size := info.Size()
-		if size > maxCacheFileSize {
-			return nil, errors.New("telemetry cache file too large")
-		}
+func TrackCommandError(cmd *cobra.Command, e error) {
```
johansteyn, line 53: The TrackCommand and TrackCommandError functions are almost identical. We can combine them into a single TrackCommand function that takes both parameters, and just call it as appropriate.

fmenezes:
ready for another look @johansteyn

internal/telemetry/event_test.go
```diff
@@ -0,0 +1,303 @@
...
+
+import (
+	"errors"
+	"runtime"
+	"testing"
+	"time"
+
+	"github.com/spf13/cobra"
+
+	"github.com/mongodb/mongocli/internal/config"
```
andreaangiolillo, line 25: run go import

internal/cli/root/atlas/builder.go
```diff
@@ -124,7 +124,7 @@ func Builder(profile *string) *cobra.Command {
 			if check, isHb := notifier.shouldCheck(); check {
 				_ = notifier.notifyIfApplicable(isHb)
 			}
-			telemetry.TrackCommand(cmd)
+			telemetry.TrackCommand(cmd, nil)
```
andreaangiolillo, line 127: [q] Why in atlascli the error is always `nill`?

internal/telemetry/tracker.go
```diff
@@ -0,0 +1,109 @@
...
+
+	return &tracker{
+		fs:               afero.NewOsFs(),
+		maxCacheFileSize: defaultMaxCacheFileSize,
+		cacheDir:         cacheDir,
+	}, nil
+}
+
+func (t *tracker) track(cmd *cobra.Command, e error) error {
+	options := []eventOpt{withCommandPath(cmd), withDuration(cmd), withFlags(cmd), withProfile(), withVersion(), withOS(), withAuthMethod(), withService(), withProjectID(cmd), withOrgID(cmd), withTerminal(), withInstaller(t.fs)}
```
andreaangiolillo, line 57: [q] Why are you creating different functions to set the properties if they are always set? Why not have one function `newProperties` that adds all the properties for an event?

fmenezes: The code was originally together linter was complaining about it getting too big and it was had to follow.

Since we already use the option pattern in the CLI it felt the right way to modularize the code.

internal/cli/root/atlas/builder.go
fmenezes: Success calls won't have an error only, originally I had TrackCommand and TrackCommandError.  As @johansteyn pointed they were nearly identical so he suggest to merge the functions.

internal/telemetry/event_test.go
fmenezes: If this is now a requirement we should have a test for it.

internal/telemetry/tracker.go
andreaangiolillo: Thanks!

internal/telemetry/event_test.go
fmenezes: I see what you mean, turns out `goimports` respects manually grouped imports, we should manually remove all empty lines and then run `goimports`

johansteyn:
Felipe, I've been trying different edge cases to verify if errors are being tracked as expected.

If I run "atlas config describe" without an argument, or with an invalid argument the errors are tracked as expected.

But if I delete my config file and run those same commands, the errors are still displayed in stdout but they are not tracked.

colm-quinn:
> Felipe, I've been trying different edge cases to verify if errors are being tracked as expected.
> 
> If I run "atlas config describe" without an argument, or with an invalid argument the errors are tracked as expected.
> 
> But if I delete my config file and run those same commands, the errors are still displayed in stdout but they are not tracked.

Are they stored in the cache until a new config is created or fully not tracked (lost)?

johansteyn:
At the moment all tracking is to the cache (file), and the error is not written to the cache at all (ie. it's lost).

I am wondering if the absence of a config file means it's not able to determine whether or not telemetry is enabled...

johansteyn:
Telemetry should be enabled despite the lack of a config file, because I have the environment variable set:

`MONGODB_ATLAS_TELEMETRY_FEATURE_FLAG=true`

fmenezes:
@johansteyn I've just tested and I agree, telemetry is not tracking because our default state is false currently. In other words when config file is not present we haven't saved `telemetry_enabled = true` at least once so when the system reads it, it returns false. I think this bug requires a Jira of it's own.

johansteyn:
I have submitted a separate ticket: CLOUDP-122243

johansteyn (approved):
LGTM.
I have submitted a separate ticket for tracking errors in the absence of a config file: CLOUDP-122243

PR #1190, opened 2022-05-06
fix: fix go.mod

gssbzn:
## Proposed changes

Not sure when this happened but our go.mod is all over the place, you are suppose to have two sections, one for direct and one for indirect dependencies but we had multiple sections all with a combination of all and making the file really hard to mantain and probably breaking some tooling around it


colm-quinn (dismissed):
LGTM

Also on how it happened, is the linter not catching these? I was going to suggest the update to go 1.18, but that looks like only generics aren't getting well handled by golangci-lint

gssbzn:
@colm-quinn there's no linter for `go.mod` extralegally enough, and it's a file that's suppose to be managed by tooling, my feeling somebody added things manually and just the tooling couldn't cope and decided to do the best it could

fmenezes (dismissed):
lgtm

PR #833, opened 2021-09-08
Backfill docs content for mongocli config commands

melissamahoney-mongodb:
This PR backfills docs content for the following config commands:
DOCSP-16826: Backfill mongocli-config-set
DOCSP-16825: Backfill mongocli-config-rename
DOCSP-16824: Backfill mongocli-config-list
DOCSP-16823: Backfill mongocli-config-describe
DOCSP-16822: Backfill mongocli-config-delete
DOCSP-16827: Backfill mongocli-config

Backfilled content includes adding:
- Detail to existing help content
- Long descriptions with additional detail where necessary
- Descriptions of arguments
- Examples

PR #1208, opened 2022-05-10
 docs: Backfill docs content for mongocli config commands

gssbzn:
## Proposed changes

Replaces #833

Alternatively, if this is a very minor, and self-explanatory change, feel free to remove this section.
-->

### DIFF
--- a/docs/atlascli/command/atlas-config-rename.txt
+++ b/docs/atlascli/command/atlas-config-rename.txt
@@ -19,7 +19,27 @@ Syntax
 
 .. code-block::
 
-   atlas config rename <oldName> <newName> [options]
+   atlas config rename <oldProfileName> <newProfileName> [options]
+
+Arguments
+---------
+
+.. list-table::
+   :header-rows: 1
+   :widths: 20 10 10 60
+
+   * - Name
+     - Type
+     - Required
+     - Description
+   * - oldProfileName
+     - string
+     - true
+     - Name of the profile to rename.
+   * - newProfileName
+     - string
+     - true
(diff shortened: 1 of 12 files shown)


docs/atlascli/command/atlas-config-rename.txt
```diff
@@ -53,3 +73,11 @@ Inherited Options
      - false
      - Profile to use from your configuration file.
 
+Examples
+--------
+
+.. code-block::
+
+   Rename a profile called myProfile to testProfile:
```
sarahsimpers, line 81: **Question:** does this line need to be part of the code-block directive? It results in the description of the command being copied along with the command, which is annoying for a user that is trying to copy and paste the command only into the CLI. I see we did this in [other places](https://www.mongodb.com/docs/mongocli/stable/command/mongocli-auth-login/#examples) so maybe we had a reason for that, but just asking. I would think it goes above the code-block but below "Examples" for all of these.

sarahsimpers (approved):
LGTM with one question. Thank you @gssbzn !!

docs/atlascli/command/atlas-config-rename.txt
gssbzn: it doesn't but it's not a change for this PR, this was kind of discussed with @melissamahoney-mongodb when we worked on https://github.com/mongodb-labs/cobra2snooty, that lib is the one doing the formatting, we can look into how to improve this but it will require a change on the lib and probably fixing a lot of the current docs to adapt to it

PR #1209, opened 2022-05-10
CLOUDP-121846: Telemetry - Support DO_NOT_TRACK Environment Variable

fmenezes:
## Proposed changes

Telemetry - Support DO_NOT_TRACK Environment Variable

_Jira ticket:_ CLOUDP-121846

Closes #[issue number]

## Checklist

- [x] I have updated [e2e/E2E-TESTS.md](https://github.com/mongodb/mongodb-atlas-cli/blob/master/e2e/E2E-TESTS.md) (if an e2e test has been added)

### DIFF
--- a/internal/config/profile.go
+++ b/internal/config/profile.go
@@ -457,16 +457,13 @@ func (p *Profile) IsTelemetryEnabledSet() bool {
 // TelemetryEnabled get the configured telemetry enabled value.
 func TelemetryEnabled() bool { return Default().TelemetryEnabled() }
 func (p *Profile) TelemetryEnabled() bool {
-	if !isTelemetryFeatureFlagSet() {
-		return false
-	}
-	return p.GetBool(telemetryEnabled)
+	return isTelemetryFeatureAllowed() && p.GetBool(telemetryEnabled)
 }
 
 // SetTelemetryEnabled sets the telemetry enabled value.
 func SetTelemetryEnabled(v bool) { Default().SetTelemetryEnabled(v) }
 func (p *Profile) SetTelemetryEnabled(v bool) {
-	if !isTelemetryFeatureFlagSet() {
+	if !isTelemetryFeatureAllowed() {
 		return
 	}
 	SetGlobal(telemetryEnabled, v)
@@ -477,7 +474,7 @@ func boolEnv(key string) bool {
 	return ok && IsTrue(value)
 }
 
-func isTelemetryFeatureFlagSet() bool {
+func isTelemetryFeatureAllowed() bool {
 	return ToolName == AtlasCLI && !boolEnv("DO_NOT_TRACK") && boolEnv(AtlasCLIEnvPrefix+"_TELEMETRY_FEATURE_FLAG")
 }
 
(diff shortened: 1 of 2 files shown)


blva (dismissed):
LGTM

tibulca (approved):

blva (dismissed):
lgtm

internal/config/profile.go
```diff
@@ -457,24 +457,25 @@ func (p *Profile) IsTelemetryEnabledSet() bool {
...
-func isTelemetryFeatureFlagSet() bool {
-	_, featureFlagSet := os.LookupEnv("MONGODB_ATLAS_TELEMETRY_FEATURE_FLAG")
-	return featureFlagSet
+func boolEnv(key string) bool {
+	value, ok := os.LookupEnv(key)
+	return ok && IsTrue(value)
+}
+
+func isTelemetryFeatureAllowed() bool {
+	return ToolName == AtlasCLI && !boolEnv("DO_NOT_TRACK") && boolEnv(AtlasCLIEnvPrefix+"_TELEMETRY_FEATURE_FLAG")
```
andreaangiolillo, line 478: Viper will add automatically the prefix to the env variable https://github.com/spf13/viper#working-with-environment-variables. I am not sure how  `doNotTrack`  is used but all the env variable should have the `MONGODB_ATLAS_CLI` prefix
```suggestion
	return ToolName == AtlasCLI && !viper.GetBool(doNotTrack) && viper.GetBool(telemetryFeatureFlag")
```

fmenezes: @andreaangiolillo the problem of using viper here is it will check the configuration file as well, and these two properties are not suppose to be in the config at all.

DO_NOT_TRACK is more of a standard so no prefix for it.

andreaangiolillo: > these two properties are not supposed to be in the config at all

ah okay. I don't understand why. Could you clarify? so far, all the env vars have been added to the conf file, so this decision seems a little strange to me. I don't understand how checking the conf file may be a problem 🤔  the conf file is not going to be changed manually by customers (we are not documenting it) so we can assume these prop will not be in the conf file.

fmenezes: We have a telemetry_enable configuration. That comes from file/env var.

Telemetry_Feature_Flag is a temporary measure created by @johansteyn so we don't track customers now, we are suppose to remove it. If we use config file we will end up with an entry that serves no purpose.

DO_NOT_TRACK is not atlascli env var, it is a standard documented in https://consoledonottrack.com which will not have a config file entry nor prefix in it.

andreaangiolillo: Thanks for the clarification! : )

johansteyn: @andreaangiolillo the ticket has a link to :https://consoledonottrack.com/

It's a generic env var that the industry seems to be adopting to disable tracking across the board, ie. it is not specific to any particular apps.

johansteyn (approved):
LGTM.

Checked out the branch and verified that it works correctly. Thanks!

PR #1212, opened 2022-05-11
CLOUDP-120444: [Atlas CLI] Telemetry - Send Track Events from the CLI to the Atlas Telemetry Endpoint

johansteyn:
## Proposed changes

This adds the functionality of sending a batch of tracking events to the new Atlas telemetry endpoint.

It uses a custom transport with a 200ms timeout when sending track events to the Atlas endpoint.

For now, the batch will consist of only the single event being tracked. If the event cannot be sent, it will be cached instead.

In a follow-up ticket we will implement the logic required to also send previously cached events. 

_Jira ticket:_ CLOUDP-120444

Closes #[issue number]

## Checklist


internal/telemetry/tracker.go
```diff
@@ -106,3 +116,24 @@ func (t *tracker) save(event Event) error {
...
+	}
+	s, err := store.New(store.AuthenticatedPreset(config.Default()), store.WithContext(ctx), store.Telemetry())
+	if err != nil {
+		return err
+	}
+	client, err := s.GetAtlasClient()
+	if err != nil {
+		return err
+	}
+	request, err := client.NewRequest(ctx, http.MethodPost, urlPath, &events)
```
fmenezes, line 133: Let's move this method to be inside the store like other endpoints, this would give us the advantage of being able to mock it and unit test this function properly

fmenezes, line 133: Let's move this method to be inside the store like other endpoints, this would give us the advantage of being able to mock it and unit test this function properly

fmenezes (changes requested):
Let's move this method to be inside the store like other endpoints, this would give us the advantage of being able to mock it and unit test this function properly

internal/telemetry/tracker.go
johansteyn: OK, I have moved the "send" function to the store package (and renamed it SendEvents).

internal/store/telemetry.go
```diff
@@ -0,0 +1,45 @@
...
+import (
+	"context"
+	"net/http"
+
+	"github.com/mongodb/mongodb-atlas-cli/internal/config"
+)
+
+const urlPath = "api/private/v1.0/telemetry/events"
+
+func SendEvents(ctx context.Context, body interface{}) error {
```
fmenezes, line 26: If you follow other store methods we normally have an interface and expose that

fmenezes, line 26: This needs to be an interface something like EventsSender or TelemetryEventSender something like that

internal/store/telemetry.go
```diff
@@ -0,0 +1,45 @@
...
+	if config.Service() != config.CloudService {
+		// Only send events to Atlas - not to AtlasGov or OpsManager or CloudManager
+		return nil
+	}
+	s, err := New(AuthenticatedPreset(config.Default()), WithContext(ctx), Telemetry())
```
fmenezes, line 31: This store creation should be done in the struct calling `SendEvents` func

johansteyn: Done in latest commit.

internal/store/store.go
```diff
@@ -87,6 +89,18 @@ var skipVerifyTransport = &http.Transport{
 	TLSClientConfig:       &tls.Config{InsecureSkipVerify: true}, //nolint:gosec // this is optional for some users,
 }
 
+var telemetryTransport = &http.Transport{
```
fmenezes, line 92: [q] why do we need a custom Transport here?

internal/store/store.go
```diff
@@ -268,6 +291,14 @@ func (s *Store) setOpsManagerClient(client *http.Client) error {
 	return nil
 }
 
+func (s *Store) GetAtlasClient() (client *atlas.Client, err error) {
```
fmenezes, line 294: [q] do we need this function?

internal/store/telemetry.go
```diff
@@ -0,0 +1,45 @@
...
+const urlPath = "api/private/v1.0/telemetry/events"
+
+type EventsSender interface {
+	SendEvents(body interface{}) error
+}
+
+func (s *Store) SendEvents(body interface{}) error {
+	switch s.service {
+	case config.CloudService:
+		request, err := s.client.(*atlas.Client).NewRequest(s.ctx, http.MethodPost, urlPath, body)
```
fmenezes, line 36: nit: avoid double casting by
```suggestion
		atlasClient := s.client.(*atlas.Client)
		request, err := atlasClient.NewRequest(s.ctx, http.MethodPost, urlPath, body)
```

internal/store/telemetry.go
```diff
@@ -0,0 +1,45 @@
...
+		if err != nil {
+			return err
+		}
+		_, err = s.client.(*atlas.Client).Do(s.ctx, request, nil)
```
fmenezes, line 40: nit:
```suggestion
		_, err = atlasClient.Do(s.ctx, request, nil)
```

fmenezes (dismissed):

internal/store/telemetry.go
johansteyn: Changed in latest commit.

internal/store/telemetry.go
johansteyn: Changed in latest commit.

fmenezes (approved):
Thanks for the changes

PR #1243, opened 2022-05-18
CLOUDP-122686: [Atlas CLI] Telemetry - Add Logic for Caching and Sending of Track Events

johansteyn:
This is where we implement the logic to decide when to cache and when to send track events.

The logic is documented in the [spec](https://docs.google.com/document/d/1TxuoGmIdpTjs6YxHLidkwMwx92ZB_vFPPKn3iORXX5w/edit#heading=h.i82u0fetyv77), which covers various scenarios and explains the decisions on 10MB cache size limit and batch size limit of 100 events per request.

## Proposed changes

_Jira ticket:_ CLOUDP-#

Closes #[issue number]

## Checklist

### DIFF
--- a/internal/telemetry/tracker.go
+++ b/internal/telemetry/tracker.go
@@ -20,6 +20,8 @@ import (
 	"errors"
 	"os"
 	"path/filepath"
+	"strconv"
+	"time"
 
 	"github.com/mongodb/mongodb-atlas-cli/internal/config"
 	"github.com/mongodb/mongodb-atlas-cli/internal/store"
@@ -30,7 +32,8 @@ const (
 	cacheFilename           = "telemetry"
 	dirPermissions          = 0700
 	filePermissions         = 0600
-	defaultMaxCacheFileSize = 100_000_000 // 100MB
+	defaultMaxCacheFileSize = 10_000_000 // 10MB
+	defaultMaxBatchSize     = 100
 )
 
 type tracker struct {
@@ -38,6 +41,7 @@ type tracker struct {
 	maxCacheFileSize int64
 	cacheDir         string
 	store            store.EventsSender
+	maxBatchSize     int
 }
 
 func newTracker(ctx context.Context) (*tracker, error) {
@@ -58,6 +62,7 @@ func newTracker(ctx context.Context) (*tracker, error) {
 		maxCacheFileSize: defaultMaxCacheFileSize,
 		cacheDir:         cacheDir,
 		store:            telemetryStore,
+		maxBatchSize:     defaultMaxBatchSize,
 	}, nil
 }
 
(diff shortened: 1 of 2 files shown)


internal/telemetry/tracker.go
```diff
@@ -63,6 +67,7 @@ func newTracker(ctx context.Context) (*tracker, error) {
 		cacheDir:         cacheDir,
 		store:            telemetryStore,
 		storeSet:         storeSet,
+		maxBatchSize:     defaultMaxBatchSize,
```
fmenezes, line 70: Personally I think this is getting too complex, I think we should have either number of items in cache OR max cache file size and always send everything.

colm-quinn: +1 on the suggestion here. Limiting to a max number of items makes the performance predictable and also achieves the key product goal of tracking the initial events as a priority.

johansteyn: Sorry for the misunderstanding - perhaps I can think up better variable names to make it clearer...

`maxCacheFileSize` is the maximum size of the cache file.
`maxBatchSize` is the maximum number of events that we'll send in a POST request.

I could rename `maxBatchSize` to `maxRequestPayloadSize` for example.

johansteyn: I understand. That will indeed simplify things, though we will sacrifice caching a large number of events, ie. it would be unreasonable to cache 20 thousand events in a 10MB cache file and send all those events within one command run.

So, I have made changes to specify only the maximum cache file size, choosing a value of 500KB. That should allow us to cache approximately 800 events, while ensuring that our POST request payload also won't exceed 500KB.

I think that's a reasonable size limit - let me know what you think.

fmenezes: I think 800 events are ok, I think it is unlikely users will ever run a situation where they ran so many commands without a profile but that is possible.

internal/telemetry/tracker.go
```diff
@@ -118,13 +116,14 @@ func (t *tracker) openCacheFile() (afero.File, error) {
...
 
+// Append a single event to the cache file.
 func (t *tracker) save(event Event) error {
 	file, err := t.openCacheFile()
 	if err != nil {
 		return err
 	}
 	defer file.Close()
-	data, err := json.Marshal(event)
+	data, err := json.MarshalIndent(event, "", "  ")
```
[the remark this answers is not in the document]
johansteyn: I had in in originally, then removed it to not waste space. But with the latest changes where we're setting a size limit of 500KB, which is far lower that the original 10MB limit, are we really concerned about space?

Having it formatted makes it so much easier to work with during development (or to demo), so I'd like to keep it formatted if it's not to serious an issue...

fmenezes (approved):
lgtm

internal/telemetry/tracker.go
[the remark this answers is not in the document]
colm-quinn: Let's make a ticket to track encoding and/or stripping space to make the final content smaller.

At the least, we could compress & encode before sending over the wire, as there's no need to send more data than required.

Note: Suggesting ticket as we're talking about optimisations, not a strict requirement.

PR #1245, opened 2022-05-18
CLOUDP-122378 mongodb version can be set only for non-shared clusters

tibulca:
## Proposed changes

_Jira ticket:_ CLOUDP-122378

Closes #[issue number]

## Checklist

- [x] I have updated [e2e/E2E-TESTS.md](https://github.com/mongodb/mongodb-atlas-cli/blob/master/e2e/E2E-TESTS.md) (if an e2e test has been added)

### DIFF
--- a/internal/cli/atlas/clusters/create.go
+++ b/internal/cli/atlas/clusters/create.go
@@ -122,9 +122,9 @@ func (opts *CreateOpts) applyOpts(out *atlas.AdvancedCluster) {
 
 	if !opts.isTenant() {
 		out.DiskSizeGB = &opts.diskSizeGB
+		out.MongoDBMajorVersion = opts.mdbVersion
 	}
 
-	out.MongoDBMajorVersion = opts.mdbVersion
 	out.ReplicationSpecs = []*atlas.AdvancedReplicationSpec{replicationSpec}
 }
 


internal/cli/atlas/clusters/create.go
```diff
@@ -122,9 +122,9 @@ func (opts *CreateOpts) applyOpts(out *atlas.AdvancedCluster) {
 
 	if !opts.isTenant() {
 		out.DiskSizeGB = &opts.diskSizeGB
+		out.MongoDBMajorVersion = opts.mdbVersion
```
tibulca, line 125: same as with the DiskSize, the db version can be set only for non-shared clusters

fmenezes: @tibulca was this bug introduced by a change in the API? Do we know when was that changed?

tibulca: no change in the API.
Your commit https://github.com/mongodb/mongodb-atlas-cli/commit/71be4036d26d4fcf2cc7ef3d27e2a472cfe21717 actually uncovered a bug in our code. 
The store was initialized before the profile was loaded while trying to get the default db major version. This was failing so the create cluster was not setting this field. 
Your change fixed this, but now actually having a version (5.0) was causing the create cluster (M0, M2 and M5) to fail, since this is readonly (see mms code: https://github.com/10gen/mms/blob/16b4c8008ae5b51cd597a97ffaa9d892acc48ac9/server/src/main/com/xgen/svc/mms/api/res/atlas/ApiAtlasClusterDescriptionV15Resource.java#L885)

tibulca: also, we have same behaviour in UI, where if you select a shared cluster you cannot set a db version

fmenezes (approved):

internal/cli/atlas/clusters/create.go
fmenezes: thanks for the explanation

PR #1254, opened 2022-05-20
CLOUDP-122239: [Atlas CLI] Telemetry - Track User Interrupts

johansteyn:
## Proposed changes

_Jira ticket:_ CLOUDP-#

Closes #[issue number]

## Checklist

### DIFF
--- a/cmd/atlas/atlas.go
+++ b/cmd/atlas/atlas.go
@@ -46,7 +46,6 @@ func Execute() {
 	}
 }
 
-// TODO: add to mongocli as well...
 func initSignalHandler(cmd *cobra.Command) {
 	c := make(chan os.Signal, 1)
 	signal.Notify(c,
@@ -55,11 +54,8 @@ func initSignalHandler(cmd *cobra.Command) {
 		syscall.SIGQUIT) // CTRL-\
 	go func() {
 		sig := <-c
-		message := fmt.Sprintf("Error: %v\n", sig)
-		fmt.Println()
-		fmt.Println(message)
-		handleError(cmd, errors.New(message))
-		os.Exit(1)
+		fmt.Printf("\nError: %s\n", sig.String())
+		handleError(cmd, errors.New(sig.String()))
 	}()
 }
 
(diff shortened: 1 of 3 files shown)


cmd/atlas/atlas.go
```diff
@@ -37,15 +40,33 @@ var (
 func Execute() {
 	ctx := telemetry.NewContext()
 	rootCmd := atlas.Builder(&profile)
+	initSignalHandler(rootCmd)
```
fmenezes, line 43: [q] is this tracking always root command or the command fired?

cmd/atlas/atlas.go
```diff
@@ -37,15 +40,33 @@ var (
...
-		})
-		os.Exit(1)
+		handleError(cmd, err)
 	}
 }
 
+func initSignalHandler(cmd *cobra.Command) {
+	c := make(chan os.Signal, 1)
+	signal.Notify(c,
+		syscall.SIGINT,  // CTRL-C
```
fmenezes, line 52: according to [docs](https://cs.opensource.google/go/go/+/master:src/os/exec_posix.go;l=17) I think we should add os.Interrupt to the list, given in Windows it might be different

internal/cli/iam/projects/list.go
```diff
@@ -46,6 +47,7 @@ func (opts *ListOpts) initStore(ctx context.Context) func() error {
 }
 
 func (opts *ListOpts) Run() error {
+	time.Sleep(time.Minute)
```
fmenezes, line 50:
```suggestion
```

cmd/atlas/atlas.go
johansteyn: Signal handling (trapping) works for the entire application, ie. across all commands. 

Whenever a signal is trapped, an error is created and passed with the rootCmd through to the handleError function , which ensures that the error is tracked in the same way that it is currently being tracked. 

You can try it out for yourself: run "atlas projects list", which will sleep for 1 minute so that you have enough time to press CTRL-C (or CTRL-Z, or CTRL-\) and see that an error is reported in stdout and also tracked.

internal/cli/iam/projects/list.go
johansteyn: This is just here temporarily (hence the *draft* PR) so that you can try it out for yourself without having to very quickly press CTRL-C. I will remove it before marking the PR ready for team review.

cmd/atlas/atlas.go
johansteyn: os.Interrupt is the same as syscall.SIGINT, looking at the source code in exec_posix.go:

`	Interrupt Signal = syscall.SIGINT`

PR #1273, opened 2022-05-30
CLOUDP-123993: Remove unnecessary mentioning to profiles

blva:
## Proposed changes

_Jira ticket:_ CLOUDP-#

Closes #[issue number]

## Checklist


internal/cli/default_setter_opts.go
```diff
@@ -230,6 +239,7 @@ func (opts *DefaultSetterOpts) AskOrg() error {
 		if err2 := telemetry.TrackAskOne(p, &manually); err2 != nil {
 			return err2
 		}
+		opts.AskedOrgsOrProjects = true
```
tibulca, line 242: why are we setting AskedOrgsOrProjects  to true here?

internal/cli/default_setter_opts.go
```diff
@@ -42,15 +42,17 @@ type ProjectOrgsLister interface {
...
+	OpsManagerURL          string
+	ProjectID              string
+	OrgID                  string
+	MongoShellPath         string
+	TelemetryEnabled       bool
+	Output                 string
+	Store                  ProjectOrgsLister
+	OutWriter              io.Writer
+	AskedOrgsOrProjects    bool
+	AskOrgsAndProjectsText string
```
tibulca, line 55: [optional]
AskOrgsAndProjectsText to control the flow (not only as a label) which is a bit confusing.
What do you think about this alternative:
- instead of AskOrgsAndProjectsText have a field `OnMultipleOrgsOrProjects func () void`
- set this field in in login.go (instead of setting opts.AskOrgsAndProjectsText) with the content from displayInitialSelectText()
- in AskOrg / AskProject instead of calling opts.displayInitialSelectText(), call opts.OnMultipleOrgsOrProjects() (if it's not nil)

internal/cli/default_setter_opts.go
blva: We should also count `Do you want to enter the Org ID manually?` as asked. Trying to track if any reference to orgs and projects was done here.

internal/cli/default_setter_opts.go
blva: Yes, that seems like a better extensible approach! Thank you

internal/cli/default_setter_opts.go
```diff
@@ -322,3 +328,9 @@ func (*DefaultSetterOpts) DefaultQuestions() []*survey.Question {
 	}
 	return q
 }
+
+func (opts *DefaultSetterOpts) runOnMultipleProjects() {
```
tibulca, line 332:
```suggestion
func (opts *DefaultSetterOpts) runOnMultipleOrgsOrProjects() {
```

tibulca (approved):

PR #1296, opened 2022-06-08
tests: fix tests for audit_log_line_scanner_test

gssbzn:
Audit logs tests starting failing locally for me. the deep equals we implemented was failing for some reason, i managed to fix this by moving to `github.com/go-test/deep` so the error was probably on the byte generation we were using 

Additional "drive by"s to improve how we were doing matrix test as we were not naming them (which added complexity on debugging the failure) and we were not nesting them in a `t.Run` call back

### DIFF
--- a/internal/decryption/audit_log_line_scanner_test.go
+++ b/internal/decryption/audit_log_line_scanner_test.go
@@ -13,115 +13,102 @@
 // limitations under the License.
 
 //go:build unit
-// +build unit
 
 package decryption
 
 import (
 	"bytes"
 	"encoding/base64"
-	"encoding/json"
 	"testing"
 	"time"
 
+	"github.com/go-test/deep"
 	"github.com/mongodb/mongodb-atlas-cli/internal/decryption/keyproviders"
+	"github.com/openlyinc/pointy"
 )
 
 func buildExpectedLog() []*AuditLogLine {
 	ts := time.UnixMilli(1647253664552)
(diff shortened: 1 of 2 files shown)


internal/decryption/audit_log_line_scanner_test.go
```diff
@@ -13,175 +13,172 @@
...
 
+	expectedLog := buildExpectedLog()
 	testCases := []struct {
+		name           string
 		input          []byte
 		expectedFormat AuditLogFormat
 		expectedLog    []*AuditLogLine
 	}{
 		{
+			name:           "json",
```
gssbzn, line 72: always name your test case, it makes it easier to debug which iteration fails in a loop

internal/decryption/audit_log_line_scanner_test.go
```diff
@@ -13,175 +13,172 @@
...
-			t.Fatalf("expected: %v got: %v", testCase.expectedFormat, format)
-		}
-		var logs []*AuditLogLine
-		for scanner.Scan() {
-			log, err := scanner.AuditLogLine()
+	for _, tt := range testCases {
+		input := tt.input
+		expectedFormat := tt.expectedFormat
+		el := expectedLog
+		t.Run(tt.name, func(t *testing.T) {
```
gssbzn, line 88: when running test cases in a loop is a good idea to wrap them in t.Run

fmenezes (approved):

PR #1313, opened 2022-06-17
[CLOUDP-124966] Update project docs, change argument name

Razeer123:
## Proposed changes

I have added examples and descriptions to the documentation to make the commands more understandable for the users. I also changed the argument name to make it more informative. 

_Jira ticket:_ CLOUDP-124966

## Checklist

- [x] I have signed the [MongoDB CLA](https://www.mongodb.com/legal/contributor-agreement)
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] I have added any necessary documentation in document requirements section listed in [CONTRIBUTING.md](https://github.com/mongodb/mongodb-atlas-cli/blob/master/CONTRIBUTING.md) (if appropriate)
- [x] I have addressed the @mongodb/docs-cloud-team comments (if appropriate)
- [x] I have updated [e2e/E2E-TESTS.md](https://github.com/mongodb/mongodb-atlas-cli/blob/master/e2e/E2E-TESTS.md) (if an e2e test has been added)
- [x] I have run `make fmt` and formatted my code

PR #1317, opened 2022-06-20
CLOUDP-124966: validate arg is an objectID

gssbzn:
Follow up to #1313

```bash
./bin/atlas privateendpoints aws describe 6c 


Error: the provided value '6c' is not a valid ID
```


andreaangiolillo (approved):
LGTM

PR #1332, opened 2022-06-24
CLOUDP-107608: Cloud Backup download restore does not surface deliveryUrl

fmenezes:
## Proposed changes

_Jira ticket:_ CLOUDP-#

Closes #[issue number]

## Checklist

### DIFF
--- a/internal/cli/atlas/backup/restores/list.go
+++ b/internal/cli/atlas/backup/restores/list.go
@@ -43,8 +43,8 @@ func (opts *ListOpts) initStore(ctx context.Context) func() error {
 	}
 }
 
-var restoreListTemplate = `ID	SNAPSHOT	CLUSTER	TYPE	EXPIRES AT{{range .Results}}
-{{.ID}}	{{.SnapshotID}}	{{.TargetClusterName}}	{{.DeliveryType}}	{{.ExpiresAt}}{{end}}
+var restoreListTemplate = `ID	SNAPSHOT	CLUSTER	TYPE	EXPIRES AT	URLs{{range .Results}}
+{{.ID}}	{{.SnapshotID}}	{{.TargetClusterName}}	{{.DeliveryType}}	{{.ExpiresAt}}	{{range $index, $element := .DeliveryURL[0]}}{{if $index}}; {{end}}{{$element}}{{end}}{{end}}
 `
 
 func (opts *ListOpts) Run() error {


fmenezes:
FYI @boooczek @dianchenghu

fmenezes:
I haven't managed to create a restore with more than 1 URL (I believe it would require large datasets) but it would look somewhat like this:
```
$ atlas backups restores list ClusterM10 -P dev
ID                         SNAPSHOT                   CLUSTER   TYPE       EXPIRES AT             URLs
62b59bacd020626db89b71b9   62b599d7d020626db89b6c86             download   2022-06-24T12:13:36Z   https://restore-62b59bacd020626db89b71b9.jc17s.mongodb-dev.net:27017/CqMdoJJXzNJZ9HnJfJqYnuYJ/restore-62b599d7d020626db89b6c86.tar.gz; https://restore-62b59bacd020626db89b71b9.jc17s.mongodb-dev.net:27017/CqMdoJJXzNJZ9HnJfJqYnuYJ/restore-62b599d7d020626db89b6c86.tar.gz; https://restore-62b59bacd020626db89b71b9.jc17s.mongodb-dev.net:27017/CqMdoJJXzNJZ9HnJfJqYnuYJ/restore-62b599d7d020626db89b6c86.tar.gz
```

dianchenghu:
Is that possible that we show the URL in a different row outside of the table?  Also, why does it say URLs when there's only one URL?
image

fmenezes:
> Is that possible that we show the URL in a different row outside of the table? Also, why does it say URLs when there's only one URL? image

There can be more than one URL and also there will be multiple rows for multiple different backup operations.

tibulca (approved):

boooczek:
> There can be more than one URL and also there will be multiple rows for multiple different backup operations.
@fmenezes Following up to what @dianchenghu suggested: a) Within one backup operation, can we have multiple URLs one under another? b) Can we call the header "URL(s)" (this way we'd indicate there can be one or more)?

internal/cli/atlas/backup/restores/restores.go
```diff
@@ -29,6 +29,7 @@ func Builder() *cobra.Command {
 
 	cmd.AddCommand(
 		ListBuilder(),
+		DescribeBuilder(),
```
andreaangiolillo, line 32: [q] Since this is a new command under atlas, shouldn’t we add it only in atlascli?

andreaangiolillo (commented):
The Jira ticket doesn’t explain which solution is the one decided by the PM. My understanding based on the comment on the ticket was that the URL was too long to show in the template. What am I missing? Also a question about adding this command in mongocli

fmenezes:
> The Jira ticket doesn’t explain which solution is the one decided by the PM. My understanding based on the comment on the ticket was that the URL was too long to show in the template. What am I missing? Also a question about adding this command in mongocli

In the Jira ticket we are going with the describe command, all describe commands are tables so the problem of it being too long won't be solved, but at least we will display only one row so we will be certain the URL can go over a line without impacting the table view.

Not to mention bu using a command that returns only one item there is no need for filtering when using json output

internal/cli/atlas/backup/restores/restores.go
fmenezes: done 👍

sarahsimpers (approved):

andreaangiolillo (approved):
LGTM

PR #1381, opened 2022-07-25
CLOUDP-127992: Add exponential backoff when watching a cluster

Razeer123:
## Proposed changes

Adds exponential backoff when watching a cluster to avoid a problem when the cluster becomes unavailable for a few seconds during the upgrade process. It happens because the upgrade endpoint logic is as follows:

1. find a cluster to upgrade
2. copy all the data to a new cluster
3. delete the old cluster
4. rename the new cluster

Between steps 3 and 4, there are a few seconds when the cluster with the name that we're watching for is not available. This PR fixes this issue by waiting exponentially to check if the upgrade process is still in progress.

**Important**: This feature will only allow the `watch` command to recover from 404 for up to 7 seconds. If it takes longer for the server to respond, the command needs to be re-run. 

_Jira ticket:_ [CLOUDP-127992](https://jira.mongodb.org/browse/CLOUDP-127992)

## Checklist

### DIFF
--- a/internal/cli/atlas/clusters/watch.go
+++ b/internal/cli/atlas/clusters/watch.go
@@ -44,18 +44,18 @@ func (opts *WatchOpts) initStore(ctx context.Context) func() error {
 	}
 }
 
+func isRetryable(err error) bool {
+	var atlasErr *atlas.ErrorResponse
+	return errors.As(err, &atlasErr) && atlasErr.ErrorCode == "CLUSTER_NOT_FOUND"
+}
+
 func (opts *WatchOpts) watcher() (bool, error) {
 	result, err := opts.store.AtlasCluster(opts.ConfigProjectID(), opts.name)
 	if err != nil {
 		return false, err
 	}
 	if result.StateName == "UPDATING" {
-		trigger := func(err error) bool {
-			var atlasErr *atlas.ErrorResponse
-			errorCode := "CLUSTER_NOT_FOUND"
-			return errors.As(err, &atlasErr) && atlasErr.ErrorCode == errorCode
-		}
-		opts.SetExponentialBackoffTrigger(trigger)
+		opts.IsRetryableErr = isRetryable
 	}
 	return result.StateName == "IDLE", nil
 }
--- a/internal/cli/atlas/clusters/watch.go
+++ b/internal/cli/atlas/clusters/watch.go
@@ -16,6 +16,7 @@ package clusters
 
 import (
 	"context"
+	"errors"
 	"fmt"
 
 	"github.com/mongodb/mongodb-atlas-cli/internal/cli"
@@ -25,14 +26,14 @@ import (
 	"github.com/mongodb/mongodb-atlas-cli/internal/store"
 	"github.com/mongodb/mongodb-atlas-cli/internal/usage"
 	"github.com/spf13/cobra"
+	atlas "go.mongodb.org/atlas/mongodbatlas"
 )
 
 type WatchOpts struct {
 	cli.GlobalOpts
 	cli.WatchOpts
-	name           string
-	previousStatus string
-	store          store.AtlasClusterDescriber
+	name  string
+	store store.AtlasClusterDescriber
 }
 
 func (opts *WatchOpts) initStore(ctx context.Context) func() error {
@@ -45,13 +46,17 @@ func (opts *WatchOpts) initStore(ctx context.Context) func() error {
 
 func (opts *WatchOpts) watcher() (bool, error) {
 	result, err := opts.store.AtlasCluster(opts.ConfigProjectID(), opts.name)
-	if opts.previousStatus == "UPDATING" && err != nil {
-		return false, &cli.UpdateError{ErrorCode: "CLUSTER_NOT_FOUND_DURING_UPDATE"}
-	}
 	if err != nil {
 		return false, err
 	}
-	opts.previousStatus = result.StateName
+	if result.StateName == "UPDATING" {
+		trigger := func(err error) bool {
+			var atlasErr *atlas.ErrorResponse
+			errorCode := "CLUSTER_NOT_FOUND"
+			return errors.As(err, &atlasErr) && atlasErr.ErrorCode == errorCode
+		}
+		opts.SetExponentialBackoffTrigger(trigger)
+	}
 	return result.StateName == "IDLE", nil
 }
 
--- a/internal/cli/atlas/clusters/watch.go
+++ b/internal/cli/atlas/clusters/watch.go
@@ -30,8 +30,9 @@ import (
 type WatchOpts struct {
 	cli.GlobalOpts
 	cli.WatchOpts
-	name  string
-	store store.AtlasClusterDescriber
+	name           string
+	previousStatus string
+	store          store.AtlasClusterDescriber
 }
 
 func (opts *WatchOpts) initStore(ctx context.Context) func() error {
@@ -44,9 +45,14 @@ func (opts *WatchOpts) initStore(ctx context.Context) func() error {
 
 func (opts *WatchOpts) watcher() (bool, error) {
 	result, err := opts.store.AtlasCluster(opts.ConfigProjectID(), opts.name)
+	if opts.previousStatus == "UPDATING" && err != nil {
+		fmt.Println(err.Error())
+		return false, &cli.UpdateError{ErrorCode: "CLUSTER_NOT_FOUND_DURING_UPDATE"}
+	}
 	if err != nil {
 		return false, err
 	}
+	opts.previousStatus = result.StateName
 	return result.StateName == "IDLE", nil
 }
 
--- a/internal/cli/atlas/clusters/watch.go
+++ b/internal/cli/atlas/clusters/watch.go
@@ -46,7 +46,6 @@ func (opts *WatchOpts) initStore(ctx context.Context) func() error {
 func (opts *WatchOpts) watcher() (bool, error) {
 	result, err := opts.store.AtlasCluster(opts.ConfigProjectID(), opts.name)
 	if opts.previousStatus == "UPDATING" && err != nil {
-		fmt.Println(err.Error())
 		return false, &cli.UpdateError{ErrorCode: "CLUSTER_NOT_FOUND_DURING_UPDATE"}
 	}
 	if err != nil {
--- a/internal/cli/watch_opts.go
+++ b/internal/cli/watch_opts.go
@@ -20,7 +20,6 @@ import (
 	"time"
 
 	"github.com/briandowns/spinner"
-	atlas "go.mongodb.org/atlas/mongodbatlas"
 )
 
 type WatchOpts struct {
@@ -35,6 +34,21 @@ const (
 )
 
 type Watcher func() (bool, error)
+
+type UpdateError struct {
+	ErrorCode string
+}
+
+func (e *UpdateError) Error() string {
+	return fmt.Sprintf("Error with code %v encountered while performing an update.", e.ErrorCode)
+}
+
+func (e *UpdateError) Is(target error) bool {
+	var v *UpdateError
+
+	return errors.As(target, &v) &&
+		e.ErrorCode == v.ErrorCode
+}
 
 // Watch allow to init the OutputOpts in a functional way.
 func (opts *WatchOpts) Watch(f Watcher) error {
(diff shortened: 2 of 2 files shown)


internal/cli/watch_opts.go
```diff
@@ -40,21 +44,44 @@ func (opts *WatchOpts) Watch(f Watcher) error {
...
 				return err
 			}
 		}
 		time.Sleep(defaultWait)
 	}
 }
 
+func (opts *WatchOpts) exponentialBackoff(err error) bool {
+	// 404 and 401 are the most common errors thrown when upgrading cluster, as it is temporarily deleted
+	if opts.n <= 8 && (strings.Contains(err.Error(), "404")) || (strings.Contains(err.Error(), "401")) {
```
gssbzn, line 75: string comparison for errors is a bad practice, please use go1.13 error handling, https://go.dev/blog/go1.13-errors
there are alot of examples on how to check for atlas api errors in the code base

gssbzn (changes requested):

internal/cli/watch_opts.go
```diff
@@ -25,11 +27,13 @@ import (
 type WatchOpts struct {
 	OutputOpts
 	s *spinner.Spinner
+	n int
 }
 
 const (
 	defaultWait = 4 * time.Second
 	speed       = 100 * time.Millisecond
+	base        = 2
```
gssbzn, line 36: unused?

Razeer123: It's used in line 76:
```
backoff := math.Pow(base, float64(opts.n))
```

internal/cli/watch_opts.go
```diff
@@ -40,21 +44,44 @@ func (opts *WatchOpts) Watch(f Watcher) error {
...
+		}
+		if done {
 			opts.stop()
 			return err
 		}
 		if !opts.IsTerminal() {
 			if _, err = fmt.Fprint(opts.ConfigWriter(), "."); err != nil {
+				if opts.exponentialBackoff(err) {
+					continue
+				}
```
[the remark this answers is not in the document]
Razeer123: You're right, I removed this piece of code

internal/cli/watch_opts.go
Razeer123: Thanks for the information, indeed, it wasn't the best way of checking. I improved it.

internal/cli/watch_opts.go
```diff
@@ -40,9 +44,17 @@ func (opts *WatchOpts) Watch(f Watcher) error {
 		return errors.New("no watcher provided")
 	}
 	opts.start()
+	opts.n = 2
 	for {
 		done, err := f()
-		if err != nil || done {
+		if err != nil {
+			if opts.exponentialBackoff(err) {
```
fmenezes, line 51: [q] will this impact cluster deletion?

Razeer123: @fmenezes This will impact watching every command.

internal/cli/watch_opts.go
```diff
@@ -55,6 +67,27 @@ func (opts *WatchOpts) Watch(f Watcher) error {
...
+}
+
+func checkForError(err error, code string) bool {
+	var atlasErr *atlas.ErrorResponse
+	if errors.As(err, &atlasErr) {
+		if atlasErr.ErrorCode == code {
+			return true
+		}
+	}
+	return false
```
gssbzn, line 88:
```suggestion
	return errors.As(err, &atlasErr) && atlasErr.ErrorCode == code 
```

internal/cli/watch_opts.go
```diff
@@ -55,6 +67,27 @@ func (opts *WatchOpts) Watch(f Watcher) error {
...
+		backoff := math.Pow(base, float64(opts.n))
+		opts.n *= 2
+		sleepTime := time.Duration(backoff) * time.Second
+		time.Sleep(sleepTime)
+		return true
+	}
+	return false
+}
+
+func checkForError(err error, code string) bool {
```
gssbzn, line 81:
```suggestion
func isAtlasError(err error, code string) bool {
```

internal/cli/watch_opts.go
gssbzn: also consider, given this is a oneliner if it really deserves a dedicated method

internal/cli/watch_opts.go
gssbzn: Assuming where @fmenezes is coming I also wanted to flag I though we discussed backingafter the first call being a success

gssbzn (commented):
almost there, one small nit

internal/cli/watch_opts.go
```diff
@@ -55,6 +67,27 @@ func (opts *WatchOpts) Watch(f Watcher) error {
 	}
 }
 
+func (opts *WatchOpts) exponentialBackoff(err error) bool {
+	if opts.n <= 8 && checkForError(err, "CLUSTER_NOT_FOUND") {
+		backoff := math.Pow(base, float64(opts.n))
```
fmenezes, line 72: if I'm following right your backoff logic is 4 seconds, 16 seconds and finally 256 seconds. That look like a lot.

How about 1 second, 2 seconds and lastly 4 seconds?

internal/cli/watch_opts.go
```diff
@@ -55,6 +67,27 @@ func (opts *WatchOpts) Watch(f Watcher) error {
...
+		opts.n *= 2
+		sleepTime := time.Duration(backoff) * time.Second
```
fmenezes, line 74: I think your logic seems a bit complicated to follow.

I had to follow the initial of 2 at line 47, the base at line 36 and the calculations at lines 72 & 74, the step of your loop at line 73, and finally the loop stop condition at line 71.

Perhaps we could try a logic without calculations how about something like `backoffTimes := []time.Duration(x * time.Seconds, y * time.Seconds, ...)` and a simple `n.i = 0` and a step by one. I think it would greatly improve code readability.

internal/cli/watch_opts.go
Razeer123: You're right, those waiting time may be exaggeration. I will update it to a much smaller one, as you're suggesting.

internal/cli/watch_opts.go
Razeer123: Thanks, looks cleaner. I removed the method and moved the variables and logic to the `exponentialBackoff`.

internal/cli/watch_opts.go
Razeer123: Good point. I removed the part with calculations, and added `backoffTimes` to the `exponentialBackoff` method. I think it should be easier to understand now.

internal/cli/watch_opts.go
```diff
@@ -68,22 +66,13 @@ func (opts *WatchOpts) Watch(f Watcher) error {
...
-
-func checkForError(err error, code string) bool {
+	backoffTimes := []time.Duration{1 * time.Second, 2 * time.Second, 4 * time.Second}
 	var atlasErr *atlas.ErrorResponse
-	if errors.As(err, &atlasErr) {
-		if atlasErr.ErrorCode == code {
-			return true
-		}
+	errorCode := "CLUSTER_NOT_FOUND"
+	if opts.n < 3 && errors.As(err, &atlasErr) && atlasErr.ErrorCode == errorCode {
```
fmenezes, line 72: how about?
```suggestion
	if opts.n < len(backoffTimes) && errors.As(err, &atlasErr) && atlasErr.ErrorCode == errorCode {
````

File: internal/cli/watch_opts.go [hunk 10]
```diff
@@ -68,22 +66,13 @@ func (opts *WatchOpts) Watch(f Watcher) error {
...
-		opts.n *= 2
-		sleepTime := time.Duration(backoff) * time.Second
-		time.Sleep(sleepTime)
-		return true
-	}
-	return false
-}
-
-func checkForError(err error, code string) bool {
+	backoffTimes := []time.Duration{1 * time.Second, 2 * time.Second, 4 * time.Second}
```
fmenezes, line 69: [q] can this be made a `const`?

internal/cli/watch_opts.go
Razeer123: Good idea, magic numbers are never a good thing.

internal/cli/watch_opts.go
Razeer123: It can't be a `const` as we're using a initializer and it's evaluated in a runtime, but `const` requires value that can be evaluated during the compile time.

internal/cli/atlas/clusters/watch.go
```diff
@@ -44,9 +45,14 @@ func (opts *WatchOpts) initStore(ctx context.Context) func() error {
 
 func (opts *WatchOpts) watcher() (bool, error) {
 	result, err := opts.store.AtlasCluster(opts.ConfigProjectID(), opts.name)
+	if opts.previousStatus == "UPDATING" && err != nil {
```
fmenezes, line 48: how about storing a flag `allowExponentialBackoff`? this way we won't need the custom error or previous status, correct?

Razeer123: Updated

fmenezes (approved):
great work, in the end looks quite small

fmenezes:
@gssbzn do you mind reviewing it again?

internal/cli/atlas/clusters/watch.go
```diff
@@ -47,6 +49,14 @@ func (opts *WatchOpts) watcher() (bool, error) {
 	if err != nil {
 		return false, err
 	}
+	if result.StateName == "UPDATING" {
+		trigger := func(err error) bool {
+			var atlasErr *atlas.ErrorResponse
+			errorCode := "CLUSTER_NOT_FOUND"
+			return errors.As(err, &atlasErr) && atlasErr.ErrorCode == errorCode
+		}
```
gssbzn, line 57: [nit] every time you loop you'll be redefining the function and you don't need closures 

```go
func isRetryable(err error) bool {
	var atlasErr *atlas.ErrorResponse
	return errors.As(err, &atlasErr) && atlasErr.ErrorCode == "CLUSTER_NOT_FOUND"
}

func (opts *WatchOpts) watcher() (bool, error) {
	result, err := opts.store.AtlasCluster(opts.ConfigProjectID(), opts.name)
	if err != nil {
		return false, err
	}
	if result.StateName == "UPDATING" {
		opts.IsRetryableErr = isRetryable
	}
	return result.StateName == "IDLE", nil
}
```

internal/cli/watch_opts.go
```diff
@@ -24,7 +24,8 @@ import (
 
 type WatchOpts struct {
 	OutputOpts
-	s *spinner.Spinner
+	s                         *spinner.Spinner
+	exponentialBackoffTrigger func(err error) bool
```
gssbzn, line 28: setters/getters are non idiomatic in go 
```suggestion
	IsRetryableErr func(err error) bool
```

internal/cli/watch_opts.go
```diff
@@ -34,14 +35,18 @@ const (
 
 type Watcher func() (bool, error)
 
+func (opts *WatchOpts) SetExponentialBackoffTrigger(trigger func(err error) bool) {
+	opts.exponentialBackoffTrigger = trigger
+}
```
gssbzn, line 40: [see above]
```suggestion
```

internal/cli/watch_opts.go
```diff
@@ -55,6 +60,23 @@ func (opts *WatchOpts) Watch(f Watcher) error {
...
+		done, err := f()
+		if opts.exponentialBackoffTrigger(err) {
+			time.Sleep(backoffTime)
+			continue
+		}
+		return done, err
+	}
+	// Should only happen after trying three times (>14 seconds)
+	return f()
+}
```
gssbzn, line 78: couple things, renaming `exponentialBackoffTrigger` to `IsRetryableErr` then similar to recrating variables in loops recreating the times is no bueno
```suggestion
var backoffTimes = []time.Duration{2 * time.Second, 4 * time.Second, 8 * time.Second}

func (opts *WatchOpts) exponentialBackoff(f Watcher) (bool, error) {
	if opts.IsRetryableErr == nil {
		return f()
	}

	for _, backoffTime := range backoffTimes {
		done, err := f()
		if err != nil && opts.IsRetryableErr(err) {
			time.Sleep(backoffTime)
			continue
		}
		return done, err
	}
	// Should only happen after trying three times (>14 seconds)
	return f()
}
```

gssbzn: also optional and a matter of preference, I find the continue logic hard to read 
```suggestion
var backoffTimes = []time.Duration{2 * time.Second, 4 * time.Second, 8 * time.Second}

func (opts *WatchOpts) exponentialBackoff(f Watcher) (bool, error) {
	if opts.IsRetryableErr == nil {
		return f()
	}

	for _, backoffTime := range backoffTimes {
		if done, err := f(); err == nil || !opts.IsRetryableErr(err) {
			return done, err
		}
		time.Sleep(backoffTime)
	}
	// Should only happen after trying three times (>14 seconds)
	return f()
}
```

gssbzn (commented):
this is a really good improvement, some small nits on style

gssbzn (approved):
🚀 thanks for all the improvements here, great work on this one

PR #1514, opened 2022-11-08
DOCSP-26348 Add examples for atlas networking peering azure create

kanchana-mongodb:
- [DOCSP-26348](https://jira.mongodb.org/browse/DOCSP-26348) adds example for the network peering azure create command


andreaangiolillo (approved):
LGTM

PR #1611, opened 2023-01-19
(DOCSP-16656) Backfills docs content for dbusers and events commands

sarahsimpers:
## Proposed changes

Backfills docs content for dbusers and events commands

_Jira ticket:_ 
https://jira.mongodb.org/browse/DOCSP-16656
https://jira.mongodb.org/browse/DOCSP-16657
https://jira.mongodb.org/browse/DOCSP-16658
https://jira.mongodb.org/browse/DOCSP-16659
https://jira.mongodb.org/browse/DOCSP-16661

## Checklist


internal/usage/usage.go
```diff
@@ -34,7 +34,7 @@ const (
 	NExamples                = "Maximum number of example queries to provide that will be improved by a suggested index."
 	NIndexes                 = "Maximum number of indexes to suggest."
 	MDBVersion               = "Major MongoDB version of the cluster."
-	AuthDB                   = "Authentication database name."
+	AuthDB                   = "Authentication database name. If the user authenticates with AWS IAM, x.509, or LDAP, this value should be $external. If the user authenticates with SCRAM-SHA, this value should be admin."
```
gssbzn, line 37: this change affects ops manager/cloud manager docs as we are reusing the same docs and AWS IAM is not valid for those.
If you want to update this copy please create a new variable and use the new variable

internal/usage/usage.go
```diff
@@ -34,7 +34,8 @@ const (
...
+	MCLIAuthDB               = "Authentication database name."
```
gssbzn, line 38: it would've been easier to create `AtlasAuthDB` but if this is the way you want to play it please update to sue this new variable

https://github.com/mongodb/mongodb-atlas-cli/blob/73bd6d30c60771c02af52599b644aafaf323312f/internal/cli/opsmanager/dbusers/create.go#L131

gssbzn (changes requested):

internal/usage/usage.go
sarahsimpers: @gssbzn That makes sense, I can change it to do an Atlas variable as the exception. Right now I'm trying to figure out if I can make some examples only show up for Atlas CLI, so this PR may take a little time while I figure that out

sarahsimpers:
@gssbzn I removed the AWS examples altogether for now, may investigate later if I can add them in for only Atlas ones using annotations and editing the template. I also changed it to use a new variable for the Atlas CLI, not for MCLI. RFAL when you have time. Thank you!

internal/cli/atlas/dbusers/delete.go
```diff
@@ -54,11 +55,16 @@ func DeleteBuilder() *cobra.Command {
 	cmd := &cobra.Command{
 		Use:     "delete <username>",
 		Aliases: []string{"rm"},
-		Short:   "Delete a database user for your project.",
+		Short:   "Remove the specified database user from your project.",
```
kanchana-mongodb, line 58: To me, "your project" here sounds like command automatically detects project. Because user has to provide project ID also, should this be:
```suggestion
		Short:   "Remove the specified database user from the specified project.",
```

internal/cli/atlas/dbusers/delete.go
```diff
@@ -54,11 +55,16 @@ func DeleteBuilder() *cobra.Command {
...
 		Args:    require.ExactArgs(1),
 		Annotations: map[string]string{
-			"usernameDesc": "Username to delete from the MongoDB database.",
+			"usernameDesc": "Username to delete from the MongoDB database. The format of the username depends on the user's method of authentication.",
```
kanchana-mongodb, line 61: Question: Does this delete the user from the database or does this disassociate the user from the project? I am guessing the latter from the command and suggest the following: 
```suggestion
			"usernameDesc": "Username of user in the MongoDB database to delete from the specified project. The format of the username depends on the user's method of authentication.",
```

internal/cli/atlas/dbusers/delete.go
gssbzn: there's a bit of nuance here, users may have set a default project, in which case the projectId flag is optional, adding this for context as I have no opinion on one copy being better than the other

internal/cli/atlas/dbusers/describe.go
```diff
@@ -95,7 +92,11 @@ func DescribeBuilder() *cobra.Command {
 		},
 	}
 
-	cmd.Flags().StringVar(&opts.authDB, flag.AuthDB, convert.AdminDB, usage.AuthDB)
+	if config.BinName() == config.AtlasCLI {
+		cmd.Flags().StringVar(&opts.authDB, flag.AuthDB, convert.AdminDB, usage.AtlasAuthDB)
+	} else {
+		cmd.Flags().StringVar(&opts.authDB, flag.AuthDB, convert.AdminDB, usage.AuthDB)
+	}
```
gssbzn, line 99: there's no need for this `if` you can safety use `usage.AtlasAuthDB` with no major impact, the issue was with `usage.AuthDB` being reused for both ops manager and atlas commands, this is not about mongocli vs atlas cli

internal/cli/atlas/dbusers/describe.go
```diff
@@ -62,20 +62,17 @@ func DescribeBuilder() *cobra.Command {
...
+		Example: fmt.Sprintf(`  # Return the details for the SCRAM SHA-authenticating database user named myDbUser:
   %[1]s dbuser describe myDbUser --authDB admin --output json
 
-  # The following example returns a MongoDB database user with X.509 authentication on an Atlas cluster. Prepend $external with \ to escape the special-use character.
-  %[1]s dbuser describe CN=ellen@example.com,OU=users,DC=example,DC=com --authDB \$external --output json
-
-  # The following example returns a MongoDB database user with AWS IAM authentication on an Atlas cluster. Prepend $external with \ to escape the special-use character.
-  %[1]s dbuser describe arn:aws:iam::772401394250:user/my-test-user --authDB \$external --output json`, cli.ExampleAtlasEntryPoint()),
+  # Return the details for the X.509-authenticating database user with the RFC 2253 Distinguished Name CN=ellen@example.com,OU=users,DC=example,DC=com. Prepend $external with \ to escape the special-use character:
+  %[1]s dbuser describe CN=ellen@example.com,OU=users,DC=example,DC=com --authDB \$external --output json`, cli.ExampleAtlasEntryPoint()),
```
gssbzn, line 75: you can safetly bring back the AWS IAM example, there's not problem with showing it for both atlas cli and mongocli, the only issue was with ops manager commands and only on the flags, examples are different

internal/cli/atlas/dbusers/delete.go
sarahsimpers: @kanchana-mongodb aside from Gus's comment, I wanted to avoid using "specified" twice per short description because it seems a little stiff and awkward. I do think it's important to say that this is the "specified user". So if you have any suggestions that are better than "your project" given that context, please let me know

internal/cli/atlas/dbusers/describe.go
sarahsimpers: Oh ok, I will change it

internal/cli/atlas/dbusers/delete.go
sarahsimpers: I think that command would be atlas projects users delete and this does something different, but I will leave that for @gssbzn to answer

gssbzn: yes, this is about clusters/deployments access and not project access if that makes it any clearer

gssbzn (approved):
LGTM

PR #1613, opened 2023-01-20
feat: validate iamRoleID is an object ID

gssbzn:
Add an extra validation for AWS IAM role ID as we have had a couple of support cases that assume is in the AWS format when it's an ObjectID


blva (approved):
LGTM, great catch

wtrocki (approved):

PR #1619, opened 2023-01-23
CLOUDP-154647: Reduce cost of Atlas E2E Suite - M30 instances

gssbzn:
CLOUDP-154647: conditionally set the tier for gov or commercial so commercial can be an M10 and gov can be the now available M20


AndreiMatei2206 (approved):
LGTM

PR #1626, opened 2023-01-24
(DOCSP-16702) Backfills docs content for perf advisor, quickstart, processes

sarahsimpers:
## Proposed changes

Backfills docs content for performanceAdvisor, Quickstart, and Processes commands.

I also found a few commands for performanceAdvisor where docs said that no options were required, but upon testing found that the processName flag was required, so I changed it to required.

_Jira ticket:_
https://jira.mongodb.org/browse/DOCSP-16702
https://jira.mongodb.org/browse/DOCSP-16704
https://jira.mongodb.org/browse/DOCSP-16705
https://jira.mongodb.org/browse/DOCSP-16706
https://jira.mongodb.org/browse/DOCSP-16734
https://jira.mongodb.org/browse/DOCSP-16735
https://jira.mongodb.org/browse/DOCSP-16736

## Checklist


kyuan-mongodb (approved):
LGTM % minor typo!

docs/mongocli/command/mongocli-iam-users-invite.txt
```diff
@@ -69,7 +69,7 @@ Options
    * - --password
      - string
      - false
-     - User's password.
+     - Password for the database user.
```
gssbzn, line 72: this is similar to the problem we had before, we have reused the constant in more than one context and in this case it doesn't apply, this is an Atlas User and not database user, similar to before it'll require to create a new constant with the new copy specific to database users

sarahsimpers: @gssbzn Thanks, I updated it to use a new variable and changed the description for this one back to be more generic.

internal/cli/performanceadvisor/suggestedindexes/list.go
```diff
@@ -108,5 +111,8 @@ func ListBuilder() *cobra.Command {
 
 	cmd.Flags().StringVar(&opts.ProjectID, flag.ProjectID, "", usage.ProjectID)
 	cmd.Flags().StringVarP(&opts.Output, flag.Output, flag.OutputShort, "", usage.FormatOut)
+
+	_ = cmd.MarkFlagRequired(flag.ProcessName)
```
gssbzn, line 115:
```suggestion
```
This not true for cloud manager/ops manager and we reuse the command

gssbzn: check https://github.com/mongodb/mongodb-atlas-cli/blob/4255220f79167ec6438d3dbe01118cc5aecdb3d7/internal/cli/performance_advisor_opts.go#L46-L53

sarahsimpers: Will remove this where I added it, thanks

internal/cli/atlas/quickstart/quick_start.go
```diff
@@ -556,15 +558,15 @@ func Builder() *cobra.Command {
 
 	cmd.Flags().StringVar(&opts.ClusterName, flag.ClusterName, "", usage.ClusterName)
 	cmd.Flags().StringVar(&opts.Tier, flag.Tier, DefaultAtlasTier, usage.Tier)
-	cmd.Flags().StringVar(&opts.Provider, flag.Provider, "", usage.Provider)
-	cmd.Flags().StringVarP(&opts.Region, flag.Region, flag.RegionShort, "", usage.Region)
+	cmd.Flags().StringVar(&opts.Provider, flag.Provider, defaultProvider, usage.Provider)
+	cmd.Flags().StringVarP(&opts.Region, flag.Region, flag.RegionShort, defaultRegion, usage.Region)
```
gssbzn, line 562: a default here has no difference from a user setting the flag causing the command to assume the user set the flag and skipping the question 
TL;DR let's not change the defaults even if they are the default when no flag is set it'll cause the program to ask the user to confirm
```suggestion
	cmd.Flags().StringVar(&opts.Provider, flag.Provider, "", usage.Provider)
	cmd.Flags().StringVarP(&opts.Region, flag.Region, flag.RegionShort, "", usage.Region)
```

gssbzn (commented):
only wnat comment left, also can you please open a  ticket for us to split the performance advisor commands into atlas and cloud/manager/ops manager, right now some of the things you are trying to fix are correct but we can't because we reused the command and that's an us problem not a docs problem

internal/cli/atlas/quickstart/quick_start.go
sarahsimpers: Oh ok, sorry about that! Removed. Thanks!

gssbzn (approved):
LGTM thanks for all the work here

PR #1627, opened 2023-01-24
docs: update examples to use atlas cli

gssbzn:
Update examples to use atlas CLI and add example for multi cloud cluster

I had the json for multicloud cluster laying around and decided to add it here


blva (approved):
LGTM

PR #1637, opened 2023-01-26
Basic Dockerfile for experimentation

jamesbroadhead:
On M1 macOS, `docker buildx build --platform=linux/amd64 .`


Dockerfile
```diff
@@ -0,0 +1,24 @@
+# syntax=docker/dockerfile:1.3-labs
+#
+# On M1 macOS, `docker buildx build --platform=linux/amd64 .`
+#
+
+FROM registry.access.redhat.com/ubi8/ubi-minimal:8.6
```
wtrocki, line 6: ❤️

Dockerfile
```diff
@@ -0,0 +1,24 @@
...
+enabled=1
+gpgkey=https://pgp.mongodb.com/server-6.0.asc
+EOF
+
+RUN microdnf install yum &&\
+    yum -y update &&\
+    yum clean all &&\
+    microdnf clean all
+
+RUN yum install -y mongodb-atlas
```
wtrocki, line 22: In practical sense it would be better to use local file instead and have separate container build CI/CD job.

wtrocki:
I would look to see if we can improve this a little bit to show team how this can be done (specifically for the 2 CLIs)

.github/workflows/docker.yaml
```diff
@@ -0,0 +1,33 @@
+name: Build CLI Docker image
```
gssbzn, line 1: what about https://goreleaser.com/customization/docker/

wtrocki: Sounds good.
I totally missed that gorelaser files are present: 
https://github.com/mongodb/mongodb-atlas-cli/blob/50e37b59ba912d661654fd4cf8f152356da942c3/build/package/.atlascli.goreleaser.yml#L14

wtrocki:
@gssbzn Some context: Ideally release process should use the same binaries as the one that we have on github. Since release doesn't seem to be running by GitHub I wanted to understand and learn reasons and automation

gssbzn:
> Since release doesn't seem to be running by GitHub I wanted to understand and learn reasons and automation

https://github.com/mongodb/mongodb-atlas-cli/blob/master/RELEASING.md mentions goreleaser and the evg task that does it

PR #1641, opened 2023-01-27
CLOUDP-155672: cli ops manager clusters don't show default RW concern

gssbzn:
## Proposed changes

_Jira ticket:_ CLOUDP-155672

We missed a place to copy the settings between the automation config and the internal cluster state


blva (approved):
LGTM

PR #1643, opened 2023-01-28
CLOUDP-155414: Add prompt for cluster termination protection to quickstart

AndreiMatei2206:
## Proposed changes
Add flag to enable cluster termination protection when creating a cluster interactively through the `quickstart` command.

Screenshot 2023-01-30 at 15 30 41

_Jira ticket:_ CLOUDP-155414

## Checklist


internal/cli/atlas/quickstart/confirm_cluster_setup.go
```diff
@@ -53,12 +53,14 @@ Cluster Disk Size (GiB):		%.1f`, opts.Tier, diskSize)
 [Confirm cluster settings]
 Cluster Name:				%s%s
 Cloud Provider and Region:		%s
+Cluster Termination Protection:		%t
```
AndreiMatei2206, line 56: @dianchenghu @boooczek This is how the confirmation message looks (I highlighted the message for cluster termination protection):
Screenshot 2023-01-29 at 20 45 13

I've seen that for "Load sample data" we only show it if the user has set the flag to `true`, should we do the same thing for cluster termination protection? Or should we always display it, even if it was set to false? And if so, are boolean values ok (true/false) as answers, or should we change it to Yes/No?

boooczek: My opinion: Lets hide when "false". Also it should be consistent with other questions of this type so should be Yes / No.

I'm also not sure about the flag --enableTerminationProtection in --help. If they want to disable they'd do --disableTerminationProtection? 
On that --help please also update the copy slightly to say "Users can't delete a cluster..."

AndreiMatei2206: Ok, will update the confirmation message.

By default termination protection is off when creating a cluster, so we would only need a flag that enables it

dianchenghu: Agree with hide when "false", since the quickstart target users might not need to worry about data protection at this stage.  For the "true/false" copy problem, I think we could change the label to "Enable Cluster Termination Protection: Yes" which might be more consistent with the flag.

AndreiMatei2206: When termination protection is disabled (the message is hidden)
Screenshot 2023-01-30 at 15 35 38

When termination protection is enabled through the flag:
Screenshot 2023-01-30 at 15 37 25

When termination protection flag is not specified, and started creating a cluster through the interactive setup:
Screenshot 2023-01-30 at 15 38 40

dianchenghu: The first two images look good to me! For the last one, I'm not sure if that additional step is required, but I won’t suggest adding it to the interactive flow. Could you confirm with @boooczek  on it?

gssbzn: >"Users can't delete a cluster..."

FYi the original copy is based on docs style guides, which favor active voice while the current change makes it more passive

AndreiMatei2206: Discussed offline: We don't want to alter the interactive setup process any more - it should only ask for options that are crucial to setting up a cluster. Therefore, we will be reverting any prompts related to cluster termination, but keep the flag.

gssbzn (approved):
LGTM if LGTDOCS, please make sure to merge master as I just updated hte copy for projectid and profile and it may end up with conflicts here since youa re also adding some docs

sarahsimpers (approved):
LGTM

PR #1655, opened 2023-02-01
task: update setup commands to use the new mutually exclusive feature

gssbzn:
Update the manual check for exclusive flags to use the built in cobra feature
Plus some drive by copy changes


gssbzn:
evergreen retry

blva (approved):

wtrocki (approved):
Awesome feature from cobra. Verified locally by running help

PR #1664, opened 2023-02-06
Fixed labels for kubernetes E2E tests

igor-karpukhin:
## Proposed changes

Fixed labels for kubernetes E2E tests

## Checklist

### DIFF
--- a/test/e2e/atlas/kubernetes_config_generate_test.go
+++ b/test/e2e/atlas/kubernetes_config_generate_test.go
@@ -27,6 +27,7 @@ import (
 	"strings"
 	"testing"
 
+	"github.com/mongodb/mongodb-atlas-cli/internal/kubernetes/operator/features"
 	"github.com/mongodb/mongodb-atlas-cli/internal/kubernetes/operator/pointers"
 	"github.com/mongodb/mongodb-atlas-cli/internal/kubernetes/operator/resources"
 	"github.com/mongodb/mongodb-atlas-cli/internal/search"
@@ -47,6 +48,10 @@ import (
 )
 
 const targetNamespace = "importer-namespace"
+
+var expectedLabels = map[string]string{
+	features.ResourceVersion: features.LatestOperatorVersion,
+}
 
 func getK8SEntities(data []byte) ([]runtime.Object, error) {
 	b := bufio.NewReader(bytes.NewReader(data))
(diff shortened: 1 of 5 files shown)


wtrocki (dismissed):
LGTM based on integration test passing

test/e2e/atlas/kubernetes_config_generate_test.go
```diff
@@ -47,6 +48,11 @@ import (
 )
 
 const targetNamespace = "importer-namespace"
+const latestOperatorVersion = "1.6.0"
```
andreaangiolillo, line 51: [q] Does this var need to be updated for each operator release?

test/e2e/atlas/kubernetes_config_generate_test.go
```diff
@@ -92,7 +98,7 @@ func InitialSetupWithTeam(t *testing.T) KubernetesConfigGenerateProjectSuite {
 	s.generator = newAtlasE2ETestGenerator(s.t)
 	s.generator.generateTeam("Kubernetes")
 	s.generator.generateEmptyProject(fmt.Sprintf("Kubernetes-%s", s.generator.projectName))
-	s.expectedProject = referenceProject(s.generator.projectName, targetNamespace)
+	s.expectedProject = referenceProject(s.generator.projectName, targetNamespace, expectedLabels)
```
wtrocki, line 101: Maybe we can go with option of removing label instead of creating one?

wtrocki: I think we can fix it in follow up PR.

test/e2e/atlas/kubernetes_config_generate_test.go
igor-karpukhin: @andreaangiolillo I fixed that, but I moved this constant to `crds.go`. As well as I removed the automatic calculation of what version is the latest, as go maps do not preserve the order. Yes, this needs to be updated for each minor (y-stream) release.

test/e2e/atlas/kubernetes_config_generate_test.go
igor-karpukhin: With the "CRD versions" feature there always will be labels, so, I'd say we add and validate them instead of removing. Also, tests can be extended to verify previous version of the operator

gssbzn (approved):
LGTM

test/e2e/atlas/kubernetes_config_generate_test.go
andreaangiolillo: isn't there a way to automatically get this value? For instance, you can add a step in the evergreen task to check the latest operator release and add it to an env variable that you can access inside this test.

PR #1670, opened 2023-02-07
CLOUDP-155907: example integration for the new client

wtrocki:
Example integration of the v2 client into Atlas CLI.
Note that integration is not ready to be merged and is only serving as example.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -36,7 +36,7 @@ require (
 	github.com/stretchr/testify v1.8.1
 	github.com/tangzero/inflector v1.0.0
 	github.com/withfig/autocomplete-tools/packages/cobra v1.2.0
-	go.mongodb.org/atlas v0.22.0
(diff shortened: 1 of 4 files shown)


go.mod
```diff
@@ -36,7 +36,7 @@ require (
 	github.com/stretchr/testify v1.8.1
 	github.com/tangzero/inflector v1.0.0
 	github.com/withfig/autocomplete-tools/packages/cobra v1.2.0
-	go.mongodb.org/atlas v0.22.0
+	go.mongodb.org/atlas v0.21.1-0.20230206173512-87e8a19684b0
```
wtrocki, line 39: NOTE: After v2 gets released version will change to the official release (currently it is points to git commit). 
We cannot use branch name as go tools do not support branches starting with `v.*`

gssbzn:
@wtrocki I'm guessing #1679 replaces this one, can we close if that's true?

PR #1679, opened 2023-02-13
CLOUDP-161238: organizations endpoint example

wtrocki:
## Why

Demo end-to-end Integration with the v2 API client. 

## What

- [x] Create and initialize client with the same amount of customization 
- [x] Replace organizations describe endpoint

## Decisions taken

 - Instead of casting interface I have decided to create a separate v2 client that can be used when needed. 
 - Hardcoded dependency. Can be updated by running  `go get go.mongodb.org/atlas@sdkv2`
 - Distinction between atlas and mongodb api calls were kept despite having the same method calls.

## Verification

```
atlas organizations list 
atlas organizations describe {your org}
```

## Suggestions for improving user experience

- Tag names ending with the `Api` seem redundant. If we remove them we can minimize the number of migration steps.
- `ApiOrganizationView` can be transformed to `Organization`


go.mod
```diff
@@ -36,7 +36,7 @@ require (
 	github.com/stretchr/testify v1.8.1
 	github.com/tangzero/inflector v1.0.0
 	github.com/withfig/autocomplete-tools/packages/cobra v1.2.0
-	go.mongodb.org/atlas v0.22.0
+	go.mongodb.org/atlas v0.21.1-0.20230213105757-80a065fea80c
```
wtrocki, line 39: This looks like downgrade but in fact it does uses the latest version of the client

internal/store/store.go
```diff
@@ -31,6 +31,7 @@ import (
 	"github.com/mongodb-forks/digest"
 	"github.com/mongodb/mongodb-atlas-cli/internal/config"
 	"github.com/mongodb/mongodb-atlas-cli/internal/log"
+	apiv2 "go.mongodb.org/atlas/api/v1alpha"
```
drinkbird, line 34: do we intentionally load `v1alpha` as `apiv2`?

wtrocki: Good catch. It is confusing.
v1alpha is current client version. 
v2 is an atlas api version it targets.

I think we will need to settle with different names to minimize search and replace in future. Something like `atlasapi` or `atlasv2`

go.mod
andreaangiolillo: Could you clarify why? I see that the latest one is v0.22.0 https://github.com/mongodb/go-client-mongodb-atlas/releases

Screenshot 2023-02-14 at 09 16 01

internal/store/organizations.go
```diff
@@ -55,13 +56,15 @@ func (s *Store) Organizations(opts *atlas.OrganizationsListOptions) (*atlas.Orga
 }
 
 // Organization encapsulate the logic to manage different cloud providers.
-func (s *Store) Organization(id string) (*atlas.Organization, error) {
+func (s *Store) Organization(id string) (*v1alpha.ApiOrganizationView, error) {
```
andreaangiolillo, line 59: Is `*v1alpha.ApiOrganizationView` coming from the autogenerated sdk? Could we use v2 for all the autogenerated code?

internal/store/store.go
```diff
@@ -297,6 +306,26 @@ response:
...
 }
 
+/**
+* Creates client for v2 generated API.
+ */
+func (s *Store) createV2Client(client *http.Client) error {
+	modifiers := []apiv2.ClientModifier{
+		apiv2.UseHTTPClient(client),
+		apiv2.UseUserAgent(config.UserAgent),
+		apiv2.UseDebug(log.IsDebugLevel())}
```
andreaangiolillo, line 316: [q] Shouldn't we set versioning as well?

go.mod
gssbzn: if I do `go get -u go.mongodb.org/atlas@master` I get `go.mongodb.org/atlas v0.22.1-0.20230203154520-61aa70417b31` this probably means your branch if out of date with master, ie it doesn't include the commit that was tagged as 0.22.0, can you make sure the preview branch is up to date with master, this will be a recurring task

go.sum
```diff
@@ -377,8 +377,10 @@ github.com/yuin/goldmark v1.1.32/go.mod h1:3hX8gzYuyVAZsxl0MRgGTJEmQBFcNTphYh9de
 github.com/yuin/goldmark v1.2.1/go.mod h1:3hX8gzYuyVAZsxl0MRgGTJEmQBFcNTphYh9decYSb74=
 github.com/yuin/goldmark v1.3.5/go.mod h1:mwnBkeHKe2W/ZEtQ+71ViKU8L12m81fl3OWwC1Zlc8k=
 github.com/yuin/goldmark v1.4.13/go.mod h1:6yULJ656Px+3vBD8DxQVa3kxgyrAnzto9xy5taEt/CY=
-go.mongodb.org/atlas v0.22.0 h1:si5aZBfQrKG156HcpBUZloFg78DezDuzV4Kj/ysKqjw=
-go.mongodb.org/atlas v0.22.0/go.mod h1:XTjsxWgoOSwaZrQUvhTEuwjymxnF0r12RPibZuW1Uts=
+go.mongodb.org/atlas v0.21.1-0.20230206173512-87e8a19684b0 h1:memEF+gOYBck2HSMLOumae0crJNAkpq+KjhTm0/1Vsc=
+go.mongodb.org/atlas v0.21.1-0.20230206173512-87e8a19684b0/go.mod h1:wEY7WlN8ch+KideMvGcGpNoN4/raY4CfT8lax0FzXCU=
+go.mongodb.org/atlas v0.21.1-0.20230213105757-80a065fea80c h1:vjd28N4B9LqBpU4OZ7Seiy6ZZHenrce/ybk5jflVm2g=
+go.mongodb.org/atlas v0.21.1-0.20230213105757-80a065fea80c/go.mod h1:wEY7WlN8ch+KideMvGcGpNoN4/raY4CfT8lax0FzXCU=
```
gssbzn, line 383: make sure to run `go mod tidy` after modifying the `go.mod`  file

internal/store/organizations.go
```diff
@@ -55,13 +56,15 @@ func (s *Store) Organizations(opts *atlas.OrganizationsListOptions) (*atlas.Orga
...
 	switch s.service {
 	case config.CloudService, config.CloudGovService:
-		result, _, err := s.client.(*atlas.Client).Organizations.Get(s.ctx, id)
+		result, res, err := s.clientv2.OrganizationsApi.GetOrganization(s.ctx, id).Execute()
+		defer res.Body.Close()
```
gssbzn, line 63: doesn't the generated code handle this? this feels like a big overside on the generated client

internal/store/organizations.go
```diff
@@ -55,13 +56,15 @@ func (s *Store) Organizations(opts *atlas.OrganizationsListOptions) (*atlas.Orga
...
 		return result, err
 	case config.CloudManagerService, config.OpsManagerService:
-		result, _, err := s.client.(*opsmngr.Client).Organizations.Get(s.ctx, id)
+		result, res, err := s.clientv2.OrganizationsApi.GetOrganization(s.ctx, id).Execute()
+		defer res.Body.Close()
```
gssbzn, line 67: this is not needed, our manual clients self clean 
```suggestion
```

gssbzn, line 67: please rollback these lines, don't change cloud manager/ops manager calls APIs are different, configuration is different this may introduce big side effects

internal/cli/iam/organizations/describe_test.go
```diff
@@ -33,7 +33,7 @@ func TestDescribe_Run(t *testing.T) {
 	mockStore.
 		EXPECT().
 		Organization(gomock.Eq("5a0a1e7e0f2912c554080adc")).
-		Return(&mongodbatlas.Organization{}, nil).
+		Return(&v1alpha.ApiOrganizationView{}, nil).
```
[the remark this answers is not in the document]
wtrocki: Yes.

internal/store/store.go
```diff
@@ -297,6 +306,26 @@ response:
...
+
+	if s.baseURL != "" {
+		modifiers = append(modifiers, apiv2.UseBaseURL(s.baseURL))
+	}
+	newClient, err := apiv2.NewClient(modifiers...)
+	if err != nil {
+		return err
+	}
+	s.clientv2 = newClient
+	return nil
```
gssbzn, line 326: [nit] let's stick with the same naming conventions to avoid confusion
```suggestion
	opts := []apiv2.ClientModifier{
		apiv2.UseHTTPClient(client),
		apiv2.UseUserAgent(config.UserAgent),
		apiv2.UseDebug(log.IsDebugLevel())}

	if s.baseURL != "" {
		opts = append(modifiers, apiv2.UseBaseURL(s.baseURL))
	}
	c, err := apiv2.NewClient(opts...)
	if err != nil {
		return err
	}
	s.clientv2 = c
	return nil
```

internal/store/organizations.go
wtrocki: Would it mean that we will need to supply two models to store?

gssbzn: or an `interface{}` not the best use of strong typing from go but this is what we've done before in cases where they were completely different models

Another goal for a different project, backport the generated common APIs to the ops manager client, the opsmanager client "inherits" from the Atlas clients and overrides where needed but ops manager will never have an openapi spec so any change there will need to be manual

go.mod
wtrocki: I will merge the master branch into sdkv2.

internal/cli/iam/organizations/describe_test.go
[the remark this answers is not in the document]
wtrocki: We do remove `ApiAtlas and View`. Looks like Organization view for some reason was missing Atlas in the name and wasn't transformed

internal/store/organizations.go
wtrocki: > Is *v1alpha.ApiOrganizationView coming from the autogenerated sdk?
Yes. v1alpha just import name that we can change. See other thread.

There is a lot of confusion between golang client version and Atlas API version.
Atlas V2 API client is currently in v1alpha version etc.

internal/store/organizations.go
wtrocki: I want to create separate jira for this specifically.
Body is closed already. Client does handle it. 
It is well known bug in the golangcilint to flag those as not closed when response is returned from method. I will replace this statement with golang cilint exception and link to jira

internal/store/organizations.go
gssbzn: check the projects store https://github.com/mongodb/mongodb-atlas-cli/blob/17c025603e81bed15935b3b55c23ed7bc2b89951/internal/store/projects.go#L70 they are completely different models between ops manager and atlas and yet it works

internal/store/store.go
drinkbird: Indeed it isn't very clear. How about prefixing all apis with `api` and all clients with `client`?

e.g. `clientv1alpa`, `apiv2` etc

wtrocki: it is even more confusing as golang recomends to have `api` prefix for parts that are public interface.

To avoid confusion I settled with atlasv2 as name of the API and it's version. This will also allow us to minimize amount of changes (not including actual client version we can change that in the cli by changing only imports)

internal/store/store.go
wtrocki: What do you mean by versioning? 
If you mean sending information about version to backend?

The client does that at the moment - each client does include a version of the versioned API.
We have also follow up project for some versioning features for atlas cli.

internal/store/organizations.go
wtrocki: That would be at the cost of the casting for request and response objects. 
I have done an investigation and I feel that we have options:

### 1. Use atlasv2 models in arguments and responses. 
Cast and change those atlasv2 models to old models inside store interface to support ops manager. 

#### Benefits

- We can auto-discover new features
- Atlas CLI will be benefiting from versioned API (new versions of models can be used on demand
- Separation/depreciation of the old client will be easier and would not require much engineering

### 2. Keep v1 models and transpose them to the v2 objects fo atlas cli

Technically it will mean less work if we keep interfaces as they are now and transpose them inside the method 
to the v2 format. However in situations when a field will be missing in v1 and required in v2 we will end up 

### 3. Weak store interface and model casting depending on the CLI type

Cli can create or operate with different models depending on the Atlas/OpsManager being present. 

### 4. Split store interface and create store depending on the CLI type

Since we have different cli instances that will be using different clients it actually makes sense to split the store into two different implementations:

- OpsManager Store (that will be using v1 client)
- Atlas Store (that will be using v1 client initially and migrate to v2 client by individual commands)

We will need to introduce another abstraction for each command that will create a required object for the store. 

@swattyT @gssbzn Sorry for the long thread.  It will probably make sense to move this thread to spec doc at some point.

internal/store/organizations.go
wtrocki: Created Jira: CLOUDP-161495

internal/store/store.go
andreaangiolillo: I meant setting the `application/vnd.atlas.2022-10-15+json` in the header of the request.

> The client does that at the moment - each client does include a version of the versioned API.

Ah okay. Thanks for the clarifications.

wtrocki:
We have this PR against the `v2client` branch where we can do more experiments and integration with atlas setup command.

internal/store/organizations.go
gssbzn: Projects are different between ops manager and atlas, here's what we've done bere https://github.com/mongodb/mongodb-atlas-cli/blob/17c025603e81bed15935b3b55c23ed7bc2b89951/internal/store/projects.go, not saying this is the best way to do it but it's an example, feel free to be opionated here but some word of advice

1. won't work becuase there are cases (like projects) where the structs are different between ops manager and atlas, different fields and options

2. not a fan since removing v1 will be harder but open to it

3. This is what we've done before

4. this may add some overhead, the store was built for the same case we are disucssing here having a way to let commands call the "store" and nto have to worry if it's ops manager or atlas

wtrocki: I was considering option no3 or no4. It seems that we both agree with 3 is the way to go (and it is implemented in this PR already.

wtrocki: As variation of the no4 we can use generics or have 2 conditional return types. 
I did play with a generic-based store and I found it quite tricky.
Finally returning 2 different types depending on the cli seemed like the best choice:

- it doesn't require casting 
- we can instantly look for the type we need.

wtrocki:
To summarize:

1. In this case I think we will return interface{} as we do not process object after response. 
```
Organization(string) (*interface{}, error)
```
2. For cases where we need to read and process files we should return multiple objects for convenience:
```
	Organization(string) (*atlasv2.ApiOrganizationView, *atlas.Organization, error)
```

andreaangiolillo (approved):
LGTM. Thanks for addressing all the comments. 💯

gssbzn (approved):
LGTM

gssbzn:
@wtrocki before merging can you please run a manual patch with the e2e test for organizations, evergreen doesn't track feature branches

internal/store/organizations.go
gssbzn: generics are new and I haven't had a time to play much with them so no strong opinions on using them

wtrocki:
@gssbzn for some reason integration tests fails on the evergreen:
https://spruce.mongodb.com/task/mongocli_master_e2e_generic_atlas_iam_e2e_patch_a6661aede4821881111b51a9c5f0d921e7d000c6_63f4b78930661548fc7ef1f1_23_02_21_12_25_59/tests?execution=0&sortBy=STATUS&sortDir=ASC


Running command locally (also tests) seems to work. I have created separate issue to take look into this while working on the specification.

internal/store/organizations.go
```diff
@@ -29,7 +29,7 @@ type OrganizationLister interface {
 }
 
 type OrganizationDescriber interface {
-	Organization(string) (*atlas.Organization, error)
+	Organization(string) (interface{}, error)
```
fmenezes, line 32: I understand the need for `interface{}` here, but this to me feels like splitting the stores is the right way of approaching 💔

internal/store/organizations.go
fmenezes: oh man, this is really hard to follow 💔

wtrocki:
E2E passing now: https://spruce.mongodb.com/task/mongocli_master_e2e_generic_atlas_iam_e2e_patch_a6661aede4821881111b51a9c5f0d921e7d000c6_63f4dc383627e0388b6df10c_23_02_21_14_59_21/logs?execution=0

internal/store/store.go
```diff
@@ -316,8 +316,10 @@ func (s *Store) createV2Client(client *http.Client) error {
 		atlasv2.UseDebug(log.IsDebugLevel())}
 
 	if s.baseURL != "" {
-		opts = append(opts, atlasv2.UseBaseURL(s.baseURL))
+		urlWithoutSuffix := strings.TrimSuffix(s.baseURL, "/")
+		opts = append(opts, atlasv2.UseBaseURL(urlWithoutSuffix))
```
gssbzn, line 320: can this be handled by the SDK, for example this is alreayd done by the manually generated one (tho in the inverse order or adding the missing slash)
https://github.com/mongodb/go-client-mongodb-atlas/blob/01d7604ca665b81f12e797bc8976ec1d76b0449d/mongodbatlas/mongodbatlas.go#L419-L421

wtrocki: Sure

wtrocki: https://github.com/mongodb/go-client-mongodb-atlas/pull/385/files

wtrocki: Reverted current code.

gssbzn: approved there

gssbzn (approved):

wtrocki:
Huge thanks to the team for the amazing review and support! 
image

internal/store/organizations.go
colm-quinn: @fmenezes what improvements would you suggest here? We are the first users of the SDK and have a great chance to improve it before impacting our customers.

I'd recommend a cross team conversation here to apply what we're learning and improve the SDK for everyone.

PR #1692, opened 2023-02-20
CLOUDP-162738: Decouple telemetry and the global config

gssbzn:
## Proposed changes

_Jira ticket:_ CLOUDP-162738

## Further comments

telemetry current depends on the global state of the config, introducing some risks like what I've seeing of unit tests deleting config files, this PR decouples the telemtry options from the global state

I've manually tested telemetry still works as expected

### DIFF
--- a/internal/telemetry/event.go
+++ b/internal/telemetry/event.go
@@ -17,7 +17,7 @@ package telemetry
 import (
 	"crypto/sha256"
 	"encoding/base64"
-	"os"
+	"io"
 	"regexp"
 	"runtime"
 	"strings"
@@ -56,9 +56,13 @@ func withHelpCommand(cmd *cobra.Command, args []string) eventOpt {
 	}
 }
 
-func withProfile() eventOpt { // either "default" or base64 hash
+type ConfigNameGetter interface {
+	Name() string
+}
+
+func withProfile(c ConfigNameGetter) eventOpt { // either "default" or base64 hash
 	return func(event Event) {
-		if config.Name() == config.DefaultProfile {
+		if c.Name() == config.DefaultProfile {
 			event.Properties["profile"] = config.DefaultProfile
 			return
 		}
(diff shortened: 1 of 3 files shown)


internal/telemetry/event_test.go
```diff
@@ -452,3 +373,41 @@ func TestWithHelpCommand_NotFound(t *testing.T) {
 	_, ok := e.Properties["help_command"]
 	assert.False(t, ok)
 }
+
+type configMock struct {
```
gssbzn, line 377: created a common mock object but happy to split if necessary

internal/telemetry/event.go
```diff
@@ -105,7 +109,12 @@ func withEmpty(e bool) eventOpt {
 	}
 }
 
-func withCommandPath(cmd *cobra.Command) eventOpt {
+type CmdName interface {
```
wtrocki, line 112: golangcilint rule - every interface should have comment.
Interfaces should be declared on the top of the file etc.

internal/telemetry/event.go
```diff
@@ -82,13 +86,13 @@ func withChoice(c string) eventOpt {
...
 func sanitizeSelectOption(v string) string {
-	parenthesesRegex := regexp.MustCompile(`^.*\(([^\(\)]*)\)$`)
+	parenthesesRegex := regexp.MustCompile(`^.*\(([^()]*)\)$`)
 
 	return parenthesesRegex.ReplaceAllString(v, "$1")
 }
 
 func sanitizePrompt(q string) string {
-	bracketsRegex := regexp.MustCompile(`\[[^\]\[]*\]`)
+	bracketsRegex := regexp.MustCompile(`\[[^]\[]*]`)
```
wtrocki, line 95: unrelated change?

gssbzn: drive by, intellij was complaining of unnecessary scape

internal/telemetry/event.go
gssbzn: the comment rule is disable by default, if we were to enable it it may be a large backfill 
interfaces at the top  do you know the rule for this?, personally I like declare things as close as possible to where they are needed

colm-quinn (approved):
LGTM! Thanks for breaking these out for mocking

wtrocki (approved):

PR #1693, opened 2023-02-20
CLOUDP-162861: Remove unnnecessary new constructors from the auth pkg

gssbzn:
## Proposed changes

_Jira ticket:_ CLOUDP-162861

Trying to clean up auth login/register and setup is becoming too complected due to a big amount of technical debt and inter-dependency, starting small with a small clean up of unnecessary methods that expose a lot of the issues I'm seeing

I'll pause trying to fix the unit tests deleting config files after this PR since it's becoming bigger and I have other tickets to look at

### DIFF
--- a/internal/cli/atlas/setup/setup_cmd.go
+++ b/internal/cli/atlas/setup/setup_cmd.go
@@ -83,7 +83,7 @@ This command will help you
 	return opts.quickstart.Run()
 }
 
-func (opts *Opts) PreRun(ctx context.Context) error {
+func (opts *Opts) PreRun(ctx context.Context) {
 	opts.skipRegister = true
 	opts.skipLogin = true
 
@@ -109,8 +109,6 @@ func (opts *Opts) PreRun(ctx context.Context) error {
 		opts.skipRegister = false
 	default:
 	}
-
-	return nil
 }
 
 // Builder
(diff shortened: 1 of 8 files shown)


internal/cli/atlas/setup/setup_cmd.go
```diff
@@ -83,7 +83,7 @@ This command will help you
 	return opts.quickstart.Run()
 }
 
-func (opts *Opts) PreRun(ctx context.Context) error {
```
gssbzn, line 86: never returns an error

internal/cli/atlas/setup/setup_cmd.go
```diff
@@ -129,8 +127,8 @@ func Builder() *cobra.Command {
 	qsOpts.LabelValue = labelValue
 	opts := &Opts{
 		register:   auth.NewRegisterFlow(loginOpts),
-		login:      auth.NewLoginFlow(loginOpts),
-		quickstart: quickstart.NewQuickstartFlow(qsOpts),
+		login:      loginOpts,
+		quickstart: qsOpts,
```
gssbzn, line 131: quick-start has 2 logins, bringing the count o 3

internal/cli/atlas/setup/setup_cmd_test.go
```diff
@@ -168,7 +168,8 @@ func Test_setupOpts_RunSkipRegister(t *testing.T) {
 		Return(nil).
 		Times(1)
 
-	require.NoError(t, opts.PreRun(ctx))
+	// FIXME: CLOUDP-162860 side effect dependant test
```
gssbzn, line 171: this test is for run but somehow depends on prerun having been called

internal/cli/auth/login.go
```diff
@@ -48,12 +48,12 @@ type LoginConfig interface {
 }
 
 const (
-	AlreadyAuthenticatedError      = "you are already authenticated with an API key (Public key: %s)"
+	alreadyAuthenticatedError      = "you are already authenticated with an API key (Public key: %s)"
```
gssbzn, line 51: reducing visibility, I have a follow up task to clean up these const seems they are only used in 1 place but it requires some changes to tests

wtrocki (approved):

PR #1700, opened 2023-02-22
CLOUDP-140450: [AtlasCLI] Investigate chocolatey checksum failure

gssbzn:
## Proposed changes

_Jira ticket:_ CLOUDP-140450

## Further comments

Change to do the checksum of the local file instead of trying to download from the download center, this makes sene to make sure the file hasn't been compromise between generation and publishing and also to avoid the current behaviour where we are doing the download before the file is available

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -5,7 +5,7 @@ COVERAGE=coverage.out
 
 MCLI_SOURCE_FILES?=./cmd/mongocli
 MCLI_BINARY_NAME=mongocli
-MCLI_VERSION?=$(shell git tag --list 'mongocli/v*' --sort=taggerdate | tail -1 | cut -d "v" -f 2 )
+MCLI_VERSION?=$(shell git describe --match "mongocli/v*" | cut -d "v" -f 2)
 MCLI_GIT_SHA?=$(shell git rev-parse HEAD)
 MCLI_DESTINATION=./bin/$(MCLI_BINARY_NAME)
 MCLI_INSTALL_PATH="${GOPATH}/bin/$(MCLI_BINARY_NAME)"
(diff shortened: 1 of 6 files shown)


Makefile
```diff
@@ -5,24 +5,18 @@ COVERAGE=coverage.out
 
 MCLI_SOURCE_FILES?=./cmd/mongocli
 MCLI_BINARY_NAME=mongocli
-MCLI_VERSION?=$(shell git tag --list 'mongocli/v*' --sort=taggerdate | tail -1 | cut -d "v" -f 2 )
+MCLI_VERSION?=$(shell git describe --match "mongocli/v*" | cut -d "v" -f 2)
```
gssbzn, line 8:
```bash
./bin/atlas --version
atlascli version: 1.5.1-18-g4f4ba197
git version: 4f4ba197e78c52fab065f02e58e08d63b6613ad0
Go version: go1.19
   os: darwin
   arch: arm64
   compiler: gc
```

vs
```
./bin/atlas --version
atlascli version: 1.5.1-next
git version: 4f4ba197e78c52fab065f02e58e08d63b6613ad0
Go version: go1.19
   os: darwin
   arch: arm64
   compiler: gc
```

build/package/generate-msi.sh
```diff
@@ -23,32 +23,31 @@ export NOTARY_SIGNING_KEY
 
 go-msi check-env
 
-VERSION="$(git tag --list "${TOOL_NAME:?}/v*" --sort=taggerdate | tail -1 | cut -d "v" -f 2)"
-
+VERSION_GIT="$(git describe --match "${TOOL_NAME:?}/v*" | cut -d "v" -f 2)"
```
gssbzn, line 26: bringing back a version that includes the state of the branch so uploaded nuspec can be distinguished from production releases

tools/chocolateypkg/chocolateypkg.go
```diff
@@ -94,33 +93,37 @@ func replaceNuspec(dir, version string) error {
...
-func replaceInstallScript(dir, url string) error {
-	checkSum, err := generateSha256(url)
+func generateInstallScript(dir, version string) error {
+	file := fmt.Sprintf("mongodb-atlas-cli_%s_windows_x86_64.msi", version)
+	checkSum, err := sha256sum(path.Join(dir, file))
 	if err != nil {
 		return err
 	}
+	// For previews this link will be invalid
+	url := fmt.Sprintf("https://fastdl.mongodb.org/mongocli/mongodb-atlas-cli_%s_windows_x86_64.msi", version)
```
gssbzn, line 125: do we care? I tried to have point to our internal previews but it becomes complex as I need access to more evg variables for this to work

fmenezes: I personally think we should not bother about this on preview as we do/should not publish to chocolatey on non final versions.

fmenezes (approved):

PR #1701, opened 2023-02-22
CLOUDP-153757: [Atlas - CLI] Watch for a given cluster state

gssbzn:
## Proposed changes

_Jira ticket:_ CLOUDP-153757

## Checklist

### DIFF
--- a/docs/atlascli/command/atlas-clusters-watch.txt
+++ b/docs/atlascli/command/atlas-clusters-watch.txt
@@ -55,6 +55,10 @@ Options
      - Type
      - Required
      - Description
+   * - --goal
+     - strings
+     - false
+     - Goal state to watch for. This value defaults to [IDLE].
    * - -h, --help
      - 
      - false
--- a/docs/mongocli/command/mongocli-atlas-clusters-watch.txt
+++ b/docs/mongocli/command/mongocli-atlas-clusters-watch.txt
@@ -55,6 +55,10 @@ Options
      - Type
      - Required
      - Description
+   * - --goal
+     - strings
+     - false
+     - Goal state to watch for. This value defaults to [IDLE].
    * - -h, --help
      - 
      - false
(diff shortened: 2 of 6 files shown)


andreaangiolillo:
- How does the command behave when waiting for `DELETING`?
- Are customers asking for this feature? I don't see the use case beside IDLE and DELETING. I would update the command to wait only for IDLE or DELETION instead of a goal flag.

gssbzn:
>How does the command behave when waiting for DELETING?

DELETION will never be seen as it'll hit a 404 first

>Are customers asking for this feature?
🤷

AndreiMatei2206:
For the `DELETED`, in the docs it is one of the states that can identify a cluster 😶‍🌫️

> Enum: "IDLE", "CREATING", "UPDATING", "DELETING", "DELETED", "REPAIRING"

So I was thinking that we could maybe define "custom" states, i.e. for "DELETED" case we could watch for "DELETING" and then the 404 error, or for "UPDATED" case we could watch for "UPDATING" and then "IDLE".

gssbzn:
> So I was thinking that we could maybe define "custom" states, i.e. for "DELETED" case we could watch for "DELETING" and then the 404 error, or for "UPDATED" case we could watch for "UPDATING" and then "IDLE".

To echo Andrea's question

>Are customers asking for this feature?

What's the use case of this feature, having a hard time to understand the benefit of these custom states, 

This comment in particular 

>  or for "UPDATED" case we could watch for "UPDATING" and then "IDLE"

Seems like you want to watch it reached updating, then idle in that order, does this matter? why does it matter in the particular order?

can't it be solve by 
```
atlas clusters watch myRS --goal UPDATING &&  atlas clusters watch myRS --goal IDDLE
```

AndreiMatei2206:
I don't think customers are asking for this feature, I saw some hacky sleeps being done as part of CFN to wait for a cluster to be deleted (https://github.com/mongodb/mongodbatlas-cloudformation-resources/pull/182#discussion_r1063408171) - so watching for a particular state seemed like a nice addition to the this command.

For there is no "UPDATED" state defined in the API, so I was thinking that having that abstraction (which can be solved exactly how you mentioned) would be useful. The order of states would be useful for the command to not execute indefinitely. The primary use case I had in mind was were short feedback loops like:

- User makes action on cluster (this will put the cluster into a `{verb}ING` state)
- User uses watch to wait until that action is done on the cluster (until the state of the cluster reaches `IDLE` again)

gssbzn:
Closing this for now @AndreiMatei2206 to drive the conversation on the feature

PR #1703, opened 2023-02-23
CLOUDP-162860: Review and fix interdependencies between login/register/setup/quickstart

gssbzn:
## Proposed changes

_Jira ticket:_ CLOUDP-162860

## Further comments
This PR is probably messier than I would've liked but the inter dependencies between  login/register/setup/quickstart  were complex and hard to fix without pulling the commands apart which makes the changes harder to isolate (a small change in one had big changes is another command)

The main goal of this PR has been to remove dependency on some global states like config allowing it to be injected and mocked in more places

Another big change is flattening some of embedded structs to allow easier reuse and less surprises in the future like including login twice in the same command

## Manual QA Performed

- [x] Log in
- [x] Log in when already logged in
- [x] Log in when already logged in (expired credentials)
- [x] Log in with API keys in profile 
- [x] Log in let code expire and get confirmation to regenerate 
- [x] Register
- [x] Register when already logged in 
- [x] Register when already logged in  (expired credentials) **change here this is now allowed**
- [x] Register with API keys in profile
- [x] Quickstart
- [x] Quickstart logged in
- [x] Quickstart with API keys in profile
- [x] Quickstart register
- [x] Quickstart with --force
- [x] Set up 
- [x] Set up logged in
- [x] Set up  with API keys in profile
- [x] Set up register
- [x] Set up with --force
- [x] Set up logged in and --force
- [x] Set up  with API keys in profile --force


wtrocki (approved):

internal/cli/auth/login.go
```diff
@@ -171,16 +122,16 @@ func (opts *LoginOpts) Run(ctx context.Context) error {
...
 }
 
-func (opts *LoginOpts) CheckProfile(ctx context.Context) error {
+func (opts *LoginOpts) checkProfile(ctx context.Context) error {
 	if err := opts.InitStore(ctx); err != nil {
 		return err
 	}
 	if config.OrgID() != "" && !opts.OrgExists(config.OrgID()) {
-		config.SetOrgID("")
+		opts.config.Set("org_id", "")
```
tibulca, line 130: can we use the config costs from [internal/config/profile.go](https://github.com/mongodb/mongodb-atlas-cli/blob/cf90d8088fefddb5f6f979d8cbfe7932993b79be/internal/config/profile.go#L49)?

tibulca (approved):
LGTM, great work untangling these flows and making things more clear 👌

internal/cli/auth/login.go
```diff
@@ -291,43 +241,35 @@ func (opts *LoginOpts) oauthFlow(ctx context.Context) error {
...
+		opts.config = c
+		if config.OpsManagerURL() != "" {
+			opts.OpsManagerURL = config.OpsManagerURL()
+		}
+		// ignore expired tokens since logging in
+		if err := opts.RefreshAccessToken(ctx); err != nil && !errors.Is(err, cli.ErrInvalidRefreshToken) {
+			return err
+		}
+		// clean up any expired tokens
+		opts.config.Set(config.AccessTokenField, "")
```
tibulca, line 270: this happens before we run validate.NoAccessToken validation (the new line 297) so the user will be allowed to re-login without doing first an `auth logout`. Is this intentional?

```
opts.InitFlow(config.Default()),
				opts.LoginPreRun(cmd.Context(), config.Default()),
				validate.NoAPIKeys,
				validate.NoAccessToken,
```

PR #1708, opened 2023-02-27
CLOUDP-150028: atlas setup command should detect existence of cluster

wtrocki:
WIP for early feedback about architecture changes.

# Description

When running `atlas setup` users should be notified about having existing clusters in the atlas.
Some of the clusters might be limited (M0 on Atlas etc.) so we need to notify about issues

## API 

As for now, API doesn't give us the ability to search for clusters based on `tier`. 
That is why I have decided to settle with a warning for users having existing clusters (any existing ones) and asking them if they would like to continue. Alternatively, we could read the total values and traverse thru all pages for lookout for specific cluster type.  However this approach will be bundling an dynamic business logic (one m0 cluster possible) into CLI which is not great from update point of view.

### DIFF
--- a/internal/cli/atlas/clusters/list.go
+++ b/internal/cli/atlas/clusters/list.go
@@ -48,12 +48,12 @@ var listTemplate = `ID	NAME	MDB VER	STATE{{range .Results}}
 
 func (opts *ListOpts) Run() error {
 	listOpts := opts.NewListOptions()
-	r, err := opts.store.ProjectClusters(opts.ConfigProjectID(), listOpts)
+	r1, r2, err := opts.store.ProjectClusters(opts.ConfigProjectID(), listOpts)
 	if err != nil {
 		return err
 	}
 
-	return opts.Print(r)
+	return opts.PrintFirst(r1, r2)
 }
 
 // mongocli atlas cluster(s) list --projectId projectId [--page N] [--limit N].
--- a/internal/cli/atlas/quickstart/quick_start.go
+++ b/internal/cli/atlas/quickstart/quick_start.go
@@ -244,6 +244,15 @@ func (opts *Opts) Run() error {
 		return dErr
 	}
 
+	r1, r2, err := opts.store.ProjectClusters(opts.ConfigProjectID(), &atlas.ListOptions{
+		PageNum: 1, ItemsPerPage: 1, IncludeCount: true})
+	if err != nil {
+		return err
+	}
+	if (r1 != nil && r1.TotalCount > 0) || (r2 != nil && r2.TotalCount > 0) {
+		fmt.Print("TODO: warn users about account having multiple clusters ")
+	}
+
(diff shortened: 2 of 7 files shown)


internal/cli/output_opts.go
```diff
@@ -106,6 +106,16 @@ func (opts *OutputOpts) Print(o interface{}) error {
 	return err
 }
 
+// Print first non nil object from objectsToPrint array.
+func (opts *OutputOpts) PrintFirst(objectsToPrint ...interface{}) error {
```
wtrocki, line 110: A new method that will enable us to work with multiple objects returned by store API.

internal/cli/atlas/clusters/list.go
```diff
@@ -48,12 +48,12 @@ var listTemplate = `ID	NAME	MDB VER	STATE{{range .Results}}
...
 func (opts *ListOpts) Run() error {
 	listOpts := opts.NewListOptions()
-	r, err := opts.store.ProjectClusters(opts.ConfigProjectID(), listOpts)
+	r1, r2, err := opts.store.ProjectClusters(opts.ConfigProjectID(), listOpts)
 	if err != nil {
 		return err
 	}
 
-	return opts.Print(r)
+	return opts.PrintFirst(r1, r2)
```
wtrocki, line 56: How printing will work in this scenario

internal/cli/atlas/quickstart/quick_start.go
```diff
@@ -244,6 +244,15 @@ func (opts *Opts) Run() error {
 		return dErr
 	}
 
+	r1, r2, err := opts.store.ProjectClusters(opts.ConfigProjectID(), &atlas.ListOptions{
+		PageNum: 1, ItemsPerPage: 1, IncludeCount: true})
+	if err != nil {
+		return err
+	}
+	if (r1 != nil && r1.TotalCount > 0) || (r2 != nil && r2.TotalCount > 0) {
```
wtrocki, line 252: can be later replaced by the generics helper method as well.

internal/cli/atlas/quickstart/quick_start.go
```diff
@@ -244,6 +244,15 @@ func (opts *Opts) Run() error {
 		return dErr
 	}
 
+	r1, r2, err := opts.store.ProjectClusters(opts.ConfigProjectID(), &atlas.ListOptions{
```
wtrocki, line 247: We do not use `setup` for non atlas case for this command.
I guess we can even create new interface that returns only Atlas API?

gssbzn: https://github.com/mongodb/mongodb-atlas-cli/blob/04845ef93e04e42bc1327e4d5a1253b6a58cab96/internal/cli/opsmanager/clusters/list.go#L75

wtrocki:
This was experiment command. Since then we moved to different approach of handling store structures. Closing

PR #1729, opened 2023-03-06
task: remove legacy usages of +build tags

gssbzn:
`+ build` has been replaced by `//build` since go 1.18, intellij now flags when using both to remove the legacy annotation, this is a basic find and replace to make intellij visual linter nicer


gssbzn:
evergreen refresh

wtrocki (approved):

PR #1732, opened 2023-03-06
CLOUDP-165331: [cli] update to github.com/aws/aws-sdk-go-v2

gssbzn:
## Proposed changes

_Jira ticket:_ CLOUDP-165331

## Further comments

Trying to upgrade to the new aws sdk lib, the new lib is modularized which should help reduce binary size, currently it manages to reduce ~3MB from the final binary 
```
#before
51M  2 Mar 15:55 atlas
#after
48M  6 Mar 15:19 atlas
```

[e2e test](https://spruce.mongodb.com/version/6408566f0305b9049e114d73/tasks?page=0&sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC&variant=%5Ee2e_decryption%24) (don't review if not passing)


internal/decryption/keyproviders/aws.go
```diff
@@ -37,69 +37,35 @@ type AWSKeyIdentifier struct {
 	SecretAccessKey string
 	SessionToken    string
 
-	credentials *credentials.Credentials
+	cfg aws.Config
```
gssbzn, line 40: credential is no longer a thing, it seems closer to how we did [gcp](https://github.com/mongodb/mongodb-atlas-cli/blob/dc9e02e34c51ad670b21d83ca97d73836034f19c/internal/decryption/keyproviders/gcp.go) so following that pattern

internal/decryption/keyproviders/key_provider.go
```diff
@@ -41,15 +40,15 @@ type KeyStoreIdentifier struct {
 	Provider KeyStoreProvider
 }
 
-func provideInput(prompt, defaultInput string) (string, error) {
-	if strings.TrimSpace(strings.ToLower(os.Getenv("CI"))) == "true" {
-		return defaultInput, nil
+func provideInput(prompt string) (string, error) {
```
gssbzn, line 43: the linter somehow started to complain `defaultInput` was always `""` so removing as an arg

internal/decryption/keyproviders/key_provider.go
```diff
@@ -41,15 +40,15 @@ type KeyStoreIdentifier struct {
...
+	if _, ok := os.LookupEnv("CI"); ok {
```
gssbzn, line 44: normally you don't care the value, just if set

internal/decryption/keyproviders/aws.go
fmenezes: the thing here is in GCP there is no way to set secrets by hand not even in their cli, you're obliged to get a secret and provide it on a config I my memory is not completely bananas

fmenezes: For AWS we normally have the API KEY / API SECRET KEY to input via parameters on their CLI, I imagine that you may be able to do that on the new SDK

fmenezes:
code changes look good, just waiting on my previous point

internal/decryption/keyproviders/aws.go
fmenezes: looking at their SDK I've found this https://github.com/aws/aws-sdk-go-v2/blob/main/credentials/static_provider.go#L29-L53 I think we would need to check and ask for credentials

gssbzn: I was trying to follow https://aws.github.io/aws-sdk-go-v2/docs/migrating/#client-construction, I think you did catch a case of provided creds as the new `config.LoadDefaultConfig` replaces the both `credentials.EnvProvider`, `credentials.SharedCredentialsProvider` and given the test was passing I called it a day, I suppose the test is then using env variables(?)

The comment here on gcp approach is mostly because for gcp we do the same on saving the conf and not the creds, and we also have the issue on the context not being passed but using the background context

gssbzn: I added the static credential provider to config.LoadDefaultConfig, which hopefully addresses your comment

fmenezes: 👍

fmenezes (dismissed):

fmenezes:
looks like tests are failing, perhaps check if the variables are filled first

internal/decryption/keyproviders/aws.go
gssbzn: this actually doesn't work the same anymore and it won't chain the providers being a last write win, seems like the new SDK doesn't support that feature anymore so holding on this PR for now

fmenezes: I've been rereading this PR I think the only thing we have to do is check if the three variables are filled if so override the default, otherwise, we should commit the `credentials.SharedCredentialsProvider`

internal/cli/decryption/key_providers.go
```diff
@@ -22,6 +22,7 @@ import (
 func KeyProvidersBuilder() *cobra.Command {
 	cmd := &cobra.Command{
 		Use:     "keyProviders",
+		Hidden:  true,
```
fmenezes: This command has to stay on for OM, it lists headers for log files

internal/cli/atlas/logs/decrypt.go
```diff
@@ -96,8 +96,9 @@ func DecryptBuilder() *cobra.Command {
 	opts := &DecryptOpts{}
 	opts.Fs = afero.NewOsFs()
 	cmd := &cobra.Command{
-		Use:   "decrypt",
-		Short: "Decrypts an audit log file with the provided AWS, GCP or Azure key management services.",
+		Use:    "decrypt",
+		Hidden: true,
```
fmenezes, line 100: 👍

internal/cli/decryption/key_providers.go
fmenezes: nvm you've fixed it

internal/decryption/keyproviders/aws.go
gssbzn: I think this should be now working as it was before

fmenezes (approved):
thank you for the changes, great work here

PR #1754, opened 2023-03-10
CLOUDP-161495: disable bodyclose rule

wtrocki:
See: https://jira.mongodb.org/browse/CLOUDP-161495

TL:DR - there are so many false positives for bodyclose that this rule only works when stream is opened only within the same golang scope (same code block). Any functions, or external library response objects are reported as errors.
Considering the very poor maintenance of the rule (those issues were reported years ago) I recomend to disable rule rather than adding +400 exceptions. 

This PR is an alternative to current approach:
Supplying extra defer for all sdk calls (which would result in noop as body is closed already)

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -64,7 +64,6 @@ linters-settings:
 linters:
   disable-all: true
   enable:
-    - bodyclose # checks whether HTTP response body is closed successfully [fast: false, auto-fix: false]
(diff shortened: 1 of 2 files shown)


matt-condon (approved):
SGTM

wtrocki:
@andreaangiolillo @fmenezes Looking for another opinion before we merge that change.

fmenezes (approved):
LGTM

PR #1764, opened 2023-03-14
Docs: contributing guide for e2e tests

wtrocki:
Instruction on how to debug e2e tests.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -141,7 +141,7 @@ To debug e2e tests we can use following run configuration
 
 In following snippet:
 1. Adjust program to the test you want to run
-2. Review MCLI_E2E_BINARY to point to mongodbcli or atlas depending on what CLI you want to test.
+2. Review `MCLI_E2E_BINARY` to point to `mongodbcli` or `atlas` depending on what CLI e2e tests should use.
 3. Run `make build-all` to ensure that latest binary files are used in tests.
 
 > NOTE: Debugging works only for e2e automation. Commands cannot be debugged during that process. 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -139,8 +139,12 @@ To debug e2e tests we can use following run configuration
       } 
 ```
 
-Before running command makesure you run `make build-all` and adjusted program to run your desired e2e test
+In following snippet:
+1. Adjust program to the test you want to run
+2. Review MCLI_E2E_BINARY to point to mongodbcli or atlas depending on what CLI you want to test.
+3. Run `make build-all` to ensure that latest binary files are used in tests.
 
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -86,6 +86,27 @@ The following is a short list of commands that can be run in the root of the pro
...
+
+          "request": "launch",
+          "mode": "auto",
+          "program": "${workspaceRoot}/test/e2e/iam/project_api_keys_test.go",
+          "showLog": true,
+          "env": {},
+          "buildFlags": "-tags e2e"
+      } 
+    ]
+```
```
gssbzn, line 108: shouldn't this be added to the existing `#### Debugging in VSCode` section?

swattyT (dismissed):
Thanks for adding!

CONTRIBUTING.md
wtrocki: I thought about it actually and since you mentioned it I will move it now.

gssbzn (approved):

CONTRIBUTING.md
```diff
@@ -141,7 +141,7 @@ To debug e2e tests we can use following run configuration
 
 In following snippet:
 1. Adjust program to the test you want to run
-2. Review MCLI_E2E_BINARY to point to mongodbcli or atlas depending on what CLI you want to test.
```
gssbzn: https://github.com/mongodb/mongodb-atlas-cli/blob/35ee197eda86bb657cd0dc9c7fac4c14aec42088/test/e2e/cli.go#L25-L49

Also pointing MCLI_E2E_BINARY to the atlas bin may be dangerous for iam commands as they will assume an `iam` prefix alas `mongocli iam projects` where for atlas it'll expect `atlas projects`

wtrocki: Clarified that. We have learned that the hard way yesterday :)

gssbzn: it shouldn't matter at come point I considered making `cli.Bin()` env aware but then I hit the iam wall making it not a straight forward change

At the same time in the current state of the codebase it doesn't matter which bin you are using because the node (the cobra command) will be the same for both binaries and we get away with it

If using the `mongocli` bin for the SDK migration is a problem then for the project the comamnds will have to be duplicated, one set for atlas bin and one set for mongocli

wtrocki: > If using the mongocli bin for the SDK migration is a problem then for the project the commands will have to be duplicated, one set for atlas bin and one set for mongocli

CC @swattyT @matt-condon - very important comment.

gssbzn: and to be clear from where I see it using the `mongocli` bin should be a non-issue, it's just an entry point and it makes no difference if it's called `mongocli`, `atlas` or `cly-mccliface`, at the end it calls the same logiic

PR #1770, opened 2023-03-16
CLOUDP-164969: Migrate Atlas CLI Organizations Client to v2 SDK Client

matt-condon:
## Proposed changes

Updating the migration of Atlas CLI Organizations commands to the auto-generated v2 atlas SDK

Proposing an alternative approach to the store re-architecture, mapping new v2 client struct properties to the existing structs that are returned. 

This should:
- address backwards compatibility concerns
- mean changes to incorporate the new client version are effectively all done in one place
- allow us to validate with the existing e2e tests
  - in returning `interface{}`, we will need some rearchitecture of e2e tests or breaking template change

_Jira ticket:_ CLOUDP-164969

Closes #[issue number]

## Checklist

### DIFF
--- a/internal/cli/default_setter_opts.go
+++ b/internal/cli/default_setter_opts.go
@@ -36,7 +36,7 @@ import (
 type ProjectOrgsLister interface {
 	Project(id string) (interface{}, error)
 	Projects(*atlas.ListOptions) (interface{}, error)
-	Organization(id string) (interface{}, error)
+	Organization(id string) (*atlas.Organization, error)
 	Organizations(*atlas.OrganizationsListOptions) (*atlas.Organizations, error)
 	GetOrgProjects(string, *atlas.ProjectsListOptions) (interface{}, error)
 }
--- a/internal/store/organizations.go
+++ b/internal/store/organizations.go
@@ -19,6 +19,7 @@ import (
 
 	"github.com/mongodb/mongodb-atlas-cli/internal/config"
 	atlas "go.mongodb.org/atlas/mongodbatlas"
+	atlasv2 "go.mongodb.org/atlas/mongodbatlasv2"
 	"go.mongodb.org/ops-manager/opsmngr"
 )
 
@@ -29,7 +30,7 @@ type OrganizationLister interface {
 }
 
 type OrganizationDescriber interface {
-	Organization(string) (interface{}, error)
+	Organization(string) (*atlas.Organization, error)
 }
 
 type OrganizationCreator interface {
@@ -44,6 +45,7 @@ type OrganizationDeleter interface {
 func (s *Store) Organizations(opts *atlas.OrganizationsListOptions) (*atlas.Organizations, error) {
 	switch s.service {
 	case config.CloudService, config.CloudGovService:
+		//TODO: Migrate once OrganizationsListOptions.IncludeDeletedOrgs property is generated
 		result, _, err := s.client.(*atlas.Client).Organizations.List(s.ctx, opts)
 		return result, err
 	case config.CloudManagerService, config.OpsManagerService:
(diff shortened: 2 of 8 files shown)


internal/store/organizations.go
```diff
@@ -55,11 +57,12 @@ func (s *Store) Organizations(opts *atlas.OrganizationsListOptions) (*atlas.Orga
 }
 
 // Organization encapsulate the logic to manage different cloud providers.
-func (s *Store) Organization(id string) (interface{}, error) {
+func (s *Store) Organization(id string) (*atlas.Organization, error) {
 	switch s.service {
 	case config.CloudService, config.CloudGovService:
 		result, _, err := s.clientv2.OrganizationsApi.GetOrganization(s.ctx, id).Execute()
-		return result, err
+		newOrg := atlas.Organization{ID: *result.Id, IsDeleted: result.IsDeleted, Name: result.Name, Links: mapLinks(result.Links)}
```
blva, line 64: [q] what happens to the result if the endpoint throws an error? (e.g. 404)  might we get some panic error trying to create the org object here?

internal/store/organizations.go
```diff
@@ -44,6 +45,7 @@ type OrganizationDeleter interface {
 func (s *Store) Organizations(opts *atlas.OrganizationsListOptions) (*atlas.Organizations, error) {
 	switch s.service {
 	case config.CloudService, config.CloudGovService:
+		//TODO: Migrate once OrganizationsListOptions.IncludeDeletedOrgs property is generated
```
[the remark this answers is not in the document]
wtrocki: This will never be generated. It is typical ops manager property

internal/store/organizations.go
matt-condon: Good point, thanks!
Updated with some error/nil handling (and moved to a separate function)

internal/store/organizations.go
```diff
@@ -93,3 +99,27 @@ func (s *Store) DeleteOrganization(id string) error {
...
 	}
 }
+
+func mapOrganizationResult(result *atlasv2.Organization, err error) (*atlas.Organization, error) {
+	if err != nil {
+		return nil, err
+	}
+	if result == nil {
+		return nil, errors.New("GetOrganization response empty")
+	}
```
andreaangiolillo, line 109: in this case you would get a 404 error so I believe you don't need this

internal/store/organizations.go
```diff
@@ -55,11 +58,12 @@ func (s *Store) Organizations(opts *atlas.OrganizationsListOptions) (*atlas.Orga
 }
 
 // Organization encapsulate the logic to manage different cloud providers.
-func (s *Store) Organization(id string) (interface{}, error) {
+func (s *Store) Organization(id string) (*atlas.Organization, error) {
 	switch s.service {
 	case config.CloudService, config.CloudGovService:
 		result, _, err := s.clientv2.OrganizationsApi.GetOrganization(s.ctx, id).Execute()
-		return result, err
+		newOrg, err := mapOrganizationResult(result, err)
```
andreaangiolillo, line 65:
```suggestion
		return mapOrganizationResult(result, err)
```

wtrocki:
Mapping can get tricky for very large response objects (+200 lines). It would not allow us to migrate most of the methods and usage of v1 - eventually remove v1 client from atlas(which was part of the scope)

This is great step forward thought: 

1. why not return 2 objects (without mapping) and make a decision based on which object is not nil. Pick template etc.
2. If we want to improve on that. I would ask us a question: what is the value of the store at this point if we need to differentiate between objects? do mapping even make sense considering our plans and state of the project?

My personal take that the most basic solution right now that will help you to minimize issues is to
duplicate entire store package (leave current store untouched) and functional for mongodbcli.

Use that store for atlas cli along with the template mapping. 
It is clean, simple and error free solution.

internal/cli/default_setter_opts.go
```diff
@@ -36,7 +36,7 @@ import (
 type ProjectOrgsLister interface {
 	Project(id string) (interface{}, error)
 	Projects(*atlas.ListOptions) (interface{}, error)
-	Organization(id string) (interface{}, error)
+	Organization(id string) (*atlas.Organization, error)
```
andreaangiolillo, line 39: [q] Did we consider renaming the current `Organization` interface to `OrganizationDeprecated` and creating a new `Organization` interface that uses only the SDK v2?

andreaangiolillo (commented):
Left a few comments that are not blocking the PR. Thanks!

PR #1815, opened 2023-03-28
CLOUDP-167804: Migrate serverless instances commands to v2 client

andmatei:
## Proposed changes

Migrate Serverless Instances commands to use v2 client.

_Jira ticket:_ [CLOUDP-167804](https://jira.mongodb.org/browse/CLOUDP-167804)

## Checklist


andmatei:
For the first commit, the build will be failing as `kubernetes/operator` is also using the Store, but is accessing the field `AutoScaling` that is not present in the v2 client.
Checked on the backend model, and this issue is not an OpenAPI annotation issue, but it seems like a bug in operator.

Here is the error from Evergreen:

```
internal/kubernetes/operator/deployment/deployment.go:442:17: deployment.AutoScaling undefined (type *mongodbatlasv2.ServerlessInstanceDescription has no field or method AutoScaling)
```

wtrocki:
the parts that are passed to operator need to be v1 structures. It looks like we do use the same sdk in both and objects are passed between operator and cli. 

We will need to work with Kubernetes team to get models updated later phase. 
For now we need to do v2=>v1 mapping for entire kubernetes suite. 

CC @igor-karpukhin

wtrocki:
As solution, we can duplicate the method used by Kubernetes (leave it untouched) and introduce new one for the CLI itself

internal/store/serverless_instances.go
```diff
@@ -28,7 +29,7 @@ type ServerlessInstanceLister interface {
 }
 
 type ServerlessInstanceDescriber interface {
-	ServerlessInstance(string, string) (*atlas.Cluster, error)
+	ServerlessInstance(string, string) (*atlasv2.ServerlessInstanceDescription, error)
```
wtrocki, line 32:
```suggestion
         ServerlessInstance(string, string) (*atlas.Cluster, error)
	ServerlessInstanceV2(string, string) (*atlasv2.ServerlessInstanceDescription, error)
```

wtrocki:
Rebase needed.

wtrocki:
@andmatei this should be fixed now

internal/cli/mongocli/serverless/create.go
gssbzn, line 1: should we be modifying these? I though the idea of duplicating was leaving these commands alone, no?

internal/store/serverless_instances.go
```diff
@@ -19,42 +19,60 @@ import (
...
 
 // ServerlessInstances encapsulates the logic to manage different cloud providers.
-func (s *Store) ServerlessInstances(projectID string, listOps *atlas.ListOptions) (*atlas.ClustersResponse, error) {
+func (s *Store) ServerlessInstances(projectID string, listOps *atlas.ListOptions) ([]atlasv2.ServerlessInstanceDescription, error) {
 	switch s.service {
 	case config.CloudService:
-		result, _, err := s.client.(*atlas.Client).ServerlessInstances.List(s.ctx, projectID, listOps)
-		return result, err
+		result, _, err := s.clientv2.ServerlessInstancesApi.ListServerlessInstances(s.ctx, projectID).
+			IncludeCount(listOps.IncludeCount).
```
gssbzn, line 53: we don't expose `IncludeCount` in the cli and assume this to always be true, be sure this is not defaulting to false and the behavior of this default is a bit spotty

matt-condon (approved):
LGTM

internal/store/serverless_instances.go
```diff
@@ -19,42 +19,60 @@ import (
...
+			ItemsPerPage(int32(listOps.ItemsPerPage)).
+			PageNum(int32(listOps.PageNum)).
+			Execute()
+
+		return result.Results, err
```
gssbzn, line 58: you are loosing the envelop object here which means I'm pretty sure this doesn't work anymore as the template is not updated, please don't unwrap the envelop object as people use it to get the pagination information

internal/store/serverless_instances.go
```diff
@@ -19,42 +19,60 @@ import (
...
+	switch s.service {
+	case config.CloudService:
+		result, _, err := s.clientv2.ServerlessInstancesApi.GetServerlessInstance(s.ctx, projectID, clusterName).Execute()
+		return result, err
+	default:
+		return nil, fmt.Errorf("%w: %s", errUnsupportedService, s.service)
+	}
+}
+
+// Used by Kubernetes v1 ServerlessInstance encapsulates the logic to manage different cloud providers.
```
gssbzn, line 75: pretty sure there's a ticket to migrate the k8s commands, can you link it?

test/e2e/atlas/serverless_test.go
```diff
@@ -105,12 +105,12 @@ func TestServerless(t *testing.T) {
 		resp, err := cmd.CombinedOutput()
 		req.NoError(err, string(resp))
 
-		var clusters mongodbatlas.ClustersResponse
+		var clusters []atlasv2.ServerlessInstanceDescription
```
gssbzn, line 108: this is a change of behavior and a bad one

gssbzn (changes requested):
mainly requesting changes because of the change of behavior for index

internal/cli/mongocli/serverless/create.go
matt-condon: To achieve that for all cases we'd have to duplicate all the stores too (we only duplicated the functions for the stores that were shared by On-Prem and Atlas)

Since these are Atlas-exclusive endpoints/commands, we are migrating for both sets
(not sure what future plans are for Atlas commands residing under MongoCLI)

internal/store/serverless_instances.go
andmatei: Oopsie - I can revert this. Just a [q] on this, is there any particular reason we are keeping this enveloped and other list responses not?

gssbzn: Well paint me curious because we have been always trying to keep the envelope, only exception there were some really old (before APIx times) sdk methods that were removing it and since the sdk wasn't returning it there was little we could do.

We keep the envelope as it includes the number of pages information which otherwise would be lost

internal/cli/mongocli/serverless/create.go
wtrocki: @gssbzn Very good question. I believe that there is a benefit to us migrating also mongodb and Atlas commands using the same stores. If after split we going to have 2 separate SDKs - we can easily duplicate the current mongodb-client to use dedicated API that will live their own life. V1 will be used only in mongodbcli commands that have different atlas counterparts.

test/e2e/atlas/serverless_test.go
wtrocki: We would need more details? Do you meant that object is not generic enough?

gssbzn: for example this is running this branch 
```bash
./bin/atlas serverless ls -P dev
Error: template: output:1:29: executing "output" at <.Results>: can't evaluate field Results in type []mongodbatlasv2.ServerlessInstanceDescription
```

andmatei: This is the template used, and I missed changing the `.Results`:
```
var listTemplate = `ID	NAME	MDB VER	STATE{{range .Results}}
{{.Id}}	{{.Name}}	{{.MongoDBVersion}}	{{.StateName}}{{end}}
```

So it looks like we can safely unwrap the envelope

internal/store/serverless_instances.go
andmatei: I've seen this around, and in some of the commands we've migrated we remove the envelope from the v2 client. Should we consider returning the enveloped version of all paginated views now that these are present in the new sdk?

wtrocki (dismissed):
Reviewed

go.sum
```diff
@@ -606,6 +606,7 @@ golang.org/x/sys v0.0.0-20220422013727-9388b58f7150/go.mod h1:oPkhp1MJrh7nUepCBc
 golang.org/x/sys v0.0.0-20220520151302-bc2c85ada10a/go.mod h1:oPkhp1MJrh7nUepCBck5+mAzfO9JrbApNNgaTdGDITg=
 golang.org/x/sys v0.0.0-20220722155257-8c9f86f7a55f/go.mod h1:oPkhp1MJrh7nUepCBck5+mAzfO9JrbApNNgaTdGDITg=
 golang.org/x/sys v0.0.0-20220908164124-27713097b956/go.mod h1:oPkhp1MJrh7nUepCBck5+mAzfO9JrbApNNgaTdGDITg=
+golang.org/x/sys v0.6.0/go.mod h1:oPkhp1MJrh7nUepCBck5+mAzfO9JrbApNNgaTdGDITg=
```
gssbzn, line 609: can you run `go mod tidy` and commit that if any changes, wondering why this showing up here

andmatei: I had an error that `go.sum` was out of sync with `go.mod` so I ran it then. If I run it again now it doesn't change anything

internal/store/serverless_instances.go
gssbzn: >Would make sense for us to apply default: true on openapi leve

Should already be
Screenshot 2023-04-20 at 17 00 14

But then we had this PR https://github.com/mongodb/go-client-mongodb-atlas/pull/443 becuase it seems like the generated client was not always respecting defaults

> Which is weird as in the link it looks like includeCount is set to false...

@wtrocki may be related to the above PR?

internal/store/serverless_instances.go
gssbzn: >  and in some of the commands we've migrated we remove the envelope

I did a quick search of the store and no store is returning .Results like this is (a tleat on master)

>. Should we consider returning the enveloped version of all paginated views now that these are present in the new sdk?

Yes when possible but at least we should maintain behavior, which this change was breaking

go.sum
gssbzn: yep, filipe caught this and there's now  a test t enforce it

internal/store/serverless_instances.go
wtrocki: @andmatei `./bin/atlas serverless list -P dev -o=json --debug` would show you what was the url present.

When we do not include value (it is not set in the struct) we will set it to default value. 
That is why we would not need include count present in any of the stores. 
https://github.com/mongodb/go-client-mongodb-atlas/blob/68e7e21de87ddc93a4c6aae9943f40a0891cbe1a/mongodbatlasv2/api_serverless_instances.go#L557-L558

wtrocki: It is set to default value but never present in URL. Fixing in latest dev. Good we caught that.

wtrocki: Pushed commit. It should be working now

internal/cli/atlas/serverless/create.go
```diff
@@ -61,12 +63,12 @@ func (opts *CreateOpts) Run() error {
...
-func (opts *CreateOpts) newServerlessCreateRequestParams() *atlas.ServerlessCreateRequestParams {
-	return &atlas.ServerlessCreateRequestParams{
+func (opts *CreateOpts) newServerlessCreateRequestParams() *atlasv2.ServerlessInstanceDescriptionCreate {
+	return &atlasv2.ServerlessInstanceDescriptionCreate{
 		Name: opts.instanceName,
-		ProviderSettings: &atlas.ServerlessProviderSettings{
+		ProviderSettings: atlasv2.ServerlessProviderSettings{
 			BackingProviderName: opts.provider,
-			ProviderName:        providerName,
+			ProviderName:        pointer.Get(providerName),
```
gssbzn, line 71: let's save the pointer util for literals
```suggestion
			ProviderName:        &providerName,
```

test/e2e/atlas/serverless_test.go
```diff
@@ -105,12 +105,12 @@ func TestServerless(t *testing.T) {
...
 		req.NoError(err, string(resp))
 
-		var clusters mongodbatlas.ClustersResponse
+		var clusters atlasv2.PaginatedServerlessInstanceDescription
 		err = json.Unmarshal(resp, &clusters)
 		req.NoError(err)
 
 		a := assert.New(t)
-		a.NotEmpty(clusters.Results)
+		a.NotEmpty(clusters)
```
gssbzn, line 113: leftover from a previous change
```suggestion
		a.NotEmpty(clusters.Results)
```

gssbzn (approved):
LGTM after some small fixes but the main blcoker is gone so removing the required changes from the PR so it doesn't need another pass from me, thanks for all the work here 💯

gssbzn (dismissed):
LGTM

matt-condon (changes requested):
Just adding "Request Changes" to temporarily hold out on merging this one.

@drinkbird mentioned that there's an ongoing discussion about doing a user outreach first, so just want clarification on that

matt-condon:
Was discussed in the triage call, looks like we don't want to remove atlas commands from mongocli yet, so may have to revert some of these changes 😢 

CC @gssbzn @wtrocki @colm-quinn

andmatei:
Should we migrate those commands to use the v2 client, or should duplicate the stores in this case too?

matt-condon:
I think just use the v2 client across the board. 
No sense duplicating the stores then removing one set a week or two later

matt-condon (approved):
Thank you! LGTM

PR #1851, opened 2023-04-11
Splitting alert store to atlas and ops manager.

wtrocki:
This PR targets to be as simple as possible to nail approach we want for splitting out stores.
I did not linked stores to simplify review and get early feedback.

IMHO if we nail the approach we might just migrate all stores in the same PR without migrating to v2. 
This way we can:

- Ensure that we have constend approach and naming across PRs
- Fastrack work on every shared commands that would no longer be shared.


matt-condon:
Great idea!
This would also allow us to maintain the behaviour of MongoCLI when running commands in atlas context

My suggestion; keep the existing stores exactly as they are, and duplicate the ```config.CloudService, config.CloudGovService``` case (as you've done)

This way mongoCLI in Atlas context can continue using the sdk v1 for now while we migrate to sdkv2 for atlasCLI 

Only other suggestion is to go with subdirectories/packages

So instead of:

-  `internal/store/alerts_atlas.go` and `internal/store/alerts.go`
-  `internal/store/atlas/alerts.go` and `internal/store/mongocli/alerts.go`

This would be a bit more consistent with the approach taken in recreating commands and would be a bit cleaner for any future CLI split efforts

wtrocki:
I will make changes only for the alert store for this PR to simplify review and I will follow the same approach for the rest of the stores.

fmenezes:
@wtrocki have you considered 2 packages instead of duplicating the names? is it much more work?

wtrocki:
@fmenezes I just tried duplication of the packages and that seems like a lot of extra rewiring for all commands for atlas and mongodb cli. I have done couple of approaches for the packages and stores and I think that the best way forward is:

- Leave current package completely unchanged (for mongodbcli)
- Create completely new atlas package (with completely new store interface and subsequently move there all stores that are for atlas (currently only shared ones to not conflict with others migrating commands at the same time)

wtrocki:
Actually that seem to cause issues with mocks, require a lot of duplication of tests and we cannot really make store in that atlas specific without migrating all in one go. I would settle with separate files for duplicated commands for the moment

wtrocki:
PR with continued attempt to split the stores. It looks like it will require couple days to rewire everything:
https://github.com/mongodb/mongodb-atlas-cli/pull/1854

Challenges are mostly around duplication/conflicts with existing migration that is happening when we work on this.

fmenezes:
I like this way better than `api`

wtrocki:
Thanks to @fmenezes feedabck on other PR I have decided to bring that approach to main PR:

- Store implementation is duplicated. It probably needs some cleanup but it would be premature to do so.
- Individual stores are also duplicated. This is to split usage of v1 in mongodbcli

We not going to duplicate any other atlas specific commands yet as this would require a lot of extra work (migration in two places). Atlas folder is only for atlas specific commands where commands haver their mongodb specific version.

internal/store/atlas/store.go
```diff
@@ -0,0 +1,498 @@
...
+	}
+
+	if store.ctx == nil {
+		store.ctx = context.Background()
+	}
+
+	return store, nil
+}
+
+type ManifestGetter interface {
```
fmenezes, line 476: I know you're avoiding improvements but I think this can go

internal/store/atlas/store.go
```diff
@@ -0,0 +1,498 @@
...
+	responseHeaderTimeout     = 1 * time.Minute
+	telemetryTimeout          = 1 * time.Second
+	tlsHandshakeTimeout       = 5 * time.Second
+	timeout                   = 5 * time.Second
+	keepAlive                 = 30 * time.Second
+	maxIdleConns              = 5
+	maxIdleConnsPerHost       = 4
+	idleConnTimeout           = 30 * time.Second
+	expectContinueTimeout     = 1 * time.Second
+	versionManifestStaticPath = "https://opsmanager.mongodb.com/"
```
gssbzn, line 51: not needed for atlas

internal/store/atlas/store.go
```diff
@@ -0,0 +1,498 @@
...
+		Timeout:   timeout,
+		KeepAlive: keepAlive,
+	}).DialContext,
+	MaxIdleConns:          maxIdleConns,
+	MaxIdleConnsPerHost:   maxIdleConnsPerHost,
+	Proxy:                 http.ProxyFromEnvironment,
+	IdleConnTimeout:       idleConnTimeout,
+	ExpectContinueTimeout: expectContinueTimeout,
+	TLSClientConfig:       &tls.Config{InsecureSkipVerify: true}, //nolint:gosec // this is optional for some users,
+}
```
gssbzn, line 97: not needed for atlas

internal/store/atlas/store.go
```diff
@@ -0,0 +1,498 @@
...
+			KeepAlive: keepAlive,
+		}).DialContext,
+		MaxIdleConns:          maxIdleConns,
+		MaxIdleConnsPerHost:   maxIdleConnsPerHost,
+		Proxy:                 http.ProxyFromEnvironment,
+		IdleConnTimeout:       idleConnTimeout,
+		ExpectContinueTimeout: expectContinueTimeout,
+		TLSClientConfig:       tlsClientConfig,
+	}
+}
```
gssbzn, line 132: not needed for atlas

internal/store/atlas/store.go
```diff
@@ -0,0 +1,498 @@
...
+type Store struct {
+	service       string
+	baseURL       string
+	caCertificate string
+	skipVerify    bool
+	telemetry     bool
+	username      string
+	password      string
+	accessToken   *atlasauth.Token
+	client        interface{}
```
fmenezes, line 66: if possible let's have only SDKv2 and no old client

internal/store/atlas/store.go
```diff
@@ -0,0 +1,498 @@
...
+	c, err := atlasv2.NewClient(opts...)
+	if err != nil {
+		return err
+	}
+	s.clientv2 = c
+	return nil
+}
+
+// setOpsManagerClient sets the internal client to use an Ops Manager client and methods.
+func (s *Store) setOpsManagerClient(client *http.Client) error {
```
gssbzn, line 330: any reason to keep this?

internal/store/atlas/store.go
```diff
@@ -0,0 +1,498 @@
...
+	s.baseURL = versionManifestStaticPath
+	if baseURL := c.OpsManagerVersionManifestURL(); baseURL != "" {
+		s.baseURL = baseURL
+	}
+	if err := s.setOpsManagerClient(http.DefaultClient); err != nil {
+		return nil, err
+	}
+
+	return s, nil
+}
```
gssbzn, line 498: not needed for atlas

internal/store/atlas/store.go
wtrocki: Yes. I have attempted to clean up the ops manager but found out that there is also common logic used in atlas cli.
Logic uses opsManagerURL for atlas in the config that allows us to set custom url.

internal/store/atlas/store.go
wtrocki: It would be done in follow-up PR. I intentionally did not wanted to migrate to v2 client on this as it would make review much harder.

wtrocki:
Awesome feedback! I have done some major improvements on the store (mostly removing stuff).
Let me know if that makes sense.

wtrocki:
Added evergreen gov tests to see if I did not broke anything gov related 🙈

internal/store/atlas/store.go
wtrocki: Done. Left 2 clients for now as we going to migrate things after merging to master. 
I did left jira to remove v1 client once migration finishes so this will be addressed eventually.

wtrocki:
Gov clusters, mongodbcli and atlas cli working as they should

internal/cli/atlas/alerts/acknowledge.go
```diff
@@ -23,7 +23,7 @@ import (
 	"github.com/mongodb/mongodb-atlas-cli/internal/cli/require"
 	"github.com/mongodb/mongodb-atlas-cli/internal/config"
 	"github.com/mongodb/mongodb-atlas-cli/internal/flag"
-	"github.com/mongodb/mongodb-atlas-cli/internal/store"
+	atlasStore "github.com/mongodb/mongodb-atlas-cli/internal/store/atlas"
```
gssbzn, line 26: any reason to include atlas in the name? keeping it `store` reduces the number of changes and I don't belive there be a case were you will include both stores right?
```suggestion
	store "github.com/mongodb/mongodb-atlas-cli/internal/store/atlas"
```

internal/cli/atlas/alerts/acknowledge_test.go
```diff
@@ -22,7 +22,7 @@ import (
 
 	"github.com/golang/mock/gomock"
 	"github.com/mongodb/mongodb-atlas-cli/internal/flag"
-	"github.com/mongodb/mongodb-atlas-cli/internal/mocks"
+	"github.com/mongodb/mongodb-atlas-cli/internal/mocks/atlas"
```
gssbzn, line 25:
```suggestion
	mocks "github.com/mongodb/mongodb-atlas-cli/internal/mocks/atlas"
```

gssbzn (dismissed):
LGTM with the recommendation on keeping imports as `mocks` for mocks and `store` for atlasStore as it will reduces the amount of changes and makes it an easy find and replace

wtrocki:
Superb. I will retrofit your recommendation to PR with the rest of the changes as well

internal/cli/atlas/alerts/acknowledge_test.go
wtrocki: Done in all stores/cmd

internal/cli/atlas/alerts/acknowledge.go
wtrocki: Yes. No 2 stores should be present.
I wanted to make sure we aren't confusing stores (duplicated files). 
I wanted to make sure it is easy to search replace. 
store as identifier is used everywhere like `ops.store` etc. 
store as name is failry generic so it would be actually much harder to replace it.
Will do change but it feels like more explicit is better.

matt-condon (approved):
LGTM

gssbzn (approved):
LGTM

PR #1854, opened 2023-04-11
Sdk migration/duplicate store

wtrocki:
WIP of the WIP

Create a new duplicate of the store only for atlas commands. 
This PR attempts to explore how much it will require to have 2 separate stores:

1) atlas only v2 (called `api`)
2) mixed v1 and ops manager (store)

## Progress

- [x] Duplicate package
- [x] Remove Ops Manager from first half of the stores 
- [ ] Remove Ops Manager second half of the stores 
- [ ] Rewire atlas only commands
- [ ] Tests and mock generation


wtrocki:
@fmenezes Let me know what you think about this approach in general. Does that make sense. I can try to make this work tommorow.

fmenezes:
yes, this was my original point

fmenezes:
I know naming things is hard but I've found the term `api` a bit confusing 😅

fmenezes:
My suggestion is: we can simply move it to `store/atlas` it would be identical to what you've done and it would keep the name `store`

fmenezes:
I'm not a big fan of the term `store` but I guess it has a meaning now-a-days

wtrocki:
Closing as I have applied step by step approach instead in the other PR

PR #1867, opened 2023-04-14
task: rename loadSampleData to match go naming convention

gssbzn:
Spotted the other day while trying to check if we had a watch for the load sample data


fmenezes (approved):

fmenezes:
We desperately need a watch for loadSampleData

PR #1886, opened 2023-04-19
Update organization commands to point to atlas stores

wtrocki:
Organizations and related endpoints have been missing connection to the commands.
Change focusing only on rewiring commands to use new stores. There are not v2 changes.


internal/cli/atlas/organizations/create.go
```diff
@@ -1,86 +0,0 @@
-// Copyright 2023 MongoDB Inc
```
wtrocki, line 1: This haven't been used (IAM package is used for mongodb cli)

internal/cli/atlas/organizations/create_atlas.go
```diff
@@ -38,7 +38,7 @@ type CreateAtlasOpts struct {
 	ownerID           string
 	apiKeyDescription string
 	apiKeyRole        []string
-	store             store.AtlasOrganizationCreator
```
wtrocki, line 41: Since store is atlas there is no sense for us to use Atlas as name

matt-condon (approved):
Thank you for this!

PR #1888, opened 2023-04-19
CLOUDP-171830: error handling example

wtrocki:
## Proposed changes

Example for error handling using SDK helper method.
I will apply changes across all stores after the initial review.

### DIFF
--- a/internal/store/custom_dns.go
+++ b/internal/store/custom_dns.go
@@ -21,6 +21,7 @@ import (
 
 	"github.com/mongodb/mongodb-atlas-cli/internal/config"
 	atlasv2 "go.mongodb.org/atlas/mongodbatlasv2"
+	sdkv2 "go.mongodb.org/atlas/sdk"
 )
 
 type CustomDNSEnabler interface {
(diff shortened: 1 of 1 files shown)


internal/store/custom_dns.go
```diff
@@ -56,7 +57,12 @@ func (s *Store) DisableCustomDNS(projectID string) (*atlasv2.AWSCustomDNSEnabled
 		customDNSSetting := &atlasv2.AWSCustomDNSEnabled{
 			Enabled: false,
 		}
-		result, _, err := s.clientv2.AWSClustersDNSApi.ToggleAWSCustomDNS(s.ctx, projectID).AWSCustomDNSEnabled(*customDNSSetting).Execute()
+		result, _, err := s.clientv2.AWSClustersDNSApi.ToggleAWSCustomDNS(s.ctx, projectID).
+			AWSCustomDNSEnabled(*customDNSSetting).Execute()
+		apiError := sdkv2.APIError(err)
+		if apiError != nil {
+			return nil, fmt.Errorf("%v", apiError)
```
wtrocki, line 64: To minimize boilerplate we proposing returning error object directly in SDK in: https://github.com/mongodb/go-client-mongodb-atlas/pull/459

andreaangiolillo, line 64:
```suggestion
		if apiError := sdkv2.APIError(err); apiError != nil {
			return nil, fmt.Errorf("%v", apiError)
```

andreaangiolillo: is there any reason why you are using `fmt.Errorf` without adding any additional context to the error instead of returning directly the error?

andreaangiolillo: Could you share the logic of `sdkv2.APIError` in the go SDK? Thanks!

PR #1900, opened 2023-04-24
fix: flag descriptions / response messages

raffis:
## Proposed changes

This is just a simple pr fixing some cmd descriptions I noticed when I wanted to export a backup to s3.

* The create job should give the ID. Far more useful than printing the bucket ID
* The watch command operates on jobId, not bucketId
* Problem with sharing flags cross packages. The description for snapshotId is not suitable in the export command. Nothing is restored, it's exported. Rewritten to be more generic.

## Checklist

- [x]  I have signed the [MongoDB CLA](https://www.mongodb.com/legal/contributor-agreement)

### DIFF
--- a/docs/atlascli/command/atlas-backups-exports-jobs-create.txt
+++ b/docs/atlascli/command/atlas-backups-exports-jobs-create.txt
@@ -89,7 +89,7 @@ If the command succeeds, the CLI returns output similar to the following sample.
 
 .. code-block::
 
-   Export job '<ID>' created in a bucket with ID '<ExportBucketId>'.
+   Export job '<Id>' created in a bucket with ID '<ExportBucketId>'.
    
 
 Examples
--- a/docs/mongocli/command/mongocli-atlas-backups-restores-start.txt
+++ b/docs/mongocli/command/mongocli-atlas-backups-restores-start.txt
@@ -87,7 +87,7 @@ Options
    * - --snapshotId
      - string
      - false
-     - Unique identifier of the snapshot.
+     - Unique identifier of the snapshot to restore. You must specify a snapshotId for automated restores.
    * - --targetClusterName
      - string
      - false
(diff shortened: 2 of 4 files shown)


gssbzn:
Hi @raffis thanks for your contribution, can you please run `make gen-docs` to also update the docs?

internal/usage/usage.go
```diff
@@ -206,7 +206,7 @@ dbName and collection are required only for built-in roles.`
 	NotificationType                          = "Type of alert notification. Valid values are DATADOG, EMAIL, FLOWDOCK, GROUP (Project), ORG, OPS_GENIE, PAGER_DUTY, SLACK, SMS, USER, or VICTOR_OPS."
 	NotificationUsername                      = "Name of the Atlas user to which to send notifications."
 	NotificationVictorOpsRoutingKey           = "Routing key associated with your Splunk On-Call account."
-	SnapshotID                                = "Unique identifier of the snapshot to restore. You must specify a snapshotId for automated restores."
+	SnapshotID                                = "Unique identifier of the snapshot."
```
gssbzn, line 209: I think this may have been an oversight on reusing the constant for multiple commands, 

>to restore. You must specify a snapshotId for automated restores

Still applies to restore commands but on a quick look it doesn't for commands like `snapshots describe`, do you know which commands you wanted to fix this for as I'd like to tkeep the original description for restore but create a new const for things like that snapshot describe

raffis: Yes it's in the description. I noticed the wrong description while exporting a snapshot to an s3 bucket.

gssbzn: Is it ok if I push some changes to your branch? if not let's roll this change and I can address as a separate PR

gssbzn: Sorry it took a while to get back to this, I pushed some changes on your branch to keep the original usage where applicable and the new simplified one for newer commands where it doesn't have the same meaning

gssbzn (approved):

PR #1913, opened 2023-04-27
CLOUDP-173973: Improve error handling for cluster describe if no output and name is provided

gssbzn:
## Proposed changes

_Jira ticket:_ CLOUDP-173973

Improve validation of input for clusters given certain conditions will trigger different endpoints with different types

### DIFF
--- a/internal/cli/opsmanager/clusters/describe.go
+++ b/internal/cli/opsmanager/clusters/describe.go
@@ -57,10 +57,11 @@ func (opts *DescribeOpts) Run() error {
 	return opts.Print(r)
 }
 func (opts *DescribeOpts) validateArg() error {
-	if opts.ConfigOutput() == "" {
-		if err := validate.ObjectID(opts.name); err != nil {
-			return fmt.Errorf("please provide a valid cluster ID, %w", err)
-		}
+	if opts.ConfigOutput() != "" {
+		return nil
+	}
+	if err := validate.ObjectID(opts.name); err != nil {
+		return fmt.Errorf("please provide a valid cluster ID or provide an output format to use names, %w", err)
 	}
 	return nil
 }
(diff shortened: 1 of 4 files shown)


internal/cli/opsmanager/clusters/describe.go
```diff
@@ -55,6 +56,14 @@ func (opts *DescribeOpts) Run() error {
 
 	return opts.Print(r)
 }
+func (opts *DescribeOpts) validateArg() error {
```
blva, line 59: this got me a bit confused, when we list clusters, we get ID and name

```shell
~ ❯ atlas clusters list                                             7s 13:50:52
ID                         NAME           MDB VER   STATE
644a6faa12c63d0ea45ef37a   Cluster99845   6.0.5     CREATING
```

Are we supposed to validate the name as ID? maybe it should be another flag?

blva: sorry wait I used altas here let me check OM

gssbzn: >maybe it should be another flag?
 
```
mongocli ops-manager clusters describe --help
When describing cluster with no output format please provide the cluster ID.
When using an output format the please provide the cluster name.
```


At this point any change would be a breaking change there's a ticket to figure out what to do with these commands but it hasn't been prioritized so far

blva: ah, got it! thanks for the clarification

internal/cli/opsmanager/clusters/describe.go
```diff
@@ -55,6 +56,14 @@ func (opts *DescribeOpts) Run() error {
...
+	if opts.ConfigOutput() == "" {
+		if err := validate.ObjectID(opts.name); err != nil {
+			return fmt.Errorf("please provide a valid cluster ID, %w", err)
```
blva, line 62:
```suggestion
			return fmt.Errorf("please provide a valid cluster ID or provide a cluster name and output format, %w", err)
```

blva (approved):
LGTM

blva (approved):

PR #1928, opened 2023-05-04
CLOUDP-175225: Remove incorrect alias for atlas cli

gssbzn:
## Proposed changes

_Jira ticket:_ CLOUDP-175225

Top level commands can't have aliases

### DIFF
--- a/internal/cli/root/atlas/builder.go
+++ b/internal/cli/root/atlas/builder.go
@@ -113,7 +113,6 @@ func Builder() *cobra.Command {
 	rootCmd := &cobra.Command{
 		Version: version.Version,
 		Use:     atlas,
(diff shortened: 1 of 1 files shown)


internal/cli/root/atlas/builder.go
```diff
@@ -113,7 +113,6 @@ func Builder() *cobra.Command {
 	rootCmd := &cobra.Command{
 		Version: version.Version,
 		Use:     atlas,
-		Aliases: []string{config.ToolName},
```
blva, line 116: [q] could this be a breaking change or does the alias not work in this case?

gssbzn: It does nothing, it's a fake alias

blva (approved):

PR #1962, opened 2023-05-24
Auto generation POC

fmenezes:
⚠️ DO NOT MERGE ⚠️ 

Jiras: CLOUDP-179039, CLOUDP-181608

This is a POC of auto-generating commands using openapi-generator-cli

# Demo
https://github.com/mongodb/mongodb-atlas-cli/assets/1765216/bae365c3-8c8d-40a0-9e8e-89b077ea4882

# Design
* Some design choices were made considering scripted usage only and non-interactive commands.
* All commands that pass a request body to the API have to pass a json via `--file` or standard input
* Unlike our CLI today there is no `--output` instead all commands write json and have a `--format` that you can use to filter data (similar to `docker inspect`)
* Watchers are possible to implement but require manual intervention ([ref](https://github.com/mongodb/mongodb-atlas-cli/pull/1962/files#diff-4ea51727c8d167dba13ec0ce64bd5e36d21d5927e1410bd9c2ddd530dfdf7066R5-R14))
* Docs are taken straight from the API, but can be customized ([example](https://github.com/mongodb/mongodb-atlas-cli/pull/1962/files#diff-4ea51727c8d167dba13ec0ce64bd5e36d21d5927e1410bd9c2ddd530dfdf7066R4))
* commands are grouped following how our APIs are grouped, so there is no deep nesting for commands anymore
* all generated commands are under `atlas api`

# Considerations
* I've used https://openapi-generator.tech/ to generate commands straight from open API spec, currently I'm downloading the transformed openapi spec that SDK is using, which is a requirement given we need our data types matching across given we use SDK in the generated code.
* I'm using `go.mod` to fetch which version of SDK is currently used so regenerating commands would be dependent on SDK upgrades, as in a PR is opened to update SDK and we run `make gen-openapi` on the same PR


tools/openapi-generator/config/go-templates/cmd.mustache
```diff
@@ -0,0 +1,104 @@
...
+		},
+	}
+{{#pathParams}}
+	cmd.Flags().StringVar(&opts.{{paramName}}, "{{paramName}}", {{defaultValue}}, "{{description}}")
+{{/pathParams}}
+{{#queryParams}}
+	cmd.Flags().StringVar(&opts.{{paramName}}, "{{paramName}}", {{defaultValue}}, "{{description}}")
+{{/queryParams}}
+{{#pathParams}}{{#required}}
+	_ = cmd.MarkFlagRequired("{{paramName}}"){{/required}}{{/pathParams}}
```
wtrocki, line 81:
```suggestion
	_ = cmd.MarkFlagRequired("{{paramName}}"){{/required}}{{/pathParams}}	
	// Take look on how much it has
	{{.}}
	// Most Likely we need 
	cmd.Flags().{{dataType}}Var or camelized version of it. 
```

wtrocki:
Truly astonishing how close this looks like to user written code.

fmenezes:
reopen

fmenezes:
@wtrocki I've decided to embrace our ideas and for json input/ouput works like a charm, fully functional now

wtrocki:
@fmenezes Since the last break we also squashed all oneOf objects into flattened structures. 
That also makes things much easier as it removes the need for a generator to pick types. 

We still have problems with mutually exclusive fields that need to be figured out somehow otherwise end users have the burden of figuring out. https://github.com/mongodb/mongodb-atlas-cli/pull/2208

fmenezes:
> @fmenezes Since the last break we also squashed all oneOf objects into flattened structures. That also makes things much easier as it removes the need for a generator to pick types.
> 
> We still have problems with mutually exclusive fields that need to be figured out somehow otherwise end users have the burden of figuring out what fields to use. Right now we have manual validation. For example here: #2208

I'll talk to Product but I think our goal is for scripting in which API errors would be expected

tools/openapi-generator/config/go-templates/cmd.mustache
```diff
@@ -123,30 +140,79 @@ func (opts *{{#lambda.camelcase}}{{operationId}}{{/lambda.camelcase}}Opts) run(c
...
+					{{key}}: *opts.resp.{{value}},
+					{{/vendorExtensions.x-watcher-params.2}}
+				}
+				resp, _, err := opts.client.{{classname}}.{{#lambda.titlecase}}{{.}}{{/lambda.titlecase}}WithParams(ctx, params).Execute()
+				if err != nil {
+					return err
+				}
+				if resp.{{vendorExtensions.x-watcher-return-property}} != nil && *resp.{{vendorExtensions.x-watcher-return-property}} == "{{vendorExtensions.x-watcher-state}}" {
+					return nil
+				}
```
tibulca, line 182: probably a `time.Sleep(2 * time.Second)` would help here to avoid rate limiting

fmenezes: done! added 4 since that is the default for the current watcher

tibulca:
bump

PR #1963, opened 2023-05-25
CLOUDP-179008: Only create foliage tickets for the task and not test on cleanup

gssbzn:
## Proposed changes

_Jira ticket:_ CLOUDP-179008

enable foliage to only open a ticket per task and not per test


swattyT (approved):
LGTM!

wtrocki (approved):

PR #1980, opened 2023-06-01
CLOUDP-178273 : [Monitoring and Logs API] Migrate DownloadLog Command to v2 SDK

swattyT:
## Proposed changes

_Jira ticket:_  [CLOUDP-178273](https://jira.mongodb.org/browse/CLOUDP-178273)

Closes #[issue number]

## Checklist


internal/store/logs.go
```diff
@@ -82,7 +85,26 @@ func (s *Store) Collect(groupID string, newLog *opsmngr.LogCollectionJob) (*opsm
 func (s *Store) DownloadLog(groupID, host, name string, out io.Writer, opts *atlas.DateRangetOptions) error {
 	switch s.service {
 	case config.CloudService, config.CloudGovService:
-		_, err := s.client.(*atlas.Client).Logs.Get(s.ctx, groupID, host, name, out, opts)
+		fileBaseName := strings.TrimSuffix(name, filepath.Ext(name))
+		result := s.clientv2.MonitoringAndLogsApi.GetHostLogs(s.ctx, groupID, host, fileBaseName)
+		if opts != nil {
+			if opts.StartDate != "" {
```
drinkbird, line 91: If this comes as `nil` instead of an empty string, what will the code do?

internal/store/logs.go
```diff
@@ -82,7 +85,26 @@ func (s *Store) Collect(groupID string, newLog *opsmngr.LogCollectionJob) (*opsm
...
+				start, _ := strconv.ParseInt(opts.StartDate, 10, 64)
+				result = result.StartDate(start)
```
drinkbird, line 93: if the parse operation above fails, what will be assigned here? nil or empty string? 

if opts.EndDate is nil, what will the code do?

internal/store/logs.go
wtrocki: In this particular example we dealing with string datatype that cannot be nil. 
https://github.com/mongodb/go-client-mongodb-atlas/blob/c3435ec31149df5973ef01f992bb72d598615a39/mongodbatlas/logs.go#L41-L42

internal/store/logs.go
```diff
@@ -82,7 +85,26 @@ func (s *Store) Collect(groupID string, newLog *opsmngr.LogCollectionJob) (*opsm
 func (s *Store) DownloadLog(groupID, host, name string, out io.Writer, opts *atlas.DateRangetOptions) error {
```
wtrocki, line 85: We have dedicated v2 objects to organize those arguments. Can we use that?

internal/store/logs.go
wtrocki: The problem is that we should not use v1 objects itself and v2 will actually have pointer

swattyT: `opts.StartDate` is declared as string 
```
type DateRangetOptions struct {
    StartDate string `url:"startDate,omitempty"`
    EndDate   string `url:"endDate,omitempty"`
}
```
ParseInt only accepts string value in the argument.

internal/store/logs.go
```diff
@@ -82,7 +85,26 @@ func (s *Store) Collect(groupID string, newLog *opsmngr.LogCollectionJob) (*opsm
...
+				end, _ := strconv.ParseInt(opts.EndDate, 10, 64)
+				result = result.EndDate(end)
+			}
+		}
+		logs, _, err := result.Execute()
+		if err != nil {
+			return err
+		}
+		if logs == nil {
+			return fmt.Errorf("returned file is empty")
```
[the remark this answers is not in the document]
swattyT: This is to return an `error` object.

internal/store/logs.go
swattyT: Sure, I will use `GetHostLogsWithParams` instead.

wtrocki:
[nit] [optional] I think this PR might be using the old approach for fetching data from the backend. 
With a number of arguments and parseInt required I would consider simplifying using `WithParams` struct.
See: https://github.com/mongodb/atlas-sdk-go/blob/main/docs/doc_1_concepts.md#fetching-data-from-the-back-end

Otherwise looks good

wtrocki (approved):
LGTM. Thank you for using v2 models.

matt-condon (approved):

internal/store/logs.go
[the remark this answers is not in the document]
drinkbird: I was under the impression that there is an `fmt.Error(` (without the `f`) but there isn't one, so all good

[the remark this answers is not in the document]
wtrocki: [errors.New()](https://pkg.go.dev/errors#New) enters the chat

internal/store/logs.go
wtrocki: @swattyT I would recommend logging improvement jira to cover the date time for the atlas CLI and SDKs.
Both systems have different standards but are always the same so we can write familiar helper and do transformation directly at the Cobra flag level. That will let us leave complex `start, _ := strconv.ParseInt(opts.StartDate, 10, 64)` out of the future code reviews.

PR #1471, opened 2022-10-06
CLOUDP-138980:  [AtlasCLI] Add tests to check the default format of the list and describe commands - Part 1

andreaangiolillo:
## Proposed changes

_Jira ticket:_ CLOUDP-138980

## Checklist

- [x] I have signed the [MongoDB CLA](https://www.mongodb.com/legal/contributor-agreement)
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] I have added any necessary documentation in document requirements section listed in [CONTRIBUTING.md](https://github.com/mongodb/mongodb-atlas-cli/blob/master/CONTRIBUTING.md) (if appropriate)
- [x] I have addressed the @mongodb/docs-cloud-team comments (if appropriate)
- [x] I have updated [test/README.md](https://github.com/mongodb/mongodb-atlas-cli/blob/master/test/README.md) (if an e2e test has been added)
- [x] I have run `make fmt` and formatted my code

## Further comments

This is the first PR of many where I am updating the unit test to check the default format of commands to avoid issues such as https://github.com/mongodb/mongodb-atlas-cli/pull/1470

**Next Step:**

- I will create a new ticket to update the remaining unit tests and use this PR as an example. FYI this ticket will be for new joiners.

PR #1985, opened 2023-06-05
CLOUDP-181546: slow query logs

wtrocki:
## Proposed changes

Fixing structure changes that were missing in the go template.

### DIFF
--- a/internal/cli/atlas/performanceadvisor/slowquerylogs/list.go
+++ b/internal/cli/atlas/performanceadvisor/slowquerylogs/list.go
@@ -28,7 +28,7 @@ import (
 	atlasv2 "go.mongodb.org/atlas-sdk/admin"
 )
 
-const listTemplate = `NAMESPACE	LINE{{range .SlowQuery}}
+const listTemplate = `NAMESPACE	LINE{{range .SlowQueries}}
 {{.Namespace}}	{{.Line}}{{end}}
 `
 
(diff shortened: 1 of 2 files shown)


wtrocki:
Slow Query logs seem to take time to appear as items. Looking for the way to improve integration test as it bases only on the empty response right now.

wtrocki:
After investigating problem I see that it will be quite challenging for us to simulate any responses in slowQueryLogs due to the fact that it takes some time to make them and collect them thru API.

internal/cli/atlas/performanceadvisor/slowquerylogs/list.go
```diff
@@ -28,7 +28,7 @@ import (
 	atlasv2 "go.mongodb.org/atlas-sdk/admin"
 )
 
-const listTemplate = `NAMESPACE	LINE{{range .SlowQuery}}
+const listTemplate = `NAMESPACE	LINE{{range .SlowQueries}}
```
andreaangiolillo, line 31: Could you add a test to check that the template is correct? You can use this PR as an example https://github.com/mongodb/mongodb-atlas-cli/pull/1471

wtrocki: Awesome suggestion! We already have unit test for each command but that does use raw json.
I'm investigating if we can do this for all commands with little effort track those issues in the future.

matt-condon (dismissed):

internal/cli/atlas/performanceadvisor/slowquerylogs/list.go
andreaangiolillo: there is a ticket in the backlog to add this check to all the unit tests

swattyT (approved):
LGTM!

swattyT (approved):
LGTM!

gssbzn (approved):

PR #1986, opened 2023-06-05
CLOUDP-181258: document way to obtain performance advisor processId

wtrocki:
## Proposed changes

Adding extended documentation for the atlas cli explaining how to get process values. 

## Motivation

Explain to the users how to get the exact value of the process id.

In the long run we  should consider having dedicated command or introduce suggestions.

### DIFF
--- a/internal/usage/usage.go
+++ b/internal/usage/usage.go
@@ -22,7 +22,7 @@ const (
 	Members                      = "Number of members in the replica set."
 	Shards                       = "Number of shards in the cluster."
 	ProcessName                  = "Unique identifier for the host of a MongoDB process in the following format: {hostname}:{port}."
-	ProcessNameAtlasCLI          = "Unique identifier for the host of a MongoDB process in the following format: {hostname}:{port}. List of possible values can be obtained by running: atlas processes list | jq '.results[].id'"
+	ProcessNameAtlasCLI          = "Unique identifier for the host of a MongoDB process in the following format: {hostname}:{port}. List of possible values can be obtained by retrieving one of the 'id' fields from 'atlas processes list' command."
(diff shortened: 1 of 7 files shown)


internal/usage/usage.go
```diff
@@ -22,6 +22,7 @@ const (
 	Members                      = "Number of members in the replica set."
 	Shards                       = "Number of shards in the cluster."
 	ProcessName                  = "Unique identifier for the host of a MongoDB process in the following format: {hostname}:{port}."
+	ProcessNameAtlasCLI          = "Unique identifier for the host of a MongoDB process in the following format: {hostname}:{port}. List of possible values can be obtained by running: atlas processes list | jq '.results[].id'"
```
gssbzn, line 25: can you please rever the `jq` part, it assumes you are using json as output which is not the default

wtrocki: Would you preffer json-path or no outputs flags?

gssbzn: no output

gssbzn (dismissed):
LGTM if looks good to docs

internal/usage/usage.go
```diff
@@ -22,6 +22,7 @@ const (
 	Members                      = "Number of members in the replica set."
 	Shards                       = "Number of shards in the cluster."
 	ProcessName                  = "Unique identifier for the host of a MongoDB process in the following format: {hostname}:{port}."
+	ProcessNameAtlasCLI          = "Unique identifier for the host of a MongoDB process in the following format: {hostname}:{port}. List of possible values can be obtained by retrieving one of the 'id' fields from 'atlas processes list' command."
```
sarahsimpers, line 25:
```suggestion
	ProcessNameAtlasCLI          = "Unique identifier for the host of a MongoDB process in the following format: {hostname}:{port}. You can obtain a list of possible values from the 'id' field when you run the 'atlas processes list' command."
```

sarahsimpers:
Can you run make gen-docs again? Then should be good on my end

sarahsimpers (approved):
LGTM

PR #2010, opened 2023-06-15
CLOUDP-183519: introduce custom handling for valid ISO-8601 timestamps

matt-condon:
## Proposed changes

For v1 SDK, timestamps were sent from the SDK as strings and only parsed once they reached the API 

Since migrating to autogenerated v2 Go SDK, timestamps are sent as `time.Time` structs, meaning we need to parse 

The end result is that we are using a different parser since moving to the v2 SDK, which needs to support some custom layouts that were supported by the API for the v1 SDK

_Jira ticket:_ CLOUDP-183519

## Checklist

### DIFF
--- a/internal/time/time.go
+++ b/internal/time/time.go
@@ -20,7 +20,6 @@ func ParseTimestamp(timestamp string) (time.Time, error) {
 	layouts := []string{
 		time.RFC3339,
 		"2006-01-02T15:04:05-0700",
-		"2006-01-02T15:04:05Z07:00",
 	}
 
 	var parsedTime time.Time
--- a/internal/time/time.go
+++ b/internal/time/time.go
@@ -1,0 +1,37 @@
+// Copyright 2023 MongoDB Inc
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package time
+
+import "time"
+
+func ParseTimestamp(timestamp string) (time.Time, error) {
+	layouts := []string{
+		time.RFC3339,
+		"2006-01-02T15:04:05-0700",
+		"2006-01-02T15:04:05Z07:00",
+	}
+
+	var parsedTime time.Time
+	var err error
+
+	for _, layout := range layouts {
+		parsedTime, err = time.Parse(layout, timestamp)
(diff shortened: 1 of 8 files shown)


wtrocki:
> Since migrating to autogenerated v2 Go SDK, timestamps are sent as time.Time structs, meaning we need to parse

FYI - we can change that on the SDK level. This will mean that SDK will not offer any type of date validation and leave users with understanding what date string backend supports.

internal/time/time.go
```diff
@@ -0,0 +1,37 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package time
+
+import "time"
+
+func ParseTimestamp(timestamp string) (time.Time, error) {
+	layouts := []string{
```
wtrocki, line 20: I think we look for 2 new formats where the timezone is hardcoded 
```
YYYY-MM-DDTHH:MM:SS-0700
YYYY-MM-DDTHH:MM:SSZ07:00
```

In this format timezone is hardcoded. Meaning:
 "0700" indicating the time's offset from UTC in hours and minutes.

We need to generalizethat.. but that leads us to the standard time format (with and without zulu)
image

internal/time/time.go
```diff
@@ -0,0 +1,37 @@
...
+		time.RFC3339,
+		"2006-01-02T15:04:05-0700",
```
wtrocki, line 22: This looks like `ISO 8601`

wtrocki (approved):

internal/time/time.go
wtrocki: This doesn't have representation so we can keep it in such form but I will name it as constant to make it clear.

wtrocki:
Disscussed offline. We do not need to change SDK as we do parse time to string in unified way:
https://github.com/mongodb/atlas-sdk-go/blob/a45c8e06d173e50c7467eb4497adbf3b2b01cf6a/tools/config/go-templates/client.mustache#L172

wtrocki (dismissed):

swattyT (dismissed):
LGTM! Thanks for fixing this!

swattyT (commented):
I see `time.Parse()` used at these places as well:
https://github.com/mongodb/mongodb-atlas-cli/blob/6e4fed49212a9480d1f6262b4968910164804af5/internal/convert/database_user.go#L63
https://github.com/mongodb/mongodb-atlas-cli/blob/6e4fed49212a9480d1f6262b4968910164804af5/internal/cli/atlas/datalakepipelines/availablesnapshots/list.go#L59
https://github.com/mongodb/mongodb-atlas-cli/blob/6e4fed49212a9480d1f6262b4968910164804af5/internal/cli/atlas/datalakepipelines/availablesnapshots/list.go#L66

wtrocki (approved):

swattyT (approved):
LGTM! Thanks for the changes 🥇

wtrocki:
CC @fmenezes for CLI generation efforts.

PR #2033, opened 2023-06-22
Cloudp 185931

wtrocki:
h4. WHY

Run e2e in debug mode to be able to reason the issues that are happening in e2e tests. 
We are consuming a large amount of time trying to understand e2e flakiness and also some of the changes that happen in the backend

h4. How

We need to be able to enable debug flag on demand and disable this for certain methods. Meaning that debug cannot be applied globally due to some of the commands might print out dev env links that we do not want to be visible.

### DIFF
--- a/test/e2e/cli.go
+++ b/test/e2e/cli.go
@@ -22,6 +22,9 @@ import (
 	"path/filepath"
 )
 
+// Used for commands that are safe to return debug flag
+const DebugFlag = "--debug "
+
 func Bin() (string, error) {
 	path := os.Getenv("MCLI_E2E_BINARY")
 	cliPath, err := filepath.Abs(path)
--- a/test/e2e/atlas/access_lists_test.go
+++ b/test/e2e/atlas/access_lists_test.go
@@ -45,7 +45,7 @@ func TestAccessList(t *testing.T) {
 	req.NoError(err)
 
 	t.Run("Create Forever", func(t *testing.T) {
-		cmd := exec.Command(cliPath,
+		cmd := exec.Command(cliPath, e2e.DebugFlag,
 			accessListEntity,
 			"create",
 			entry,
@@ -73,7 +73,7 @@ func TestAccessList(t *testing.T) {
 	})
 
 	t.Run("List", func(t *testing.T) {
-		cmd := exec.Command(cliPath,
+		cmd := exec.Command(cliPath, e2e.DebugFlag,
 			accessListEntity,
 			"ls",
 			"--projectId",
@@ -88,7 +88,7 @@ func TestAccessList(t *testing.T) {
 	})
 
 	t.Run("Describe", func(t *testing.T) {
-		cmd := exec.Command(cliPath,
+		cmd := exec.Command(cliPath, e2e.DebugFlag,
 			accessListEntity,
 			"describe",
 			entry,
@@ -104,7 +104,7 @@ func TestAccessList(t *testing.T) {
 	})
 
 	t.Run("Delete", func(t *testing.T) {
-		cmd := exec.Command(cliPath,
+		cmd := exec.Command(cliPath, e2e.DebugFlag,
 			accessListEntity,
 			"delete",
 			entry,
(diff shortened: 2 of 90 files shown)


wtrocki:
We will: 

Inspect output
Remove debug from tests that might not be desired to run in debug.
Rerun entire test suite.

test/e2e/cli.go
```diff
@@ -22,6 +22,9 @@ import (
 	"path/filepath"
 )
 
+// Used for commands that are safe to return debug flag
+const DebugFlag = "--debug "
```
andreaangiolillo, line 26: This seems to be always set. How would you disable this without changing the code?
Could you populate the value of `DebugFlag` based on an env variable?

wtrocki: We have tried that initially and it has been proven that we cannot blindly run debug for every command. 
This PR evaluates that idea by adding debug but only for certain commands that are safe to do so.

If I understand your proposal it means that we want to go to hybrid mode:

1. Add debug to selected commands
2. Turn it on and of in the evergreen config in the repository.

Wondering - in what cases do we want to disable debug? It feels like it should be there for e2e tests as reporting of errors is widely missing

andreaangiolillo: > Wondering - in what cases do we want to disable debug? It feels like it should be there for e2e tests as reporting of errors is widely missing

The error should already be shown in the log 🤔 . My understanding is that the debug flag shows you only the API requests.
Here is an example with and without the debug flag (the error is shown in both commands. )
```bash
atlas cluster ls                                                    1.20.3  16.18.0 bazel 5.4.0
Error: the provided value '63f4d4a47baeac59406dc13' is not a valid ID
```
```bash
atlas cluster ls --debug                                            1.20.3  16.18.0 bazel 5.4.0
Error: the provided value '63f4d4a47baeac59406dc13' is not a valid ID
telemetry: events: [{2023-06-22 12:35:35.094212 +0200 CEST m=+0.604936126 atlascli map[alias:ls arch:arm64 auth_method:api_key command:atlas-clusters-list duration:530 error:the provided value '63f4d4a47baeac59406dc13' is not a valid ID flags:[debug] git_commit:homebrew-release ops_manager_url:https://cloud-dev.mongodb.com/ org_id:60ddf55c27a5a20955a707d7 os:darwin profile:default project_id:63f4d4a47baeac59406dc13 result:ERROR service:cloud terminal:tty version:HEAD-4640eb6]}]
request:
POST https://cloud-dev.mongodb.com/api/private/v1.0/telemetry/events
response:
HTTP/1.1 202 Accepted
Referrer-Policy: strict-origin-when-cross-origin
X-Mongodb-Service-Version: gitHash=1aa86f37f8e4d740593231a26ce999d14637f431; versionString=master
Vary: Accept-Encoding
X-Content-Type-Options: nosniff
X-Java-Method: TelemetryResource::events
X-Frame-Options: DENY
Content-Length: 0
Server: mdbws
Date: Thu, 22 Jun 2023 10:35:35 GMT
Strict-Transport-Security: max-age=31536000; includeSubdomains;
X-Permitted-Cross-Domain-Policies: none
```

wtrocki: > I would say that when the log is too verbose is not helpful to find quickly the issue.

I value your opinion. I feel that we would like to avoid debug for general use and run it only for certain cases? Would that work?

This PR has been put as a proposal to address some of the issues we have seen with visibility. Example:

> '63f4d4a47baeac59406dc13' is not a valid ID

But in many cases Id is valid or we had some timing issues where without request info we need to spend +30-50 minutes investigating. With debug that would be just 2-3 minutes.
This information is always printed on top with follow-up debug info. 


I will discuss this further with my team to see if we can avoid visibility issues.

andreaangiolillo: sure, my comment is not blocking just something to consider. Thanks for the discussion

wtrocki: @andreaangiolillo very valuable feedback. Also the reason why I did create that PR. 
For such small changes (less than 1h of work) proposals do not work well. 
I'm trying not to rush with getting this merged so any perspective here is very important.

PR #2037, opened 2023-06-23
CLOUDP-181599 Add CRUD commands for “atlas dataFederation queryLimits list/create/delete/describe”

fmenezes:
## Proposed changes

_Jira ticket:_ CLOUDP-#

Closes #[issue number]

## Checklist


internal/usage/usage.go
```diff
@@ -437,4 +437,7 @@ dbName and collection are required only for built-in roles.`
 	CompactResponse                           = "Flag that enables the compact array response structure for a json output. The --compact option returns array objects as top-level responses and allows backward compatibility for scripts based on previous CLI versions. Omitting the --compact option for a json output returns array objects within a 'results' sub-array. You must specify --output json to use this option."
 	DataFederationFile                        = "Path to an optional JSON configuration file that defines data federation settings."
 	DataFederationType                        = "Type of Federated Database Instances to return"
+	DataFederation                            = "Identifier of the Federated Database Instance"
```
tibulca, line 440:
```suggestion
	DataFederation                            = "Identifier of the Federated Database Instance."
```

internal/usage/usage.go
```diff
@@ -437,4 +437,7 @@ dbName and collection are required only for built-in roles.`
...
+	Value                                     = "Value given to the query limit"
+	OverrunPolicy                             = "Action to take when the usage limit is exceeded."
```
tibulca, line 442:
```suggestion
	DataFederationType                        = "Type of Federated Database Instances to return."
	DataFederation                            = "Identifier of the Federated Database Instance."
	Value                                     = "Value given to the query limit."
	OverrunPolicy                             = "Action to take when the usage limit is exceeded."
```

internal/flag/flags.go
```diff
@@ -354,4 +354,7 @@ const (
 	CompactResponseShort                      = "c"                                         // CompactResponseShort flag
 	AWSRoleID                                 = "awsRoleId"                                 // AWSRoleID flag
 	AWSTestS3Bucket                           = "awsTestS3Bucket"                           // AWSTestS3Bucket flag
+	DataFederation                            = "dataFederation"                            // DataFederation flag
+	Value                                     = "value"                                     // Value flag
```
tibulca, line 358: Can we call it `QueryLimitValue`? Otherwise, we have the risk to be used by other unrelated commands in the future

build/ci/evergreen.yml
```diff
@@ -857,6 +857,27 @@ tasks:
           E2E_TAGS: atlas,datafederation,privatenetwork
           E2E_TEST_BUCKET: ${e2e_test_bucket}
           E2E_CLOUD_ROLE_ID: ${e2e_cloud_role_id}
+  # datafederation requires cloud provider role authentication and an s3 bucket created
+  - name: atlas_datafederation_querylimits_e2e
+    tags: ["e2e","datafederation","atlas"]
```
tibulca, line 862: question: should "querylimits" tag be in this list?
in the e2e we have `//go:build e2e || (atlas && datafederation && querylimits)`

internal/flag/flags.go
fmenezes: between --value and --queryLimitValue I believe --value is better, context informs what kind of value it is

tibulca: I was thinking about renaming only the flag constant name, not the value itself, so it won't be used accidentally from other commands (you'd have to read the internal/usage/usage.go to find out its query limit specific)
`QueryLimitValue = "value" `

fmenezes: Well, then what about other commands that also use —value? Are they going to use CommandXValue = “value” as well?

fmenezes: I don’t follow what is the benefit in multiple constants w/ the same value, that is my point here

tibulca: If another command will need a `value` param I think we'll have to create CommandXValue = “value” since the description of this field is specific to the query limit (so we won't be able to reuse it): `Value = "Value given to the query limit"`.

this file (flags.go) contains all flags, not only for query limit.
In this list, the flag name is very generic `Value` and does not say anything about its purpose (query limit value), while in usage.go is very specific `Value = "Value given to the query limit"`.
To avoid confusion, I was suggesting to just renaming it to provide more context, same as we have for example for `OperatorProjectName  = "projectName"`

An alternative would be to set a more generic description instead of `Value = "Value given to the query limit"`

fmenezes: Your example is from usage.go where we indeed have CommandXSomething = "explanation" but your comment is in flags.go where really we only have the flags name in --name.

That is why I'm a bit confused, for flags.go there is no distinction, but for usage.go there is.

Which one are you talking about?

internal/usage/usage.go
```diff
@@ -436,5 +436,8 @@ dbName and collection are required only for built-in roles.`
 	WatchTimeout                              = "Time in seconds until a watch times out. After a watch times out, the CLI no longer watches the command."
 	CompactResponse                           = "Flag that enables the compact array response structure for a json output. The --compact option returns array objects as top-level responses and allows backward compatibility for scripts based on previous CLI versions. Omitting the --compact option for a json output returns array objects within a 'results' sub-array. You must specify --output json to use this option."
 	DataFederationFile                        = "Path to an optional JSON configuration file that defines data federation settings."
-	DataFederationType                        = "Type of Federated Database Instances to return"
+	DataFederationType                        = "Type of Federated Database Instances to return."
+	DataFederation                            = "Identifier of the Federated Database Instance."
+	Value                                     = "Value given to the query limit."
```
tibulca, line 441:
```suggestion
	QueryLimitValue                                     = "Value given to the query limit."
```

internal/flag/flags.go
tibulca: I initially thought we have a 1:1 between flags.go and usage.go but found I see the flags are defined more generic (e.g. APIKey = "apiKey") while the usage is specialized and the names don't match where more context has to be added (e.g. `DatadogAPIKey = "Datadog API key that allows Atlas to access your Datadog account."`). 
Sorry for the confusion, I'm closing this thread, as I added a suggestion in [usage.go](https://github.com/mongodb/mongodb-atlas-cli/pull/2037/files#r1243401205)

fmenezes: I guess some more context my help bring clarity. flags.go and usage.go were initial attempts to have docs team owning docs review in the CLI, later on we've introduced the docs folder with txt files from cobra2snooty which later became the review mechanism we have today where docs team owns txt files.

Now-a-days usage.go and flags.go have one purpose only which is standardize flag names across commands and explanations across sibling commands.

Does that help?

tibulca (approved):
LGTM, only one non-blocking comment

internal/usage/usage.go
fmenezes: I'll address on my next PR https://github.com/mongodb/mongodb-atlas-cli/pull/2038

PR #2046, opened 2023-06-27
fix: clean up atlas user commands

gssbzn:
## Proposed changes

After moving these commands to be atlas only we can remove some of the conditions we do for ops manager


blva (approved):
LGTM

PR #2051, opened 2023-06-28
CLOUDP-187013: Atlas CLI always includes list of namespaces for perforemance advisor

gssbzn:
## Proposed changes

_Jira ticket:_ CLOUDP-187013

## Further comments

Currently the performance advisor requires to specify a namespace or it always returns empty, we need to check if it was set by users before trying to make the api call to correctly ignore the empty value.

Additionally these commands seem to have leftover code from when shared with ops maanger like supporting both host id and process name, deprecating the ops manager options and correctly documenting the flags now


internal/cli/atlas/performanceadvisor/namespaces/list.go
```diff
@@ -105,7 +104,9 @@ If you don't set the duration option or the since option, this command returns d
 	}
 
 	cmd.Flags().StringVar(&opts.HostID, flag.HostID, "", usage.HostID)
+	_ = cmd.Flags().MarkDeprecated(flag.HostID, "Flag is invalid for MongoDB Atlas")
```
colm-quinn, line 107: ❤️ Thanks for helping our users here in case they were still using the flags in scripts

internal/cli/atlas/performanceadvisor/slowquerylogs/list.go
```diff
@@ -39,7 +39,7 @@ type ListOpts struct {
 	store      store.PerformanceAdvisorSlowQueriesLister
 	since      int64
 	duration   int64
-	namespaces string
+	namespaces []string
```
colm-quinn, line 42: [q] Is this a breaking change for users, or was no data being returned previously, so they never saw a string value here?

colm-quinn (approved):
LGTM, questions aren't blocking.

wtrocki (approved):

internal/cli/atlas/performanceadvisor/namespaces/list.go
gssbzn: yeah initially I was goign to delete then though it better to deprecate first

internal/cli/atlas/performanceadvisor/slowquerylogs/list.go
gssbzn: this is internal and has no impact to users except they used to do `--namespaces` once, they can do it multiple times now

PR #2052, opened 2023-06-28
CLOUDP-187071: Enable autocomplete for proceses in performance advisor

gssbzn:
## Proposed changes

_Jira ticket:_ CLOUDP-187071

## Further comments

Enable autocomplete when doing 

```bash
atlas performanceAdvisor slowQueryLogs list --processName <tab>
```

### DIFF
--- a/internal/cli/atlas/clusters/region_tier_autocomplete.go
+++ b/internal/cli/atlas/clusters/region_tier_autocomplete.go
@@ -27,8 +27,6 @@ import (
 	"github.com/spf13/cobra"
 )
 
-type autoFunc func(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective)
-
 type autoCompleteOpts struct {
 	cli.GlobalOpts
 	providers []string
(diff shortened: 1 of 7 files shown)


internal/cli/atlas/clusters/region_tier_autocomplete.go
```diff
@@ -69,7 +67,6 @@ func (opts *autoCompleteOpts) tierSuggestions(toComplete string) ([]string, erro
 	availableTiers := map[string]bool{}
 	for _, p := range result.Results {
 		for _, i := range p.InstanceSizes {
-			_ = i
```
gssbzn, line 72: @wtrocki I saw you added these, do you remember why?

wtrocki: It seems like leftover from workaround that have been required due to the SDK issue. 
We had that done in the long running PR

wtrocki (approved):

wtrocki:
e2e fails on unrelated issues

PR #2053, opened 2023-06-28
lint: update golangci-lint to 1.53.3

gssbzn:
## Proposed changes

Update to the latest version of the linter

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -64,7 +64,6 @@ linters-settings:
 linters:
   disable-all: true
   enable:
-    - depguard # Go linter that checks if package imports are in a list of acceptable packages [fast: true, auto-fix: false]
     - dogsled # Checks assignments with too many blank identifiers (e.g. x, _, _, _, := f()) [fast: true, auto-fix: false]
     - errcheck # Errcheck is a program for checking for unchecked errors in go programs. These unchecked errors can be critical bugs in some cases [fast: false, auto-fix: false]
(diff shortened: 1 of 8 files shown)


.golangci.yml
```diff
@@ -64,7 +64,6 @@ linters-settings:
 linters:
   disable-all: true
   enable:
-    - depguard # Go linter that checks if package imports are in a list of acceptable packages [fast: true, auto-fix: false]
```
gssbzn, line 67: this was never configured and it now complains about the imports we have, it's better to disable

internal/cli/output_opts.go
```diff
@@ -183,7 +183,7 @@ func template(outputType, value string) (string, error) {
 	if outputType == goTemplateFile {
 		data, err := os.ReadFile(value)
 		if err != nil {
-			return "", fmt.Errorf("%w: %s, %v", errTemplate, value, err)
+			return "", fmt.Errorf("%w: %s, %v", errTemplate, value, err.Error())
```
gssbzn, line 186: is either this or upgrade to go 1.20 which supports multi error format

test/e2e/helper.go
```diff
@@ -79,14 +79,14 @@ func DeleteProjectWithRetry(t *testing.T, projectID string) {
 	t.Helper()
 	deleted := false
 	for attempts := 1; attempts <= maxRetryAttempts; attempts++ {
-		if e := deleteProject(projectID); e != nil {
```
gssbzn, line 82: for some reason one of the linters now doesn't like the order here

tibulca (approved):
LGTM

PR #2054, opened 2023-06-29
CLOUDP-187012: Create Dockerfile for AtlasCLI

nicohrubec:
## Proposed changes

Jira ticket: CLOUDP-187012

Includes a Dockerfile to build a container running the CLI. Installation done with yum so that both the CLI and mongosh are included. In addition, adds a CI/CD that builds the container, which is triggered whenever there is a change to the Dockerfile.

## Setup

Build the container:
docker build -t atlas .

Run the container which will launch you into a shell:
docker -it run atlas

## Current issues:
Mongosh does not work when target platform for build and run is set to linux/amd64. Might be a local issue. Further investigation will be done in a follow-up.


Dockerfile
```diff
@@ -0,0 +1,17 @@
+FROM registry.access.redhat.com/ubi8/ubi-minimal:8.6
+
+COPY <<EOF /etc/yum.repos.d/mongodb-org-6.0.repo
```
wtrocki, line 3: Source for that is here: 
https://github.com/mongodb/mongodb-atlas-cli/tree/736315ce2a4e0b67fe664709691f95c8baf88716/build/package/docker

Dockerfile
```diff
@@ -0,0 +1,17 @@
...
+enabled=1
+gpgkey=https://www.mongodb.org/static/pgp/server-6.0.asc
+EOF
+
+RUN microdnf install shadow-utils jq yum yum-utils &&\
+    yum -y update &&\
+    yum clean all &&\
+    microdnf clean all
+
+RUN yum install -y mongodb-atlas
```
wtrocki, line 17: We missing Entripoint and exec. That leaves the option on how the command is run:

- Bash shell is running and it runs atlas-cli. That can run 
- Just atlas-cli is running without shell (non-TTY)

I will recommend to experiment with shell and learn about entry points for CLI docker images. 

Try to answer questions like:
Do we want to use docker run for every command or we start bash shell where atlas cli can run?

.github/workflows/docker.yaml
```diff
@@ -0,0 +1,36 @@
...
+      - name: Build
+        run: make build-all                      
+      - name: Build Image
+        run: docker buildx build -t atlas-cli --platform=linux/amd64 .
+      ## Disabled for now
+      # - name: Login
+      #   run: docker login -u="${{secrets.CONTAINER_USER}}" -p="${{secrets.CONTAINER_SECRET}}" quay.io
+      # - name: Push To quay.io
+      #   id: push-to-quay
+      #   run: docker push quay.io/mongodb/atlas-cli
```
matt-condon, line 33: Let's omit this github workflow file from this PR
We can introduce it in a future follow-up story

nicohrubec: Removed

Dockerfile
nicohrubec: I added an entrypoint for now that executes the atlas binary, as discussed still up for debate how this should be done dependant on how most people will use the image

Dockerfile
```diff
@@ -0,0 +1,29 @@
...
+
+ENTRYPOINT ["atlas"]
```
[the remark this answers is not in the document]
nicohrubec: I updated the "Setup" section in the description

Dockerfile
```diff
@@ -0,0 +1,29 @@
...
+enabled=1
+gpgkey=https://www.mongodb.org/static/pgp/server-6.0.asc
+EOF
+
+RUN microdnf install shadow-utils jq yum yum-utils &&\
+    yum -y update &&\
+    yum clean all &&\
+    microdnf clean all
+
+RUN yum install -y mongodb-atlas
```
wtrocki, line 26: This will run only once with cache unless we specify the version explicitly.

Dockerfile
wtrocki: Sounds good. We can change that during development time.

Dockerfile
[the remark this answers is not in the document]
wtrocki: LGTM. We might need to run it with the name to avoid unnecessary container creation.

matt-condon:
> Would it be possible to have GitHub actions to build a container in order to ensure that docker file changes are correct.
> Github action can run only when dockerfile is modified so it will run rarely. Asking @matt-condon as we had requested to remove it?

Sure that SGTM

wtrocki:
Point 3 and 3.1 seems to be solved now.

Dockerfile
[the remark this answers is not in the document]
nicohrubec: will add developer facing docs on how to run in a follow up (when everything works as it should)

.github/workflows/docker.yml
```diff
@@ -0,0 +1,25 @@
...
+        golang: [1.19]
+    steps:
+      - name: Check out the repo
+        uses: actions/checkout@v3
+      - name: Install Go
+        uses: actions/setup-go@v3
+        with:
+          go-version: ${{ matrix.golang }}
+      - name: Build
+        run: make build-all
```
wtrocki, line 23:
```suggestion
```
Not relevant for now.

.github/workflows/docker.yml
```diff
@@ -0,0 +1,25 @@
+name: Build CLI Docker image
+
+on:
+  push:
+    paths:
+      - 'Dockerfile'
```
wtrocki, line 6:
```suggestion
      - './Dockerfile'
```

Otherwise we going to match files in the build folder

.github/workflows/docker.yml
```diff
@@ -0,0 +1,25 @@
...
+    strategy:
+      matrix:
+        golang: [1.19]
+    steps:
+      - name: Check out the repo
+        uses: actions/checkout@v3
+      - name: Install Go
+        uses: actions/setup-go@v3
+        with:
+          go-version: ${{ matrix.golang }}
```
wtrocki, line 21:
```suggestion
```

Not needed.

.github/workflows/docker.yml
```diff
@@ -0,0 +1,25 @@
...
+    paths:
+      - 'Dockerfile'
+
+jobs:
+  build_image:
+    name: Build docker image
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        golang: [1.19]
```
wtrocki, line 14:
```suggestion
```

.github/workflows/docker.yml
```diff
@@ -0,0 +1,25 @@
+name: Build CLI Docker image
+
+on:
+  push:
+    paths:
+      - 'Dockerfile'
+
+jobs:
+  build_image:
```
wtrocki, line 9:
```suggestion
    verify_image:
```

.github/workflows/docker.yml
```diff
@@ -0,0 +1,16 @@
...
+
+jobs:
+  verify_image:
+    name: Build docker image
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out the repo
+        uses: actions/checkout@v3
+      - name: Build Image
+        run: docker build -t atlas .
```
wtrocki, line 16:
```suggestion
        run: docker build --no-cache -t atlas .
```

Context: When we want to verify image but do not push it we should ignore cache to make sure that every layer works correctly with each other. In production scenario we always want cache as this affects end users.

wtrocki: Just FYI We rely here on the docker available in the actions platform (which is ok)
We should likely use github action with a controllable version of the container engine for production builds.

Dockerfile
```diff
@@ -0,0 +1,28 @@
...
+[mongodb-org-aarch64-6.0]
+name=MongoDB Repository
+baseurl=https://repo.mongodb.org/yum/redhat/8/mongodb-org/6.0/aarch64/
+gpgcheck=1
+enabled=1
+gpgkey=https://www.mongodb.org/static/pgp/server-6.0.asc
+EOF
+
+RUN microdnf install shadow-utils jq yum yum-utils &&\
+    yum -y update &&\
```
wtrocki, line 22: Worth revisiting if all of this is needed and relevant.


We have a [base](https://catalog.redhat.com/software/containers/ubi8/ubi-minimal/5c359a62bed8bd75a2c3fba8?container-tabs=security) that should be security updated and now we also perform updates on our own that creates another layer. If we use cache that layer would not be updated unless something above changes.

nicohrubec: I think only yum is absolutely necessary, the rest can be removed.

With the updates do you mean yum -y update?

Dockerfile
nicohrubec: So you think adding the version explicitly is the way to go? It makes sense because of the caching, I am just not yet sure how we can automatically obtain the latest version during the rebuilds, since we probably do not want to update this manually every time there is a new cli version.

.github/workflows/docker.yml
```diff
@@ -0,0 +1,16 @@
...
+
+jobs:
+  verify_image:
+    name: Build docker image
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out the repo
+        uses: actions/checkout@v3
+      - name: Build Image
+        run: docker buildx build --no-cache -t atlas .
```
andreaangiolillo, line 16: Can we run docker linter as part of the action? https://github.com/RedCoolBeans/dockerlint

nicohrubec: I'll look into that

Dockerfile
```diff
@@ -0,0 +1,29 @@
...
+EOF
+
+COPY <<EOF /etc/yum.repos.d/mongodb-org-aarch64-6.0.repo
+[mongodb-org-aarch64-6.0]
+name=MongoDB Repository
+baseurl=https://repo.mongodb.org/yum/redhat/8/mongodb-org/6.0/aarch64/
+gpgcheck=1
+enabled=1
+gpgkey=https://www.mongodb.org/static/pgp/server-6.0.asc
+EOF
```
andreaangiolillo, line 20: I see that the current dockerfile uses the cli version from the official mongoDB repository. Do we want to provide another dockerfile to download the cli repo and build the binary with the makefile? In this way, we will be able to test the unreleased cli versions as well.

tagging @matt-condon as I don't think this was agreed upon in the scope. (As a result, the answer "It is not in the scope so we won't do it" is more than acceptable here.)

matt-condon: > Do we want to provide another dockerfile to download the cli repo and build the binary with the makefile?

For local dev experience or for validation when the Dockerfile changes?

I think this could be considered a nice-to-have
I'll create a low-priority story for this and backlog it later if we consider it out-of-scope

andreaangiolillo: > For local dev experience or for validation when the Dockerfile changes?

for both

> I think this could be considered a nice-to-have
> I'll create a low-priority story for this and backlog it later if we consider it out-of-scope

Sure. Thanks!

matt-condon: https://jira.mongodb.org/browse/CLOUDP-188091

andreaangiolillo: Something to keep in mind when using the CLI from the MongoDB repo (this is related to [CLOUDP-187063: Override agent for Dockerized AtlasCLI](https://jira.mongodb.org/browse/CLOUDP-187063)) : we inject the information shown in the `UserAgent` field  at compilation time during the release [.atlascli.goreleaser.yml](https://github.com/mongodb/mongodb-atlas-cli/blob/09a1900f0c02caa2df2a8321bec4eb413393786a/build/package/.atlascli.goreleaser.yml#L17) or when running [make build-atlascli](https://github.com/mongodb/mongodb-atlas-cli/blob/09a1900f0c02caa2df2a8321bec4eb413393786a/Makefile#L135). If you use the CLI from the MongoDB repo, you won't be able to override these variables when running in Docker.

andreaangiolillo: One workaround may be to set the default value of [ToolName](https://github.com/mongodb/mongodb-atlas-cli/blob/5d4dd4bf038f41d7496e353f4a6083c0aa74e51b/internal/config/profile.go#L75C2-L75C10) to `atlascli-container` instead of `MongoCLI` but this should be investigated.

.github/workflows/docker.yml
nicohrubec: Thanks for the suggestion, I have added hadolint to the pipeline

matt-condon (approved):

Dockerfile
nicohrubec: Yes thanks for the pointers, I will investigate that as part of https://jira.mongodb.org/browse/CLOUDP-187063

andreaangiolillo (approved):
LGTM. Thanks for addressing the comments and for the discussion 💯

PR #2057, opened 2023-06-29
CLOUDP-187332: update sdk go to the latest release

wtrocki:
## Intention

Support versioned API format for releases. Integrate with the last release that removes all remaining OneOf/polymorphic objects.

## Testing

I did perform manual testing for affected commands.
All e2e tests are passing. 
Continuing exteding e2e tests and manual tests. 

## SDK release updates

SDK is now released using a major version prefix. I have added a note to Contributing.md about what it means.

## Simplification of the API

Changes removed a number of switch cases that would have required a lot of future maintenance. 
For clarity I have flagged most of the areas.

> NOTE: the amount of files changed is large due to import updates. 
Please ignore the first 2 commits when reviewing the code.


CONTRIBUTING.md
```diff
@@ -195,3 +195,15 @@ Reviewers, please ensure that the CLA has been signed by referring to [the contr
 
 For changes that involve user facing copy please include `docs-cloud-team` as a reviewer.
 
+## SDK integration
```
wtrocki, line 198: Important change with new SDK

internal/cli/atlas/backup/snapshots/create.go
```diff
@@ -55,7 +55,7 @@ func (opts *CreateOpts) Run() error {
 	if err != nil {
 		return commonerrors.Check(err)
 	}
-	return opts.Print(r.GetActualInstance())
```
wtrocki, line 58: Those were risky as we were returning subset of JSON as oposed to whole JSON in manual SDK

internal/kubernetes/operator/project/project.go
```diff
@@ -446,28 +446,7 @@ func buildIntegrations(intProvider store.IntegrationLister, projectID, targetNam
 }
 
 func getIntegrationType(val atlasv2.ThridPartyIntegration) string {
```
wtrocki, line 448: Key simplicity of the new SDK

internal/kubernetes/operator/project/project.go
```diff
@@ -421,12 +421,12 @@ func buildIntegrations(intProvider store.IntegrationLister, projectID, targetNam
...
 		case newRelicIntegrationType:
 			integration.LicenseKeyRef = secretRef
-			secret.Data[secrets.PasswordField] = []byte(list.NewRelic.LicenseKey)
+			secret.Data[secrets.PasswordField] = []byte(list.GetLicenseKey())
 			// Secrets with write and read tokens
 			var writeToken, readToken string
 			if includeSecrets {
-				writeToken = list.NewRelic.WriteToken
-				readToken = list.NewRelic.ReadToken
+				writeToken = list.GetWriteToken()
```
wtrocki, line 428: We should always use Getters to access fields

internal/mocks/mock_private_endpoints.go
```diff
@@ -73,10 +73,10 @@ func (m *MockPrivateEndpointDescriber) EXPECT() *MockPrivateEndpointDescriberMoc
 }
 
 // PrivateEndpoint mocks base method.
-func (m *MockPrivateEndpointDescriber) PrivateEndpoint(arg0, arg1, arg2 string) (interface{}, error) {
+func (m *MockPrivateEndpointDescriber) PrivateEndpoint(arg0, arg1, arg2 string) (*admin.EndpointService, error) {
```
wtrocki, line 76: Moving from interfaces to explicit types in mocks and stores helps us with future maintenance. 
We should move them to atlas

internal/store/atlas/teams.go
```diff
@@ -15,7 +15,7 @@
 package atlas
 
 import (
-	atlasv2 "go.mongodb.org/atlas-sdk/admin"
+	atlasv2 "go.mongodb.org/atlas-sdk/v20230201001/admin"
```
wtrocki, line 18: Every import has now major version. 
🥇 We can update by replacing imports across the board
☄️  for compatibility we can leave old imports when we see breaking changes for end users
🥈  we can use multiple versions of SDK in the CLI and update on demand ensuring consistency, backwards compatibility and also autodiscovery of new changes

wtrocki:
Fixing linting and unit tests

internal/cli/atlas/backup/snapshots/create_test.go
```diff
@@ -34,9 +34,7 @@ func TestCreateOpts_Run(t *testing.T) {
 
 	t.Run("diskBackupReplicaSet run", func(t *testing.T) {
 		expected := &atlasv2.DiskBackupSnapshot{
-			DiskBackupReplicaSet: &atlasv2.DiskBackupReplicaSet{
```
wtrocki, line 37: Some of the oneOfs contained only single object which made SDK overly verbose and introduced unneded nested structure

internal/cli/atlas/datalakepipelines/availablesnapshots/list.go
```diff
@@ -33,7 +33,7 @@ import (
 )
 
 var listTemplate = `ID	DESCRIPTION	STATUS{{range .Results}}
-{{if .DiskBackupReplicaSet}}{{.DiskBackupReplicaSet.Id}}	{{.DiskBackupReplicaSet.Description}}	{{.DiskBackupReplicaSet.Status}}{{else}}{{.DiskBackupShardedClusterSnapshot.Id}}	{{.DiskBackupShardedClusterSnapshot.Description}}	{{.DiskBackupShardedClusterSnapshot.Status}}{{end}}{{end}}`
```
wtrocki, line 36: Good example on challenges we were getting with templates and polymorphic objects.

internal/cli/atlas/integrations/create/webhook.go
```diff
@@ -58,11 +58,9 @@ func (opts *WebhookOpts) Run() error {
 
 func (opts *WebhookOpts) newWebhookIntegration() *atlasv2.ThridPartyIntegration {
 	return &atlasv2.ThridPartyIntegration{
-		Webhook: &atlasv2.Webhook{
-			Type:   &webhookIntegrationType,
-			Url:    opts.url,
-			Secret: &opts.secret,
-		},
+		Type:   &webhookIntegrationType,
```
wtrocki, line 61: With merge of the fields we lost information about what fields are part of the object.
I have POC that will help us to flag objects in the SDK: https://github.com/mongodb/atlas-sdk-go/pull/117

go.mod
```diff
@@ -40,7 +40,7 @@ require (
 	github.com/tangzero/inflector v1.0.0
 	github.com/withfig/autocomplete-tools/packages/cobra v1.2.0
 	go.mongodb.org/atlas v0.30.0
-	go.mongodb.org/atlas-sdk v0.15.0
+	go.mongodb.org/atlas-sdk/v20230201001 v20230201001.0.1-0.20230704132156-6560cf3b3bce
```
matt-condon, line 43: Q. Shouldn't this just be `go.mongodb.org/atlas-sdk/v20230201001.0.1`?

internal/cli/atlas/networking/peering/watch.go
```diff
@@ -57,28 +57,28 @@ func (opts *WatchOpts) watcher() (bool, error) {
...
-func watcherGCP(peer *atlasv2.GCPNetworkPeeringConnectionSettings) bool {
-	return *peer.Status == waitingForUser || *peer.Status == failed || *peer.Status == available
+func watcherGCP(peer *atlasv2.BaseNetworkPeeringConnectionSettings) bool {
+	return peer.GetStatus() == waitingForUser || peer.GetStatus() == failed || peer.GetStatus() == available
 }
 
-func watcherAzure(peer *atlasv2.AzureNetworkPeeringConnectionSettings) bool {
-	return *peer.Status == waitingForUser || *peer.Status == failed || *peer.Status == available
+func watcherAzure(peer *atlasv2.BaseNetworkPeeringConnectionSettings) bool {
+	return peer.GetStatus() == waitingForUser || peer.GetStatus() == failed || peer.GetStatus() == available
```
wtrocki, line 77: I did ensure we use getters in many places and avoid pointers. Getters will prevent panic errors in the future.

internal/store/containers.go
```diff
@@ -1,170 +0,0 @@
-// Copyright 2020 MongoDB Inc
```
wtrocki, line 1: Container store have been wrongly pointing to peering connections store. 
Since both methods are available under the same package i have moved those methods for clarity

internal/store/peering_connections.go
```diff
@@ -98,11 +102,124 @@ func (s *Store) DeletePeeringConnection(projectID, peerID string) error {
 }
 
 // CreatePeeringConnection encapsulates the logic to manage different cloud providers.
-func (s *Store) CreatePeeringConnection(projectID string, peer *atlasv2.BaseNetworkPeeringConnectionSettings) (interface{}, error) {
+func (s *Store) CreatePeeringConnection(projectID string, peer *atlasv2.BaseNetworkPeeringConnectionSettings) (*atlasv2.BaseNetworkPeeringConnectionSettings, error) {
```
wtrocki, line 105: We no longer need interface as we do not deal with multiple instances anymore.

internal/store/peering_connections.go
```diff
@@ -98,11 +102,124 @@ func (s *Store) DeletePeeringConnection(projectID, peerID string) error {
...
 	switch s.service {
 	case config.CloudService, config.CloudGovService:
 		result, _, err := s.clientv2.NetworkPeeringApi.CreatePeeringConnection(s.ctx, projectID, peer).Execute()
-		return result.GetActualInstance(), err
```
wtrocki, line 105: This would have caused issues in the future where parent fields would not have been included in results.

gssbzn (commented):
Given the past experience on errors introduced because of mismatch between the new structs and the output templates how are we validating here that all templates still work as expected? I would love to see unit test at least for the templates that are being modified

internal/cli/atlas/privateendpoints/azure/watch.go
```diff
@@ -51,7 +50,7 @@ func (opts *WatchOpts) watcher() (bool, error) {
 		return false, err
 	}
 
-	endpointService := result.(*atlasv2.EndpointService)
+	endpointService := result
```
matt-condon, line 53: nit; can remove `endpointService` and replace uses with `result`

wtrocki:
Fully agree. Unit tests would not be suitable for this - we need e2e coverage. 
I'm looking to add snapshot tests.

matt-condon (approved):
LGTM, really nice improvement ⭐

go.mod
wtrocki: I forgot to explain it. 
This is using an unreleased version. 
Once I merge I want to verify if dependabot is going to pick up next major release of the SDK.
I need to merge this with unrelased version and will replace it almost instantly with released one.

matt-condon (dismissed):

gssbzn:
> we need e2e coverage.

E2E only covers json output and why we still had a lot of issues before and we still keep seeing them

wtrocki:
> E2E only covers json output and why we still had a lot of issues before and we still keep seeing them

I did apply changes to unit tests following your advice. 

> E2E only covers json output

That is exactly my point. Extending e2e tests to validate both outputs should be trivial by helper methods + snapshotting, 
but it is not the goal of this PR.

gssbzn:
>Extending e2e tests to validate both outputs can be done by helper methods + snapshotting,

I'm not sure how trivial this would be, the cli doesn't support multiple outputs at the same time and we use json to make it easier to parse so you'll probably goign to end up calling all tests at least twice, once output, which may not be a bad idea on its own but it does add a lot of extra test specially around things that are slow like clusters

wtrocki:
@gssbzn Looking for your feedback on changes https://github.com/mongodb/mongodb-atlas-cli/pull/2057/commits/788720bb085cffca31cc85f5831f06645499283e 

If that is ok I will apply those across entire CLI. This change will be temporary as I believe that in long run we should not use unit tests and rely on libraries like: https://pkg.go.dev/github.com/jba/templatecheck#CheckText

gssbzn:
>@gssbzn Looking for your feedback on changes https://github.com/mongodb/mongodb-atlas-cli/commit/788720bb085cffca31cc85f5831f06645499283e

at face value I'm unsure what advantage the new lib has over the string

>This check/change will be temporary as I believe that in the long run we should not use unit tests and rely on libraries like: https://pkg.go.dev/github.com/jba/templatecheck#CheckText for table format validation

This one seems more promising tho, any reason to not go straight here instead of the intermediate step? i sok to ignore current tests that rely on the string approach 

IF you think the snapshot is worth it then do it

wtrocki:
I have a strong opinion about introducing a lot of boilerplate code causing maintenance issues using string builder. 
I hit so many issues related to updates of the format or hidden tabs or spaces and polluting test cases.

Snapshots are very common solution to solve that problem - additionally having that library in place will allow us to use snapshots to validate contracts. This will really raise our excellence in the testing capability and allow us to do more dedicated tests with limited flakiness. 
Being mindful of the external library introduced I wanted to solve both problems with the single library (minimize security updates overhead and approval process). 

WDYT about removing snapshot testing from the current PR and merging it as it is?
Then I will be working instantly to supply a solution that doesn't introduce maintenance overhead.

I'm mindfull that this PR is already very large and I do not want to introduce another +100 tests files rewrites

gssbzn:
regardless of approach this questions stand 

https://github.com/mongodb/mongodb-atlas-cli/pull/2057#pullrequestreview-1518583859

Are we ok to do a release after merging this PR?

wtrocki:
New library that I have added is quite powerful and doesn't require much code:

>--- FAIL: TestListOpts_Run (0.00s)
    /Users/w.trocki/Projects/cloud/mongodb-atlas-cli/internal/cli/atlas/backup/exports/buckets/list_test.go:59: Run() unexpected error: template: output:2:26: executing "output" at <.CloudProviderf>: can't evaluate field CloudProviderf in type admin.DiskBackupSnapshotAWSExportBucket
    
 Pushing changes covering impact on this PR

wtrocki:
For visibility, I will cover the rest of the non list commands in separate PR.
With nonlist commands, we are pretty confident in the output as we always get the whole object and verified them. 
 It is the list that can have multiple objects/variations returned based on the server side and this have been the solve source of the issues reported across the cli.

go.mod
```diff
@@ -59,6 +58,11 @@ require (
 	sigs.k8s.io/yaml v1.3.0
 )
 
+require (
+	github.com/jba/templatecheck v0.6.0
+	go.mongodb.org/atlas-sdk/v20230201002 v20230201002.0.0
+)
```
gssbzn, line 64: any reason to group these separately? can we move them to the general group?

internal/cli/atlas/accesslogs/list_test.go
```diff
@@ -69,11 +68,8 @@ func TestAccessLogListClusterName_Run(t *testing.T) {
 	if err := opts.Run(); err != nil {
 		t.Fatalf("Run() unexpected error: %v", err)
 	}
-
-	assert.Equal(t, `HOSTNAME    AUTH RESULT   LOG LINE 
-test test   true          test
-`, buf.String())
 	t.Log(buf.String())
+	test.VerifyOutputTemplate(t, listTemplate, expected)
```
gssbzn, line 72: 🥇

internal/cli/atlas/integrations/describe_test.go
```diff
@@ -61,10 +58,14 @@ func TestDescribe_Run(t *testing.T) {
 	if err := describeOpts.Run(); err != nil {
 		t.Fatalf("Run() unexpected error: %v", err)
 	}
-	assert.Equal(t, `TYPE    API TOKEN   TEAM       CHANNEL
-SLACK   testToken   testTeam    testChannel 
-`, buf.String())
 	t.Log(buf.String())
+	test.VerifyOutputTemplate(t, describeTemplateDatadogOpsGenie, expected)
```
gssbzn, line 62: this can be represented as a loop over the templates

internal/cli/atlas/networking/peering/watch.go
```diff
@@ -57,28 +57,28 @@ func (opts *WatchOpts) watcher() (bool, error) {
...
-		return watcherAzure(v), nil
-	case *atlasv2.GCPNetworkPeeringConnectionSettings:
-		return watcherGCP(v), nil
+	switch v := result.GetProviderName(); v {
+	case "AWS":
+		return watcherAWS(result), nil
+	case "AZURE":
+		return watcherAzure(result), nil
+	case "GCP":
+		return watcherGCP(result), nil
```
gssbzn, line 66: the swtch was to allow casting to the actual instance, is this switch still necessary? feels like no

internal/cli/atlas/security/ldap/verify.go
```diff
@@ -51,7 +51,7 @@ func (opts *VerifyOpts) initStore(ctx context.Context) func() error {
 	}
 }
 
-var verifyTemplate = `REQUEST ID	PROJECT ID	STATUS
+var VerifyOutputTemplate = `REQUEST ID	PROJECT ID	STATUS
```
gssbzn, line 54: any reason we want to export this?

go.mod
wtrocki: Very good point
I did that a couple times.  go mod tidy reverts it. I will investigate further.

internal/cli/atlas/integrations/describe_test.go
wtrocki: I thought of having templates defined as array in unit test. Alternatively array can be defined in the command itself (but would not be used)

internal/kubernetes/operator/project/project.go
gssbzn: I think we should delete this method it doesn't abstract anything anymore

internal/kubernetes/operator/project/project.go
```diff
@@ -545,49 +524,50 @@ func buildNetworkPeering(npProvider store.PeeringConnectionLister, projectID str
...
+func convertNetworkPeer(np atlasv2.BaseNetworkPeeringConnectionSettings, providerName provider.ProviderName) atlasV1.NetworkPeer {
+	switch np.GetProviderName() {
+	case "AWS":
+		return convertAWSNetworkPeer(&np, providerName)
+	case "GCP":
+		return convertGCPNetworkPeer(&np, providerName)
+	case "Azure":
+		return convertAzureNetworkPeer(&np, providerName)
+	default:
+		return atlasV1.NetworkPeer{}
```
gssbzn, line 536: related to another comment, these switches allowed for type casting, they don't do that anymore and seems like safe to delete

internal/cli/atlas/networking/peering/watch.go
wtrocki: Do you mean replacing the switch with ifs?
GolangCiLint forces us to use a switch for more than 2 If/else :(
I also investigated if we can ignore provider details in this case but it did generate invalid behaviour for GCP.

internal/test/test.go
```diff
@@ -47,3 +49,16 @@ func CleanupConfig() {
 	config.SetOrgID("")
 	config.SetProjectID("")
 }
+
+// VerifyTemplate validates that the given template string is valid.
+func VerifyOutputTemplate(t *testing.T, templateString string, expected interface{}) {
```
gssbzn, line 54:
```suggestion
func VerifyOutputTemplate(t *testing.T, template string, expected interface{}) {
```
no need to suggest the type, that's what the compiler is for

internal/test/test.go
```diff
@@ -47,3 +49,16 @@ func CleanupConfig() {
...
+	t.Helper()
+	parsedTemplate, err := template.New("output").Parse(templateString)
+	if err != nil {
+		t.Fatalf("Failed to validate table format template: %v", err)
+	}
+	err = templatecheck.CheckText(parsedTemplate, expected)
+	if err != nil {
+		t.Fatalf("Failed to validate table format template: %v", err)
+	}
```
gssbzn, line 63: FYI since you have fatal here the t.Log you usually run after won't be called

wtrocki: Good point. Based on that - do we care to actually print rendered template to the end users. If no then I will remove logger for buffer. In most of the cases I did moved it to be the first print after this asert.

internal/cli/atlas/networking/peering/watch.go
gssbzn: watcherAWS, watcherAzure and watcherGCP( are all now the same method, I don't think we need this switch or three methods that do the same

internal/test/test.go
wtrocki: Reason: golangcilint reports mixed import and variable name.

templateInput is ok?

wtrocki: Naming is hard :)

internal/cli/atlas/networking/peering/watch.go
gssbzn: on second look only azure and gcp are the same, aws is slightly different, so we can combine azure and gcp and move from a switch to an if but the change of the switch is not necessary, it was before if the methods were the same

internal/cli/atlas/integrations/describe_test.go
gssbzn: no strong opinion here, it's a bit of a test simplification to even maybe use a `t.Run` to encapsulate the cases

internal/test/test.go
```diff
@@ -47,3 +49,16 @@ func CleanupConfig() {
 	config.SetOrgID("")
 	config.SetProjectID("")
 }
+
+// VerifyTemplate validates that the given template string is valid.
```
gssbzn, line 53:
```suggestion
// VerifyOutputTemplate validates that the given template is valid for the given value.
```

internal/test/test.go
wtrocki: Very clear! Applying it.
The reason we have this helper is to not be strongly dependent on the library we use.

gssbzn: yeah, this helper makes perfect sense 💯

internal/test/test.go
gssbzn: I have no strong opinion on any way, either loosing this to `t.Errorf`, moving the logger to be before we call this function or deleting the logger, they all seems valid with the information I have

internal/cli/atlas/networking/peering/watch.go
wtrocki: yes. This is the tradeof between verbosity and type clarity. If we merge both

internal/kubernetes/operator/project/project.go
wtrocki: @gssbzn We are still in a position to change all *strings to string in SDK at the cost of knowing if the field is optional or not.  Leaving this as topic for discussion.

gssbzn: the sad face is because Getters are non idiomatic for go and seems like a code small introduced by the generator

wtrocki: @gssbzn I'm adding this item to the agenda:

```
Optionality of the Strings in SDK
Optional Strings are represented as *string.
Result: We have to check for nullability and have nonstandard parsing behavior.
Q: Can we sacrifice optionality feature (less clarity) for strings (or even other types like integer, float), for the cost of less boilerplate code?
```

internal/test/test.go
wtrocki: Investigating why we did not got golangcilint issue :)

internal/kubernetes/operator/project/project.go
gssbzn: we were using pointers on the manual SDK and we didn't implement getters, like I said the current implementation of having an exported field and a getter is a bit of a code smell for go having a public field but also a public getter that's not really a gtter but a "safe" getter while not being clear on it seems a bit of a weird design choice 

```go
type A struct {
  A *string
}

func (a *A) GetA() string {
  return &a.A
}

func main() {
  a := &A{}
  a.A
  a.GetA() 
}
```

internal/test/test.go
gssbzn: this linter is disabled by default, the maintainers of golangcilint don't believe in it's value

internal/kubernetes/operator/project/project.go
wtrocki: By removing those we will lose information of what consists an type.

internal/kubernetes/operator/project/project.go
wtrocki: I understand. My question is - Can we leverage change for strings to be non pointer so then we do not need to have safe methods or use `pointer.Get(...)`. 

If we do that we going to lose information on what is an optional field.

internal/test/test.go
wtrocki: Addressed outside this function by moving one print that happens after fatal. 
I think it is good to have fatal being part of the test (it matches definition of the test helpers)

internal/cli/atlas/networking/peering/watch.go
wtrocki: I would prefer to not address it as it would affect support for future provider changes.

matt-condon (approved):

PR #2071, opened 2023-07-05
CLOUDP-187063: Override agent for Dockerized AtlasCLI

nicohrubec:
## Proposed changes

Adds a host parameter to the UserAgent, which defaults to "native". If the CLI is run from the docker container the host parameter is set to "container".

Jira ticket: CLOUDP-187063
Closes: CLOUDP-187063

## Checklist

## Further comments

Testing was done manually by using a local throwaway pipeline.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,6 +1,6 @@
 # syntax=docker/dockerfile:1.3-labs
 FROM registry.access.redhat.com/ubi8/ubi-minimal:8.6
-ENV CONTAINER=""
+ENV MONGODB_ATLAS_RUNNING_FROM_CONTAINER=""
 
 COPY <<EOF /etc/yum.repos.d/mongodb-org-x86_64-6.0.repo
 [mongodb-org-x86_64-6.0]
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,6 @@
 # syntax=docker/dockerfile:1.3-labs
 FROM registry.access.redhat.com/ubi8/ubi-minimal:8.6
+ENV BUILD_CONTEXT=container
 
 COPY <<EOF /etc/yum.repos.d/mongodb-org-x86_64-6.0.repo
 [mongodb-org-x86_64-6.0]
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,6 +1,6 @@
 # syntax=docker/dockerfile:1.3-labs
 FROM registry.access.redhat.com/ubi8/ubi-minimal:8.6
-ENV BUILD_CONTEXT=""
+ENV CONTAINER=""
 
 COPY <<EOF /etc/yum.repos.d/mongodb-org-x86_64-6.0.repo
 [mongodb-org-x86_64-6.0]
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,6 +1,6 @@
 # syntax=docker/dockerfile:1.3-labs
 FROM registry.access.redhat.com/ubi8/ubi-minimal:8.6
-ENV MONGODB_ATLAS_RUNNING_FROM_CONTAINER=""
+ENV MONGODB_ATLAS_RUNNING_FROM_CONTAINER=true
 
 COPY <<EOF /etc/yum.repos.d/mongodb-org-x86_64-6.0.repo
 [mongodb-org-x86_64-6.0]
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,6 +1,6 @@
 # syntax=docker/dockerfile:1.3-labs
 FROM registry.access.redhat.com/ubi8/ubi-minimal:8.6
-ENV MONGODB_ATLAS_RUNNING_FROM_CONTAINER=true
+ENV MONGODB_ATLAS_IS_CONTAINERIZED=true
 
 COPY <<EOF /etc/yum.repos.d/mongodb-org-x86_64-6.0.repo
 [mongodb-org-x86_64-6.0]
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,6 +1,6 @@
 # syntax=docker/dockerfile:1.3-labs
 FROM registry.access.redhat.com/ubi8/ubi-minimal:8.6
-ENV BUILD_CONTEXT=container
+ARG BUILD_CONTEXT
 
 COPY <<EOF /etc/yum.repos.d/mongodb-org-x86_64-6.0.repo
 [mongodb-org-x86_64-6.0]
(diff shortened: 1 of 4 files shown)


internal/store/store.go
```diff
@@ -269,8 +269,22 @@ func WithContext(ctx context.Context) Option {
...
+
 // setAtlasClient sets the internal client to use an Atlas client and methods.
 func (s *Store) setAtlasClient(client *http.Client) error {
+	buildContext, buildContextPresent := os.LookupEnv("BUILD_CONTEXT")
+
+	if buildContextPresent {
+		appendUserAgent(buildContext)
+	}
+
+	fmt.Println(config.UserAgent)
```
matt-condon, line 287: From what I can see, the UserAgent is set from 
https://github.com/mongodb/mongodb-atlas-cli/blob/master/internal/oauth/oauth.go#L70C11-L70C11

I wonder would it be better to pass the value from there, since it should only have to be called once when authenticating the CLI

internal/store/store.go
```diff
@@ -269,8 +269,22 @@ func WithContext(ctx context.Context) Option {
 	}
 }
 
+// Appends user agent given a build context for telemetry.
+func appendUserAgent(buildContext string) {
+	if !strings.Contains(config.UserAgent, buildContext) {
+		config.UserAgent = fmt.Sprintf("%s-%s", config.UserAgent, buildContext)
```
matt-condon, line 275: Since there's a space rather than a dash in the existing pattern, I think we should consider doing the same here to be consistent (left a comment on the ticket also)
https://github.com/mongodb/mongodb-atlas-cli/blob/master/internal/config/profile.go#L76

image

nicohrubec: Addressed

internal/store/store.go
nicohrubec: I agree it seems like the better place to put it, updated

internal/oauth/oauth.go
```diff
@@ -56,6 +60,15 @@ const (
 	GovClientID = "0oabtyfelbTBdoucy297" // GovClientID for production
 )
 
+// Appends user agent if run from container.
+func maybeAddUserAgentPostfix() {
+	_, runningFromContainer := os.LookupEnv("BUILD_CONTEXT")
+
+	if runningFromContainer && !strings.Contains(config.UserAgent, userAgentContainerPostfix) {
+		config.UserAgent = fmt.Sprintf("%s (%s)", config.UserAgent, userAgentContainerPostfix)
```
matt-condon, line 68: Per @drinkbird comment on the JIRA ticket 

> Thanks for flagging this. By doing some further reading, I see that what comes after "product/version" is "(comment)". We can extend what we have to include the environment in all cases.

Source: https://www.rfc-editor.org/rfc/rfc9110#name-parameters

nicohrubec: updated

internal/oauth/oauth.go
```diff
@@ -15,24 +15,31 @@
...
-	cloudGovServiceURL    = "https://cloud.mongodbgov.com/"
+	timeout                    = 5 * time.Second
+	keepAlive                  = 30 * time.Second
+	maxIdleConns               = 5
+	maxIdleConnsPerHost        = 4
+	idleConnTimeout            = 30 * time.Second
+	expectContinueTimeout      = 1 * time.Second
+	cloudGovServiceURL         = "https://cloud.mongodbgov.com/"
+	userAgentContainerHostName = "container"
+	userAgentNativeHostName    = "native"
```
matt-condon, line 38: Q: Can we set "native" by default in [profile.go](https://github.com/mongodb/mongodb-atlas-cli/blob/master/internal/config/profile.go#L76)?

Then conditionally replace it in this file?

internal/oauth/oauth.go
```diff
@@ -56,6 +63,13 @@ const (
 	GovClientID = "0oabtyfelbTBdoucy297" // GovClientID for production
 )
 
+func addUserAgentHostParameter() {
+	if !strings.Contains(config.UserAgent, userAgentHostName) {
+		config.UserAgent = fmt.Sprintf("%s;%s)",
```
wtrocki, line 68: FYI Config values can be overridden by existing env vars:
https://www.mongodb.com/docs/atlas/cli/stable/atlas-cli-env-variables/

We do not document user agent but it might make sense to use this mechanism?

Dockerfile
```diff
@@ -1,5 +1,6 @@
 # syntax=docker/dockerfile:1.3-labs
 FROM registry.access.redhat.com/ubi8/ubi-minimal:8.6
+ENV CONTAINER=""
```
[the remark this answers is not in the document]
wtrocki: Generic name with no value set. 

```suggestion
ENV MONGODB_ATLAS_USER_AGENT="atlascli-container"
```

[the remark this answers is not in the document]
nicohrubec: I agree that the naming of the var could be improved. I think setting a different value does not really make sense though, because the variable is used as a flag and the value is never used. I set it to an empty string only because env vars in Docker cannot be undefined. (At least I did not find a way)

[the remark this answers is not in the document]
nicohrubec: I updated the variable name to be more descriptive, lmk what you think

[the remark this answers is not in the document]
wtrocki: My point is that this PR could be just an `MONGODB_ATLAS_USER_AGENT="atlascli-container"` 
Variable name cannot be changed in this case as it is comming from documented feature: https://www.mongodb.com/docs/atlas/cli/stable/atlas-cli-env-variables/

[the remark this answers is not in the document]
wtrocki: Feel free to not use that feature (not blocking anything in this PR). 
My only strong opinion is to avoid empty variables. If you need flag like now:

```
MONGODB_ATLAS_RUNNING_FROM_CONTAINER=true
```

One could argue that it introduces some level of proliferation. 
We should likely look into what we want is the effect:

For example disabling atlas cli to run any external binaries 
```
MONGODB_ATLAS_DISABLE_EXTERNAL_COMMANDS=true
MONGODB_ATLAS_USER_AGENT="atlascli-container"
```

This is not only easier to implement and maintain. It does fit well with existing features and config.
Each of those ideally should be connected with the config and documented so users will be aware of the differences when running container etc.



CONTAINER or other values are subject

internal/oauth/oauth.go
nicohrubec: Seems reasonable I changed the approach

Dockerfile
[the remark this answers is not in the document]
nicohrubec: MONGODB_ATLAS_USER_AGENT="atlascli-container" doesn't seem to change the UserAgent in my container. So I changed the approach to set default in profile and then replace in case the new env var is set to true as suggested

nicohrubec:
Waiting for feedback from Jakub before merging

nicohrubec:
Approach got approved from Data Analytics

matt-condon (approved):
LGTM, nice work!

PR #2074, opened 2023-07-07
Approach for enforcing each unit tests to verify table format by default and without extensive manual mocking

wtrocki:
## Why

Atlas CLI needs to have tests for:
- Ensuring table format is correct out of the box for every list command with minimal boilerplate
- Contract tests for API changes.

Writing unit tests for the moment requires a lot of boilerplate and is going to cause long-term maintenance issues. 
Tests written manually become operating system dependant and hard to maintain.
In this PR I'm proposing using thrid party library that solves that problem.

## Approaches

In this scenario I'm using 2 approaches:

1. Inline objects + snapshots (integrations tests) Preferred.
2. Faked objects + snapshots (  reduces friction for us and makes our unit tests simple)

For long term we can explore: (proposal)

4. example objects generated in the SDK under test folder.
5. Remove table templates and use headers in sdk https://github.com/lensesio/tableprinter/blob/master/_examples/1_simple/main.go

### DIFF
--- a/internal/cli/atlas/backup/snapshots/__snapshots__/list_test.snap
+++ b/internal/cli/atlas/backup/snapshots/__snapshots__/list_test.snap
@@ -1,0 +1,13 @@
+
+[TestList_Run - 1]
+ID          TYPE         STATUS       CREATED AT                      EXPIRES AT
+NCEDUSB     aJiM         ZdBdAaR      0001-01-01 00:00:00 +0000 UTC   0001-01-01 00:00:00 +0000 UTC
+oXcSJ       hSuLlqw      khCxBJnVI    0001-01-01 00:00:00 +0000 UTC   0001-01-01 00:00:00 +0000 UTC
+AUcYNM      YYXZfvj      ZvLQDQRyT    0001-01-01 00:00:00 +0000 UTC   0001-01-01 00:00:00 +0000 UTC
+CVthPqiv    qQrf         xOLQxjxI     0001-01-01 00:00:00 +0000 UTC   0001-01-01 00:00:00 +0000 UTC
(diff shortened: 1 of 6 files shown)


wtrocki:
IMHO using Snapshots and manual expected objects is the way to go for now.

internal/cli/atlas/backup/snapshots/__snapshots__/list_test.snap
```diff
@@ -0,0 +1,13 @@
+
+[TestList_Run - 1]
```
wtrocki, line 2: This is snapshot generated using faker (As we see it is not giving us real data)

internal/cli/atlas/integrations/describe_test.go
```diff
@@ -59,9 +59,7 @@ func TestDescribe_Run(t *testing.T) {
 	if err := describeOpts.Run(); err != nil {
 		t.Fatalf("Run() unexpected error: %v", err)
 	}
-	assert.Equal(t, `TYPE    API TOKEN   TEAM       CHANNEL
-SLACK   testToken   testTeam   testChannel
-`, buf.String())
+	snaps.MatchSnapshot(t ,buf.String())
```
wtrocki, line 62: My preferred solution. It doesn't change much but we no longer need to deal with strings and have reliable way to view differences without dealing with different operating system line breaks or tabs

gssbzn: >It doesn't change much but we no longer need to deal with strings and have reliable way to view differences without dealing with different operating system line breaks or tabs

This wasn't a problem because it never went out to the os, it all stayed n memory in the string buffer

wtrocki:
Closing at this intended only to act as example

PR #2081, opened 2023-07-12
CLOUDP-188862: Automate publishing of container image

nicohrubec:
## Proposed changes

Adds CI/CD pipeline for publishing the CLI container to Dockerhub.
Sets up a cron job that runs once a day building and pushing images to Dockerhub for both ARM64 and x86_64 architectures. The image includes the latest CLI version + mongosh. 
The following tags are pushed:
- latest
- latest tag, example: mongodb/atlas:v1.10.0
- latest tag + date, example: mongodb/atlas:v1.10.0-2023-07-27

Pushing works as expected which can be seen in the repository: https://hub.docker.com/repository/docker/mongodb/atlas/general

Jira ticket: CLOUDP-188862

Closes CLOUDP-188862

## Checklist


matt-condon:
Can you share the test run from your fork for reference?

nicohrubec:
@matt-condon Updated the description with a link to the run on my fork

.github/workflows/docker-release.yml
```diff
@@ -0,0 +1,48 @@
+name: Daily build
+on:
+  push:
+    branches:
+      - '**'
```
andreaangiolillo, line 5:
```suggestion
  push:
    branches:
      - master
```

.github/workflows/docker-release.yml
```diff
@@ -0,0 +1,48 @@
...
+  schedule:
+    - cron: "0 1 * * *"
```
andreaangiolillo, line 7:
```suggestion
    - cron: "0 1 * * *" # Every day at 1:00 AM
```

.github/workflows/docker-release.yml
```diff
@@ -0,0 +1,48 @@
...
+jobs:
+  release_image:
+    name: Build and publish docker image
+    runs-on: ubuntu-latest
+    env:
+      IMAGE_REPOSITORY: mongodb/apix_test
+      PLATFORMS: linux/amd64,linux/arm64
+    steps:
+      - name: Check out code
+        uses: actions/checkout@v3.5.3
```
andreaangiolillo, line 18:
```suggestion
        uses: actions/checkout@v3
```

.github/workflows/docker-release.yml
andreaangiolillo: Why do we want to release the docker image all the time that we marge into main? 
What would happen if two PRs are merged at the same time?

.github/workflows/docker-release.yml
```diff
@@ -0,0 +1,48 @@
+name: Daily build
+on:
+  push:
+    branches:
+      - '**'
+  schedule:
+    - cron: "0 1 * * *"
+
```
andreaangiolillo, line 8: Let's add   `workflow_dispatch: {}` so you can also start the release manually if needed
```suggestion
  workflow_dispatch: {}
```

.github/workflows/docker-release.yml
```diff
@@ -0,0 +1,48 @@
...
+      - name: Set up QEMU
+        uses: docker/setup-qemu-action@v2
+        with:
+          platforms: ${{ env.PLATFORMS }}
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v2
+        with:
+          platforms: ${{ env.PLATFORMS }}
+      - name: Login to Docker Hub
+        uses: docker/login-action@v2.1.0
```
andreaangiolillo, line 37:
```suggestion
        uses: docker/login-action@v2
```

andreaangiolillo (commented):
Great work! Left some comments to understand a little more the plan

.github/workflows/docker-release.yml
nicohrubec: This is only to make development easier for me, will remove after I am done

.github/workflows/docker-release.yml
```diff
@@ -0,0 +1,55 @@
...
+          platforms: ${{ env.PLATFORMS }}
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v2
+        with:
+          platforms: ${{ env.PLATFORMS }}
+      - name: Login to Docker Hub
+        uses: docker/login-action@v2
+        with:
+          username: "${{ secrets.CONTAINER_USER }}"
+          password: "${{ secrets.CONTAINER_SECRET }}"
```
matt-condon, line 45:
```suggestion
          username: "${{ secrets.DOCKERHUB_BOT_USER }}"
          password: "${{ secrets.DOCKERHUB_BOT_SECRET }}"
```

matt-condon (dismissed):
Nice work! LGTM
Do we have plans to document this release process?

nicohrubec:
@matt-condon we should probably do that, what about [here](https://github.com/mongodb/mongodb-atlas-cli/blob/master/RELEASING.md)?

.github/workflows/docker-release.yml
```diff
@@ -0,0 +1,52 @@
+name: Daily build
+on:
+  schedule:
+    - cron: "0 1 * * *" # Every day at 1:00 AM
+  workflow_dispatch: {}
```
wtrocki, line 5:
```suggestion
  workflow_dispatch: 
```

.github/workflows/docker-release.yml
```diff
@@ -0,0 +1,52 @@
...
+    name: Build and publish docker image
+    runs-on: ubuntu-latest
+    env:
+      IMAGE_REPOSITORY: mongodb/atlas
+      PLATFORMS: linux/amd64,linux/arm64
+    steps:
+      - name: Check out code
+        uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
```
wtrocki, line 17:
```suggestion
```

.github/workflows/docker-release.yml
```diff
@@ -0,0 +1,52 @@
...
+        run: |
+          DATE=$(date +'%Y-%m-%d')
+          echo DATE=${DATE} >> $GITHUB_ENV
+      - name: 'Get latest tag'
+        id: get-latest-tag
+        run: echo "::set-output name=tag::$(git tag --list --sort=-version:refname "atlascli*" | head -n 1)"
+      - name: Extract version
+        run: |
+          release_tag=${{ steps.get-latest-tag.outputs.tag }}
+          echo "LATEST_VERSION=${release_tag#*/}" >> $GITHUB_ENV
```
wtrocki, line 29: It is generally good pattern to keep steps self contained and visible. 
For bash scripts setting env vars we likely to benefit from shellcheck.

I advice we have those in scripts like here: https://github.com/mongodb/atlas-sdk-go/blob/main/.github/workflows/autorelease.yaml#L19

It will help us with:
- Maintenance
- Audit
- Self documentation of the release process.
- Testing outside github actions environment

.github/workflows/docker-release.yml
nicohrubec: How is this different? It seems to be the same

wtrocki: > Set fetch-depth: 0 to fetch all history for all branches and tags

We only do that when we need whole repo history for very narrow use cases. It adds a lot of time to the job execution.

.github/workflows/docker-release.yml
```diff
@@ -0,0 +1,52 @@
...
+      - name: Build and push container to Dockerhub
+        uses: docker/build-push-action@v4
+        with:
+          context: .
+          platforms: ${{ env.PLATFORMS }}
+          tags: ${{ env.IMAGE_REPOSITORY }}:latest ,
+            ${{ env.IMAGE_REPOSITORY }}:${{ env.LATEST_VERSION }} ,
+            ${{ env.IMAGE_REPOSITORY }}:${{ env.LATEST_VERSION }}-${{ env.DATE }}
+          file: Dockerfile
+          push: true
```
wtrocki, line 52:
```suggestion
          push: false
```

I would recommend testing job without pushing. If everything works and we are happy then we can have push enabled.

.github/workflows/docker-release.yml
```diff
@@ -0,0 +1,52 @@
...
+        with:
+          fetch-depth: 0
+      - name: Set date
+        id: set-date
+        run: |
+          DATE=$(date +'%Y-%m-%d')
+          echo DATE=${DATE} >> $GITHUB_ENV
+      - name: 'Get latest tag'
+        id: get-latest-tag
+        run: echo "::set-output name=tag::$(git tag --list --sort=-version:refname "atlascli*" | head -n 1)"
```
wtrocki, line 25: If RC versions are going to happen they also going to be picked up. We do have only 2 rc versions so risks are low? Are we ok to accept that? Maybe we can use github releases action to fetch latest release?

https://github.com/bruceadams/get-release

This will also ensure that tags without releases (drafts) aren't covered.

.github/workflows/docker-release.yml
nicohrubec: This is needed to fetch the tags (tested it on my fork and it breaks without). In the latest run this step took 9 seconds so I think nothing to worry about in terms of execution time

.github/workflows/docker-release.yml
nicohrubec: Updated using this action now: https://github.com/marketplace/actions/get-latest-release
This was my initial plan but was developing on my fork where I did not have the releases. I agree this should be more reliable.

.github/workflows/docker-release.yml
wtrocki: See my other comments. 
You are looking for valid releases with binaries attached. Looking for tags might catch us in the middle of the release etc.

wtrocki: Since new approach is used we no longer need it?

.github/workflows/docker-release.yml
nicohrubec: Noted and might look into this in a followup

.github/workflows/docker-release.yml
```diff
@@ -0,0 +1,52 @@
...
+        with:
+          username: "${{ secrets.DOCKERHUB_USER }}"
+          password: "${{ secrets.DOCKERHUB_SECRET }}"
+      - name: Build and push container to Dockerhub
+        uses: docker/build-push-action@v4
+        with:
+          context: .
+          platforms: ${{ env.PLATFORMS }}
+          tags: ${{ env.IMAGE_REPOSITORY }}:latest ,
+            ${{ env.IMAGE_REPOSITORY }}:${{ env.LATEST_VERSION }} ,
```
wtrocki, line 49: Less documentation and less confusion to the users.
```suggestion
```

wtrocki (approved):
LGTM. I think we should only push 2 tags per each run (latest and version)

.github/workflows/docker-release.yml
```diff
@@ -0,0 +1,52 @@
...
+    env:
+      IMAGE_REPOSITORY: mongodb/atlas
+      PLATFORMS: linux/amd64,linux/arm64
+    steps:
+      - name: Check out code
+        uses: actions/checkout@v3
+      - name: Set date
+        id: set-date
+        run: |
+          DATE=$(date +'%Y-%m-%d')
```
wtrocki, line 19: IMHO using timestamp would be safer unless we intentionally wanty  to override images daily?
```suggestion
          DATE=$(date +'%Y-%m-%d')
```

PR #2085, opened 2023-07-13
CLOUDP-164587: Upgrade Atlas CLI to go1.20

gssbzn:
## Proposed changes

_Jira ticket:_ CLOUDP-164587

## Further comments

Update go and enable double error wrapping where possible

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -167,7 +167,6 @@ linters:
   # - wsl # Whitespace Linter - Forces you to use empty lines! [fast: true, auto-fix: false]
 
 run:
-  go: '1.19'
   tests: true
   timeout: 10m
   build-tags:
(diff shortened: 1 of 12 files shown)


andreaangiolillo (commented):
Should we update the go version in the GitHub workflows?

gssbzn:
> Should we update the go version in the GitHub workflows?

no need, they should be picking up from the go.mod file

https://github.com/mongodb/mongodb-atlas-cli/blob/7ecd228a7a049e058f19860f8bbf5da727d83504/.github/workflows/code-health.yml#L39

andreaangiolillo (approved):
LGTM as the workflow seems to use the version in the go.mod

tibulca (approved):

PR #2101, opened 2023-07-21
CLOUDP-191364: [Atlas CLI] Unable to create the API key accessList entry

gssbzn:
Fixing description to make it clearer is the ID and not the public key


sarahsimpers (approved):
LGTM

PR #2120, opened 2023-08-02
fix: dependabot limit for PRs

wtrocki:
New PRs with updates are never created because of the limits we currently have on dependabot for active PRs. This PR does changes the limit as a short countermeasure.  In the long run, we should consider tweaking our config or check for more powerful tools like renovate 

Screenshot 2023-08-02 at 19 58 07

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -5,6 +5,7 @@ updates:
     schedule:
       interval: weekly
       day: tuesday
+    open-pull-requests-limit: 15
     reviewers:
       - "mongodb/mongocli"
   - package-ecosystem: github-actions
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -5,7 +5,7 @@ updates:
     schedule:
       interval: weekly
       day: tuesday
-    open-pull-requests-limit: 15
+    open-pull-requests-limit: 10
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -5,6 +5,7 @@ updates:
     schedule:
       interval: weekly
       day: tuesday
+    open-pull-requests-limit: 15
```
wtrocki, line 8: 15 is magical number for us to try - it doesn't overload PR list as default list limit is 20 and we can try how it works.

gssbzn: let's try 10, we are hitting the limit because of the k8s stuff which is going unreviewed so we need to work that with the kube team

wtrocki: Agree!

wtrocki: I also saw Kubernetes being issue. I propose we do group some PRs in the future. 
Ths draft PR contains some ideas: https://github.com/mongodb/mongodb-atlas-cli/pull/2124

wtrocki, line 8:
```suggestion
    open-pull-requests-limit: 10
```

gssbzn (approved):

PR #2124, opened 2023-08-02
fix: introduce PR groups for dependabot

wtrocki:
Just an idea to help us with more frequent and efficient PR reviews.
NOTE: Dependabot is quite limited in that regard. Renovate is much more powerful and also free.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -7,6 +7,25 @@ updates:
       day: tuesday
     reviewers:
       - "mongodb/mongocli"
+    groups:
+      internal-dependencies:
+          patterns:
+            - "go.mongodb.org*" # A single dependency name
+      golang:
+          patterns:
+            - "golang.org*" 
+      kubernetes:
+          patterns:
+            - "*k8s.io*"
+      google:
+          patterns:
+            - "*google.golang.org*"
+      aws:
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -7,6 +7,25 @@ updates:
       day: tuesday
     reviewers:
       - "mongodb/mongocli"
+    groups:
+      internal-dependencies:
+          patterns:
+            - "go.mongodb.org*" # A single dependency name
```
wtrocki, line 13: Those should have the highest priority and can be labeled differently for help rota actions

wtrocki:
FYI @gssbzn WDYT about trying this out (add remove some groups). 
If that is going to clash with existing owners json and processes  we can close the PR.

.github/dependabot.yml
gssbzn: this will include the go driver which is kind of 🤷 for us

gssbzn (approved):
:shipit: and let's see how it goes

.github/dependabot.yml
wtrocki: Removing section. It will introduce more harm than good.

gssbzn (approved):

PR #2134, opened 2023-08-03
CLOUDP-193000: Add command “atlas backups compliancepolicy copyprotection”

VincentPikand:
## Proposed changes

Added the command `atlas backups compliancepolicy copyprotection` with unit and E2E tests.
Happy flow missing for now, written down and will be addressed in the E2E ticket.
Initially the plan was to add --enable/--disable, but it was programatically (and logically IMO) much simpler to add a positional argument, so the usage would now be `atlas backups compliancepolicy copyprotection enable/disable`.

_Jira ticket:_ CLOUDP-193000

## Checklist


internal/cli/atlas/backup/compliancepolicy/copyprotection.go
```diff
@@ -0,0 +1,117 @@
...
+	return nil
+}
+
+func CopyProtectionBuilder() *cobra.Command {
+	opts := new(CopyProtectionOpts)
+	use := "copyProtection"
+	cmd := &cobra.Command{
+		Use:       use,
+		Aliases:   cli.GenerateAliases(use),
+		Args:      cobra.MatchAll(cobra.OnlyValidArgs, cobra.ExactArgs(1)),
```
blva, line 92:
```suggestion
		Args: require.ExactArgs(1),

```

blva: Using this will give better error:

before
```shell
 ./bin/atlas backups compliancepolicy copyProtection                                         1.20 13:34:11
Error: accepts 1 arg(s), received 0
````

after
```
./bin/atlas backups compliancepolicy copyProtection                                  4s  1.20 13:36:54
Error: "atlas backups compliancePolicy copyProtection" requires 1 argument, received 0

Usage:  atlas backups compliancePolicy copyProtection [flags]
```

(pls make sure to import require as well on top of the file)

File: internal/cli/atlas/backup/compliancepolicy/copyprotection.go [hunk 2]
```diff
@@ -0,0 +1,117 @@
...
+				opts.InitOutput(cmd.OutOrStdout(), ""),
+				opts.PreRun,
+			)
+		},
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return opts.Run()
+		},
+	}
+
+	cmd.Flags().StringVar(&opts.ProjectID, flag.ProjectID, "", usage.ProjectID)
```
blva, line 115: [q] where are the other flags for this command? even if we don't want to add the happy flow, can we please make sure we're adding the planned command? 

```shell
  atlas backups compliancepolicy enable [flags]
Flags:
  --authorizedEmail string    Email address of a security or legal representative
  --projectId string          Hexadecimal string that identifies the project to use. This option overrides the settings in the configuration file or environment variable.
  --force                Indicates whether to skip the confirmation prompt before proceeding with the requested
```

internal/cli/atlas/backup/compliancepolicy/copyprotection.go
```diff
@@ -0,0 +1,117 @@
...
+	opts := new(CopyProtectionOpts)
+	use := "copyProtection"
+	cmd := &cobra.Command{
+		Use:       use,
+		Aliases:   cli.GenerateAliases(use),
+		Args:      cobra.MatchAll(cobra.OnlyValidArgs, cobra.ExactArgs(1)),
+		ValidArgs: []string{enable, disable},
+		Short:     "Enable or disable copyprotection of the backup compliance policy for your project.",
+		PreRunE: func(cmd *cobra.Command, args []string) error {
+			// we throw an error if 1 argument is not present,
```
blva, line 96: nit: not sure we need this comment here as we have several examples doing the same in the code

internal/cli/atlas/backup/compliancepolicy/copyprotection.go
blva: also, curious to know why we decide to cover happy flow in e2e task, can we please create another task for the happy flow as e2e already covers other things?

blva (changes requested):

internal/cli/atlas/backup/compliancepolicy/copyprotection.go
VincentPikand: Thanks! I was unaware of the better error, updated.

internal/cli/atlas/backup/compliancepolicy/copyprotection.go
```diff
@@ -0,0 +1,117 @@
...
+	return nil
+}
+
+func (opts *CopyProtectionOpts) copyProtectionWatcher() (bool, error) {
+	res, err := opts.store.DescribeCompliancePolicy(opts.ConfigProjectID())
+	if err != nil {
+		return false, err
+	}
+	if res.GetState() == "" {
+		return false, fmt.Errorf("could not access State field")
```
gssbzn, line 67: if you don't need string interpolation 
```suggestion
		return false, errors.New("could not access State field")
```

internal/cli/atlas/backup/compliancepolicy/copyprotection.go
```diff
@@ -0,0 +1,117 @@
...
+	cli.WatchOpts
+	policy *atlasv2.DataProtectionSettings
+	store  store.CompliancePolicy
+	enable bool
+}
+
+const (
+	enable  string = "enable"
+	disable string = "disable"
+	active  string = "ACTIVE"
```
gssbzn, line 41: type should redundant on initialization 
```suggestion
	enable  = "enable"
	disable = "disable"
	active  = "ACTIVE"
```

internal/cli/atlas/backup/compliancepolicy/copyprotection.go
VincentPikand: > [q] where are the other flags for this command?


I believe you've accidentally mixed things up: this is the `copyprotection` command, not the `enable` command. Let me know if I understood you correctly. The `copyprotection` command has only 1 flag (`--projectId`), which is implemented.

>also, curious to know why we decide to cover happy flow in e2e task, can we please create another task for the happy flow as e2e already covers other things?

We can't test happy flows because we don't (yet) have any commands to enable BCP. I imagined the E2E task to uncomment one line in the describe test and add this happy flow test. It basically only exists because there was still some discussion around the enable & setup commands while I already started implementing. Given this context, would you still like me to create another task or is it fine to keep them combined?

internal/cli/atlas/backup/compliancepolicy/copyprotection.go
```diff
@@ -0,0 +1,117 @@
...
+	_, err := opts.store.UpdateCompliancePolicy(opts.ConfigProjectID(), opts.policy)
+	if err != nil {
+		return err
+	}
+	if err := opts.Watch(opts.copyProtectionWatcher); err != nil {
+		return err
+	}
+
+	fmt.Printf("Your copy protection has been set to: %v\n", opts.enable)
+	return nil
```
gssbzn, line 83: can you please use our custom printers, this approach won't support `-o json`

gssbzn (changes requested):

internal/cli/atlas/backup/compliancepolicy/copyprotection.go
VincentPikand: Removed

internal/cli/atlas/backup/compliancepolicy/copyprotection.go
VincentPikand: Could you elaborate a bit more on this? I'm not disagreeing with you, just want to use this as a learning opportunity. My reasoning was that without `string`, the type is `untyped string` (`const enable untyped string = "enable"`). With `string`, it's what you'd expect of course (`const disable string = "disable"
`). Why is this worse?

internal/cli/atlas/backup/compliancepolicy/copyprotection.go
VincentPikand: Please let me know if my updated solution works for you

internal/cli/atlas/backup/compliancepolicy/copyprotection.go
gssbzn: jumping here to mention this is missing output and this still doesn't work with `-o json`, please make sure that runing this command `-o json` will print a json of the reponse

VincentPikand: @gssbzn do we have some kind of policy that every command must support -o (or some other reason that requires the flag)? Asking because in the spec document, it was not planned to have `-o` for this command.

gssbzn: all commands that hit an API should support O, is part of making the CLI friendly for programmatic users that want to script things

VincentPikand: Thanks

internal/cli/atlas/backup/compliancepolicy/copyprotection.go
VincentPikand: Thanks for the clarification, change implemented.

internal/cli/atlas/backup/compliancepolicy/copyprotection.go
VincentPikand: Changed again to allow `-o json`

gssbzn (approved):
LGTM thanks for all the work here

blva (approved):

PR #2144, opened 2023-08-08
CLOUDP-193348: Add command "atlas backups compliancepolicy setup" L1 support

VincentPikand:
## Proposed changes

Add the command `atlas backups compliancepolicy setup`. 
A confirmation prompt will be added after L1 support has been added for all of the operations.

### Command output
```
$ bin/atlas backups compliancepolicy setup -f ./test.json

Your backup compliance policy has been set up with the following configuration:

Project:                          64b034b236930362b0c8d36b
Authorized e-mail:                firstname.lastname@example.com
Copy protection enabled:          false
Encryption at rest enabled:       false
Point-in-Time restores enabled:   false
Restore window days:              0

POLICIES
ID                         FREQUENCY INTERVAL   FREQUENCY TYPE   RETENTION
64b16beffcaed70cfbd5f520   -                    daily            5 days
64ba452852594888bd5b7070   6                    hourly           2 days
64c795d6401a8ccaea00bf85   -                    ondemand         1 days
```
_Jira ticket:_ CLOUDP-193348

## Checklist

### DIFF
--- a/internal/cli/atlas/backup/compliancepolicy/compliancepolicy.go
+++ b/internal/cli/atlas/backup/compliancepolicy/compliancepolicy.go
@@ -35,9 +35,9 @@ func Builder() *cobra.Command {
 	cmd := baseCommand()
 
 	cmd.AddCommand(
+		SetupBuilder(),
 		DescribeBuilder(),
 		CopyProtectionBuilder(),
-		SetupBuilder(),
 	)
 
 	return cmd
--- a/internal/cli/atlas/backup/compliancepolicy/compliancepolicy.go
+++ b/internal/cli/atlas/backup/compliancepolicy/compliancepolicy.go
@@ -37,6 +37,7 @@ func Builder() *cobra.Command {
 	cmd.AddCommand(
 		DescribeBuilder(),
 		CopyProtectionBuilder(),
+		SetupBuilder(),
 	)
 
 	return cmd
--- a/internal/cli/atlas/backup/compliancepolicy/copyprotection.go
+++ b/internal/cli/atlas/backup/compliancepolicy/copyprotection.go
@@ -42,8 +42,7 @@ const (
 	active  = "ACTIVE"
 )
 
-var copyProtectionTemplate = `
-Copy protection has been set to: {{.CopyProtectionEnabled}}
+var copyProtectionTemplate = `Copy protection has been set to: {{.CopyProtectionEnabled}}
 `
 
 func (opts *CopyProtectionOpts) initStore(ctx context.Context) func() error {
--- a/internal/cli/atlas/backup/compliancepolicy/copyprotection.go
+++ b/internal/cli/atlas/backup/compliancepolicy/copyprotection.go
@@ -42,7 +42,9 @@ const (
 	active  = "ACTIVE"
 )
 
-var copyProtectionTemplate = `Copy protection has been set to: {{.CopyProtectionEnabled}}`
+var copyProtectionTemplate = `
+Copy protection has been set to: {{.CopyProtectionEnabled}}
+`
 
 func (opts *CopyProtectionOpts) initStore(ctx context.Context) func() error {
 	return func() error {
--- a/internal/cli/atlas/backup/compliancepolicy/setup.go
+++ b/internal/cli/atlas/backup/compliancepolicy/setup.go
@@ -39,8 +39,7 @@ type SetupOpts struct {
 	confirm bool
 }
 
-var setupTemplate = `
-Your backup compliance policy has been set up with the following configuration:
+var setupTemplate = `Your backup compliance policy has been set up with the following configuration:
 
 Project:	{{.ProjectId}}
 Authorized e-mail:	{{.AuthorizedEmail}}
--- a/internal/cli/atlas/backup/compliancepolicy/setup.go
+++ b/internal/cli/atlas/backup/compliancepolicy/setup.go
@@ -78,9 +78,6 @@ func (opts *SetupOpts) setupWatcher() (bool, error) {
 }
 
 func (opts *SetupOpts) Run() error {
-	if !opts.confirm {
-		return errors.New("this feature is not production ready: to continue, use '--force'")
-	}
 	opts.policy.SetProjectId(opts.ConfigProjectID())
 
 	_, err := opts.store.UpdateCompliancePolicy(opts.ConfigProjectID(), opts.policy)
@@ -124,6 +121,7 @@ func SetupBuilder() *cobra.Command {
 	_ = cmd.RegisterFlagCompletionFunc(flag.Output, opts.AutoCompleteOutputFlag())
 	cmd.Flags().StringVarP(&opts.path, flag.File, flag.FileShort, "", usage.BackupCompliancePolicyFile)
 	cmd.Flags().BoolVar(&opts.confirm, flag.Force, false, usage.Force)
+	_ = cmd.Flags().MarkHidden(flag.Force)
 	_ = cmd.MarkFlagRequired(flag.File)
 	return cmd
 }
--- a/internal/cli/atlas/backup/compliancepolicy/setup.go
+++ b/internal/cli/atlas/backup/compliancepolicy/setup.go
@@ -124,6 +124,6 @@ func SetupBuilder() *cobra.Command {
 	_ = cmd.RegisterFlagCompletionFunc(flag.Output, opts.AutoCompleteOutputFlag())
 	cmd.Flags().StringVarP(&opts.path, flag.File, flag.FileShort, "", usage.BackupCompliancePolicyFile)
 	cmd.Flags().BoolVar(&opts.confirm, flag.Force, false, usage.Force)
-	cmd.MarkFlagRequired(flag.File)
+	_ = cmd.MarkFlagRequired(flag.File)
 	return cmd
 }
--- a/internal/cli/atlas/backup/compliancepolicy/setup.go
+++ b/internal/cli/atlas/backup/compliancepolicy/setup.go
@@ -32,14 +32,32 @@ import (
 type SetupOpts struct {
 	cli.GlobalOpts
 	cli.WatchOpts
-	policy  *atlasv2.DataProtectionSettings
-	store   store.CompliancePolicy
-	fs      afero.Fs
-	path    string
-	confirm bool
+	policy      *atlasv2.DataProtectionSettings
+	store       store.CompliancePolicy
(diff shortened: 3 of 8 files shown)


internal/cli/atlas/backup/compliancepolicy/setup.go
```diff
@@ -0,0 +1,129 @@
...
+	if res.GetState() == "" {
+		return false, errors.New("could not access State field")
+	}
+	return (res.GetState() == active), nil
+}
+
+func (opts *SetupOpts) Run() error {
+	if !opts.confirm {
+		return errors.New("this feature is not production ready: to continue, use '--force'")
+	}
```
VincentPikand, line 83: Right now we throw this error until the interactive confirmation flow is implemented. I am open to a better error, but too much effort shouldn't be wasted – it'll be removed in no time.

internal/cli/atlas/backup/compliancepolicy/compliancepolicy.go
```diff
@@ -37,6 +37,7 @@ func Builder() *cobra.Command {
 	cmd.AddCommand(
 		DescribeBuilder(),
 		CopyProtectionBuilder(),
+		SetupBuilder(),
```
gssbzn, line 40: can you make this the first command? they print in the order you define them here when doing help

internal/cli/atlas/backup/compliancepolicy/setup.go
gssbzn: don't add the flag would be a better way to handle this ;)

gssbzn:
can you please include in the description the output of the command when you run it

VincentPikand:
> can you please include in the description the output of the command when you run it

Added to the description of the PR.

internal/cli/atlas/backup/compliancepolicy/compliancepolicy.go
VincentPikand: Done

internal/cli/atlas/backup/compliancepolicy/setup.go
VincentPikand: Agreed. Will remove.

gssbzn:
the output in the description is for `atlas backups compliancepolicy policies describe` and not the set up command

VincentPikand:
> the output in the description is for `atlas backups compliancepolicy policies describe` and not the set up command

Sorry about that, updated.

internal/cli/atlas/backup/compliancepolicy/copyprotection.go
```diff
@@ -42,7 +42,9 @@ const (
 	active  = "ACTIVE"
 )
 
-var copyProtectionTemplate = `Copy protection has been set to: {{.CopyProtectionEnabled}}`
+var copyProtectionTemplate = `
```
gssbzn, line 45: any reason for the extra new line at the start?

internal/cli/atlas/backup/compliancepolicy/setup.go
```diff
@@ -0,0 +1,127 @@
...
+	cli.WatchOpts
+	policy  *atlasv2.DataProtectionSettings
+	store   store.CompliancePolicy
+	fs      afero.Fs
+	path    string
+	confirm bool
+}
+
+var setupTemplate = `
+Your backup compliance policy has been set up with the following configuration:
```
gssbzn, line 43: same as above, any reason we are now adding new lines at the start?

internal/cli/atlas/backup/compliancepolicy/setup.go
```diff
@@ -0,0 +1,127 @@
...
+		return false, err
+	}
+	if res.GetState() == "" {
+		return false, errors.New("could not access State field")
+	}
+	return (res.GetState() == active), nil
+}
+
+func (opts *SetupOpts) Run() error {
+	opts.policy.SetProjectId(opts.ConfigProjectID())
```
gssbzn, line 81: [q] is this necessary? it should be given by the path?
if yes I think youc an encapsulate this logic in the store

internal/cli/atlas/backup/compliancepolicy/setup.go
```diff
@@ -0,0 +1,127 @@
...
+				return err
+			}
+			return opts.Run()
+		},
+	}
+
+	cmd.Flags().StringVar(&opts.ProjectID, flag.ProjectID, "", usage.ProjectID)
+	cmd.Flags().StringVarP(&opts.Output, flag.Output, flag.OutputShort, "", usage.FormatOut)
+	_ = cmd.RegisterFlagCompletionFunc(flag.Output, opts.AutoCompleteOutputFlag())
+	cmd.Flags().StringVarP(&opts.path, flag.File, flag.FileShort, "", usage.BackupCompliancePolicyFile)
```
gssbzn, line 122:
```suggestion
	cmd.Flags().StringVarP(&opts.path, flag.File, flag.FileShort, "", usage.BackupCompliancePolicyFile)
	cmd.Flags().BoolVarP(&opts.EnableWatch, flag.EnableWatch, flag.EnableWatchShort, false, usage.EnableWatch)
```

internal/usage/usage.go
```diff
@@ -439,6 +439,7 @@ dbName and collection are required only for built-in roles.`
 	WatchTimeout                              = "Time in seconds until a watch times out. After a watch times out, the CLI no longer watches the command."
 	CompactResponse                           = "Flag that enables the compact array response structure for a json output. The --compact option returns array objects as top-level responses and allows backward compatibility for scripts based on previous CLI versions. Omitting the --compact option for a json output returns array objects within a 'results' sub-array. You must specify --output json to use this option."
 	DataFederationFile                        = "Path to an optional JSON configuration file that defines data federation settings."
+	BackupCompliancePolicyFile                = "Path to an optional JSON configuration file that defines backup compliance policy settings."
```
gssbzn, line 442: I belive this file is not optional

internal/cli/atlas/backup/compliancepolicy/setup.go
VincentPikand: What should we print if the user does not use `--watch`? I think the current output message would be misleading since we can't say for sure that it's been set up, right?

internal/cli/atlas/backup/compliancepolicy/setup.go
VincentPikand: I think there is a decision to be made. To we prioritise the projectId set in the json file, or the one in the config? My initial thought was the one in the config because then we are consistent with how the rest of the `compliancepolicy` commands use `projectId`.

internal/cli/atlas/backup/compliancepolicy/setup.go
VincentPikand: removed

internal/cli/atlas/backup/compliancepolicy/copyprotection.go
VincentPikand: removed

internal/cli/atlas/backup/compliancepolicy/setup.go
gssbzn: This goes beyond a decision we could make, for example what does the API do whit this field, is it possible that in reality this field is readonly?, what happens when you make a request and this field is not set, what happens if you set it to a different ID from the one in the path?

internal/cli/atlas/backup/compliancepolicy/setup.go
gssbzn: locking the terminal should always be opt-in, the question then becomes, what's the default state, `ENABLING`?
when watch is true and it finishes (what happens if I ctrl+c in the middle?) 
then you have
>Your backup compliance policy has been set up with the following configuration

Default should be 
>Enabling your backup compliance policy with the following configuration

internal/cli/atlas/backup/compliancepolicy/setup.go
VincentPikand: >what happens when you make a request and this field is not set

We get this response:
```json
"detail": "Invalid Backup Compliance Policy settings: [projectId] cannot be null.",
    "error": 400,
    "errorCode": "BACKUP_COMPLIANCE_POLICY_SETTINGS_INVALID",
    "parameters": [
        "[projectId] cannot be null"
    ],
    "reason": "Bad Request"
```

>what happens if you set it to a different ID from the one in the path?

It seems to me that it's ignored (unless it's in the incorrect format (e.g. 3 characters long)). I guess this means that it is read-only? 

In the context of the CLI, we just pass a `projectId` and these json/path details are hidden from us. Which takes me back to my original question:
Let's say the user passes a `--projectId` _and_ there is a `projectId` field in the json that they provided. Which one takes preference? What if `--projectId` is not present?

internal/cli/atlas/backup/compliancepolicy/setup.go
VincentPikand: >what's the default state, `ENABLING`?

Yes.

I'll change it to what you proposed.

internal/cli/atlas/backup/compliancepolicy/setup.go
gssbzn: Let's start with the easy one
>What if --projectId is not present?

`opts.ConfigProjectID()` is layered, it looks at `--projectId`, the env `MONGODB_ATLAS_PROJECT_ID`, or the config file value for `project_id` of the selected profile

now a harder one
>Let's say the user passes a --projectId and there is a projectId field in the json that they provided.

the request is 
```sh
curl https://cloud.mongodb.com/api/atlas/v2/groups/<projectId>/backupCompliancePolicy --data "{ "projectId": <projectId>, ...(other fields omitted for simplicity) }
```

you have 2 project Ids one on the url and one on the body, this is my original question
>[q] is this necessary? it should be given by the path?
if yes I think you can encapsulate this logic in the store

What I would've liked to confirm is what happens if the ID on the path is different from the ID on the body, I did this work for you, it is validated and has to be the same 

Which yes, let's encapsulate the logic on the project ID to the store, the value on the body is not meaningful only the one ont he path matters specially as int he cli you can only provide the path id via
>`--projectId`, the env `MONGODB_ATLAS_PROJECT_ID`, or the config file value for `project_id`

VincentPikand: Fixed

gssbzn (dismissed):
💯 thanks for all the work here

gssbzn (approved):

PR #2169, opened 2023-08-18
CLOUDP-195922: Refactor compliance policy E2E tests

VincentPikand:
## Proposed changes

I refactored BCP tests to be in multiple files.
While currently they make up around ~400 lines (which is already quite a lot), with the upcoming interactive modes, it's going to be a lot more. Breaking them up into makes the tests comprehendible and maintainable.

### Important
**The main change here is the file structure, code is 98% unchanged. Details of the tests should be ignored as they are unchanged**

_Jira ticket:_ CLOUDP-195922

## Checklist


test/e2e/atlas/compliancepolicy_copyprotection_test.go
```diff
@@ -0,0 +1,85 @@
...
+			"-o=json",
+			"--projectId",
+			g.projectID,
+			"--watch",
+		)
+		cmd.Env = os.Environ()
+		resp, outputErr := cmd.CombinedOutput()
+		r.NoError(outputErr, string(resp))
+
+		trimmedResponse := removeDotsFromWatching(resp)
```
VincentPikand, line 53: new logic

test/e2e/atlas/compliancepolicy_setup_test.go
```diff
@@ -0,0 +1,82 @@
...
+			"--force",
+			"--file",
+			path,
+			"--watch",
+		)
+
+		cmd.Env = os.Environ()
+		resp, outputErr := cmd.CombinedOutput()
+
+		trimmedResponse := removeDotsFromWatching(resp)
```
VincentPikand, line 72: new logic

test/e2e/atlas/compliancepolicy_test.go
```diff
@@ -0,0 +1,46 @@
...
+)
+
+func TestCompliancePolicy(t *testing.T) {
+	g := newAtlasE2ETestGeneratorWithBackup(t)
+	g.generateProject("compliancePolicy")
+
+	testCompliancePolicySetup(t, g)
+	testDescribe(t, g)
+	testPoliciesDescribe(t, g)
+	testCopyProtection(t, g)
```
VincentPikand, line 35: new logic

test/e2e/atlas/compliancepolicy_test.go
```diff
@@ -0,0 +1,46 @@
...
+
+// For tests that update BCP, we must --watch to avoid HTTP 400 Bad Request "CANNOT_UPDATE_BACKUP_COMPLIANCE_POLICY_SETTINGS_WITH_PENDING_ACTION".
+// Because we watch the command and this is a testing environment,
+// the resp output has some dots in the beginning (depending on how long it took to finish) that need to be removed.
+// It looks something like this:
+//
+// "...{"projectId": "string", ...}".
+func removeDotsFromWatching(consoleOutput []byte) []byte {
+	return []byte(strings.TrimLeft(string(consoleOutput), "."))
+}
```
VincentPikand, line 46: new logic

test/e2e/atlas/helper_test.go
```diff
@@ -770,3 +770,31 @@ func compareStingsWithHiddenPart(expectedSting, actualString string, specialChar
...
+		deleteFile(t, path)
+	})
+}
+
+func deleteFile(t *testing.T, path string) {
+	t.Helper()
+	if err := os.Remove(path); err != nil {
+		t.Errorf("Error deleting file: %v", err)
+	}
+}
```
VincentPikand, line 800: new logic

gssbzn, line 800: [nit] this can be in-lined, there's little advantage to having it as a dedicated function specially as it's currently not reused

test/e2e/atlas/compliancepolicy_test.go
```diff
@@ -0,0 +1,46 @@
...
+	testCompliancePolicySetup(t, g)
+	testDescribe(t, g)
+	testPoliciesDescribe(t, g)
+	testCopyProtection(t, g)
+}
+
+// For tests that update BCP, we must --watch to avoid HTTP 400 Bad Request "CANNOT_UPDATE_BACKUP_COMPLIANCE_POLICY_SETTINGS_WITH_PENDING_ACTION".
+// Because we watch the command and this is a testing environment,
+// the resp output has some dots in the beginning (depending on how long it took to finish) that need to be removed.
+// It looks something like this:
```
gssbzn, line 41: these seem like docs for the test not for the function, https://go.dev/doc/comment

test/e2e/atlas/compliancepolicy_test.go
```diff
@@ -0,0 +1,46 @@
...
+//
+// "...{"projectId": "string", ...}".
```
gssbzn, line 43: >"...{"projectId": "string", ...}

this sort of things are better to document with tests

test/e2e/atlas/compliancepolicy_setup_test.go
```diff
@@ -0,0 +1,82 @@
...
+		FrequencyType:     "daily",
+		RetentionUnit:     "days",
+		RetentionValue:    1,
+	}
+
+	email := authorizedEmail
+
+	policy := &atlasv2.DataProtectionSettings{
+		ScheduledPolicyItems: []atlasv2.DiskBackupApiPolicyItem{scheduledPolicyItem},
+		ProjectId:            &g.projectID,
```
gssbzn, line 48: you are passing `atlasE2ETestGenerator` when all you needed is the project ID

test/e2e/atlas/compliancepolicy_policies_describe_test.go
```diff
@@ -0,0 +1,57 @@
...
+	r.NoError(err)
+
+	t.Run("policies describe", func(t *testing.T) {
+		cmd := exec.Command(cliPath,
+			backupsEntity,
+			compliancepolicyEntity,
+			policiesEntity,
+			"describe",
+			"--projectId",
+			g.projectID,
```
gssbzn, line 44: you are passing `atlasE2ETestGenerator` when all you needed was the project ID

test/e2e/atlas/compliancepolicy_enable_test.go
```diff
@@ -0,0 +1,60 @@
...
+	"os/exec"
+	"testing"
+
+	"github.com/mongodb/mongodb-atlas-cli/test/e2e"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	atlasv2 "go.mongodb.org/atlas-sdk/v20230201004/admin"
+)
+
+func TestCompliancePolicy_enable(t *testing.T) {
```
gssbzn, line 31:
```suggestion
func TestEnableCompliancePolicy(t *testing.T) {
```

test/e2e/atlas/compliancepolicy_enable_test.go
```diff
@@ -0,0 +1,60 @@
...
+		cmd.Env = os.Environ()
+		resp, outputErr := cmd.CombinedOutput()
+		r.NoError(outputErr, string(resp))
+		var result atlasv2.DataProtectionSettings
+		require.NoError(t, json.Unmarshal(resp, &result), string(resp))
+
+		a := assert.New(t)
+
+		a.Equal(result.GetAuthorizedEmail(), authorizedEmail)
+	})
```
gssbzn, line 59: if there' won't be a sad flow consider removing the nesting, nesting is for scenarios

gssbzn:
>golangci-lint was also not happy about this, thinking the files were helper functions and requesting a t.helper() at the beginning of the functions

But they are helper functions and not actual tests

gssbzn:
>Now, to keep costs and test running times low

What costs? this doesn't seem to require a live cluster so what costs are we talking here about?

test/e2e/atlas/compliancepolicy_copyprotection_test.go
```diff
@@ -0,0 +1,85 @@
...
+		cmd := exec.Command(
+			cliPath,
+			backupsEntity,
+			compliancepolicyEntity,
+			"copyprotection",
+			invalidArgument)
+		cmd.Env = os.Environ()
+		_, err = cmd.CombinedOutput()
+		r.Error(err)
+	})
```
gssbzn, line 73: what do we gain with this test? this is the kind of test that would be better fitted for a unit tests. Also to keep in mind this command should be refactored as previously discussed to be two sub commands instead of arguments

test/e2e/atlas/compliancepolicy_copyprotection_test.go
```diff
@@ -0,0 +1,85 @@
...
+	t.Run("copyprotection happy flow", func(t *testing.T) {
+		cmd := exec.Command(
+			cliPath,
+			backupsEntity,
+			compliancepolicyEntity,
+			"copyprotection",
+			"enable",
+			"-o=json",
+			"--projectId",
+			g.projectID,
```
gssbzn, line 46: you are passing `atlasE2ETestGenerator` when all you need is the project ID

test/e2e/atlas/compliancepolicy_setup_test.go
```diff
@@ -0,0 +1,82 @@
...
+		resp, outputErr := cmd.CombinedOutput()
+
+		trimmedResponse := removeDotsFromWatching(resp)
+
+		r.NoError(outputErr, string(resp))
+		a := assert.New(t)
+
+		var result atlasv2.DataProtectionSettings
+		require.NoError(t, json.Unmarshal(trimmedResponse, &result), trimmedResponse)
+		a.Equal(1, len(result.GetScheduledPolicyItems()))
```
gssbzn, line 79:
```suggestion
		a.Len(result.GetScheduledPolicyItems(), 1)
```

test/e2e/atlas/compliancepolicy_test.go
```diff
@@ -0,0 +1,46 @@
...
+)
+
+func TestCompliancePolicy(t *testing.T) {
+	g := newAtlasE2ETestGenerator(t)
+	g.generateProject("compliancePolicy")
+
+	testCompliancePolicySetup(t, g)
+	testDescribe(t, g)
+	testPoliciesDescribe(t, g)
+	testCopyProtection(t, g)
```
gssbzn, line 35: what's the relationship between these tests, do they need to run in this order?

gssbzn (changes requested):
let's talk about test order dependency and costs more as this approach as you have already call out is hacky and may cost us on maintainability later on

test/e2e/atlas/helper_test.go
VincentPikand: removed

test/e2e/atlas/compliancepolicy_test.go
VincentPikand: Greatly simplified the comment, let me know if the new version works for you

test/e2e/atlas/compliancepolicy_test.go
VincentPikand: rewrote the comment

test/e2e/atlas/compliancepolicy_policies_describe_test.go
VincentPikand: fixed

test/e2e/atlas/compliancepolicy_enable_test.go
VincentPikand: I like the nesting because of the comment it outputs to the console and a sense of "progression" in the tests.
`=== RUN   TestEnableCompliancePolicy/enable_happy_flow`
Yes, could be renamed to just `happy_flow`, but the point still stands.
Additionally, it's kind of a title for a test – if somebody else comes to read this piece of code, they will immediately understand what the purpose is. (Answering the question "how" with "happy flow" and "what" with "TestEnableCompliancePolicy".

What do you think?

test/e2e/atlas/compliancepolicy_copyprotection_test.go
VincentPikand: fixed

test/e2e/atlas/compliancepolicy_setup_test.go
VincentPikand: fixed

test/e2e/atlas/compliancepolicy_test.go
VincentPikand: Setup has to be first, others are independent.

test/e2e/atlas/compliancepolicy_copyprotection_test.go
VincentPikand: My intention was to implement a sad flow test, but I agree, this is very "unity'. Do you have any pointers/suggestions/guidelines as to what kind of E2E sad flow tests to implement generally?
For this particular test, I will remove it when I refactor this command and test the behaviour with unit tests instead.

VincentPikand:
>But they are helper functions and not actual tests

take for example `compliancepolicy_setup_test.go`. It has actual tests in it, why would you consider it a helper function? The downside of marking it as a helper function is (from `t.Helper()` description): `When printing file and line information, that function will be skipped.`.

test/e2e/atlas/compliancepolicy_test.go
gssbzn: > others are independent.

why do we need to run them in the same test?

VincentPikand:
>let's talk about test order dependency and costs more as this approach as you have already call out is hacky and may cost us on maintainability later on

> What costs? this doesn't seem to require a live cluster so what costs are we talking here about?

This is an interesting discussion from my perspective and I'm happy to discuss it. 

First, I'll address the "costs". I mean't setting up (and tearing down) a project, It's a call to the API and CPU compute on the MMS. Now, if you say that it is negligible, we can disregard this, but I'd be curious to hear what is the threshold for you (how to approach this question generally).

Secondly, From the time perspective, with my environment and good internet, it took ~3 seconds to run. Of course this can vary quite a lot, but with ~10 tests, 30 seconds is virtually guaranteed. I believe tests running for a long time has several drawbacks so being mindful of it where possible is good, even if it is a small win. A bunch of small wins eventually add up to a big win, right?

Let me know your thoughts on this, I'm curious.

gssbzn:
> It has actual tests in it, why would you consider it a helper function? 

according to go this is a test
```go
func TestMyThing(t *testing.T)
```

this is just a function or helper

```go
func testMyThing(t *testing.T)
```

the content doesn't matter

test/e2e/atlas/compliancepolicy_test.go
VincentPikand: I mean't independent in terms of order.

VincentPikand:
> > It has actual tests in it, why would you consider it a helper function?
> 
> according to go this is a test
> 
> ```go
> func TestMyThing(t *testing.T)
> ```
> 
> this is just a function or helper
> 
> ```go
> func testMyThing(t *testing.T)
> ```
> 
> the content doesn't matter

Yes, but if the function name starts with `Test...`, `go` want's to run it independently, but we actually want to call it from the parent test. This is why I consider the test setup hacky.

gssbzn:
>This is why I consider the test setup hacky.

And this is what expose that your design may be flawed probably under a wrong assumption like costs

See for example https://github.com/mongodb/mongodb-atlas-cli/blob/a91b145756538c4fb499dc4db71cc4a183ee7b79/test/e2e/atlas/dbusers_test.go#L103-L105

test/e2e/atlas/compliancepolicy_copyprotection_test.go
```diff
@@ -0,0 +1,92 @@
...
+		cmd.Env = os.Environ()
+		resp, outputErr := cmd.CombinedOutput()
+		r.NoError(outputErr, string(resp))
+
+		trimmedResponse := removeDotsFromWatching(resp)
+
+		a := assert.New(t)
+
+		var compliancepolicy atlasv2.DataProtectionSettings
+		require.NoError(t, json.Unmarshal(trimmedResponse, &compliancepolicy), string(trimmedResponse))
```
gssbzn, line 65: can you please decide if you are using `r.NoError` or `require.NoError` and unify to one approach, in thoery the recomemndation from the library is if you have multiple uses for `require.<Assertion>` you can do `r.NoError(err)` and use `r.<Assertion>` to simplify but this has no benefit except from saving you from having to pass `t` 

This is also to flag for example your use of `a := assert.New(t)` here, you only have one use of `a.True(*compliancepolicy.CopyProtectionEnabled)` that coud save you the extra line by doing `assert.True(t, *compliancepolicy.CopyProtectionEnabled)`

test/e2e/atlas/compliancepolicy_copyprotection_test.go
gssbzn: >pointers/suggestions/guidelines

E2E are slow, ideally you want to follow the [test pyramid](https://martinfowler.com/articles/practical-test-pyramid.html), you e2e will usually be happy path or mission critical

This test in particular is testing the framework (cobra) to make sure the framework work as expected, we shouldn't test the framework, even less when the option to pass "any" argument is removed

test/e2e/atlas/compliancepolicy_copyprotection_test.go
```diff
@@ -0,0 +1,92 @@
...
+	"os/exec"
+	"testing"
+
+	"github.com/mongodb/mongodb-atlas-cli/test/e2e"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	atlasv2 "go.mongodb.org/atlas-sdk/v20230201004/admin"
+)
+
+func TestCopyProtection(t *testing.T) {
```
gssbzn, line 32:
```suggestion
func TestBackupComplicancePolicyCopyProtection(t *testing.T) {
```

test/e2e/atlas/compliancepolicy_describe_test.go
gssbzn, line 1: test/e2e/atlas/backup_compliancepolicy_describe_test.go

test/e2e/atlas/compliancepolicy_describe_test.go
```diff
@@ -0,0 +1,62 @@
...
+	"os/exec"
+	"testing"
+
+	"github.com/mongodb/mongodb-atlas-cli/test/e2e"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	atlasv2 "go.mongodb.org/atlas-sdk/v20230201004/admin"
+)
+
+func TestDescribe(t *testing.T) {
```
gssbzn, line 32:
```suggestion
func TestBackupCompliancePolicyDescribe(t *testing.T) {
```

test/e2e/atlas/compliancepolicy_describe_test.go
```diff
@@ -0,0 +1,62 @@
...
+		cmd.Env = os.Environ()
+		resp, outputErr := cmd.CombinedOutput()
+
+		r.NoError(outputErr, string(resp))
+
+		a := assert.New(t)
+		var result atlasv2.DataProtectionSettings
+		require.NoError(t, json.Unmarshal(resp, &result), string(resp))
+		a.NotEmpty(result)
+	})
```
gssbzn, line 61: See other comment, happy follow is assumed
```suggestion
		cmd := exec.Command(cliPath,
			backupsEntity,
			compliancepolicyEntity,
			"describe",
			"--projectId",
			g.projectID,
			"-o=json")
		cmd.Env = os.Environ()
		resp, outputErr := cmd.CombinedOutput()

		r.NoError(outputErr, string(resp))

		a := assert.New(t)
		var result atlasv2.DataProtectionSettings
		require.NoError(t, json.Unmarshal(resp, &result), string(resp))
		a.NotEmpty(result)
```

test/e2e/atlas/compliancepolicy_enable_test.go
gssbzn, line 1: test/e2e/atlas/backup_compliancepolicy_enable_test.go

test/e2e/atlas/compliancepolicy_enable_test.go
```diff
@@ -0,0 +1,60 @@
...
+	"os/exec"
+	"testing"
+
+	"github.com/mongodb/mongodb-atlas-cli/test/e2e"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	atlasv2 "go.mongodb.org/atlas-sdk/v20230201004/admin"
+)
+
+func TestEnableCompliancePolicy(t *testing.T) {
```
gssbzn, line 31:
```suggestion
func TestBackupCompliancePolicyEnable(t *testing.T) {
```

test/e2e/atlas/compliancepolicy_enable_test.go
gssbzn, line 1: AAny reason why enable and describe can't be the same test or at least the same file?

test/e2e/atlas/compliancepolicy_policies_describe_test.go
gssbzn, line 1: test/e2e/atlas/backup_compliancepolicy_policies_describe_test.go

test/e2e/atlas/compliancepolicy_policies_describe_test.go
```diff
@@ -0,0 +1,64 @@
...
+	r.NoError(err)
+
+	g := newAtlasE2ETestGenerator(t)
+	g.generateProject("describe-compliance-policy-policies")
+	err = enableCompliancePolicy(g.projectID)
+	if err != nil {
+		t.Fatal(fmt.Errorf("unable to enable compliance policy: %w", err))
+	}
+
+	t.Run("happy flow", func(t *testing.T) {
```
gssbzn, line 44:
```suggestion
```

test/e2e/atlas/compliancepolicy_policies_describe_test.go
```diff
@@ -0,0 +1,64 @@
...
+	cliPath, err := e2e.AtlasCLIBin()
+	r := require.New(t)
+	r.NoError(err)
+
+	g := newAtlasE2ETestGenerator(t)
+	g.generateProject("describe-compliance-policy-policies")
+	err = enableCompliancePolicy(g.projectID)
+	if err != nil {
+		t.Fatal(fmt.Errorf("unable to enable compliance policy: %w", err))
+	}
```
gssbzn, line 42:
```suggestion
	require.NoError(enableCompliancePolicy(g.projectID))
```

test/e2e/atlas/compliancepolicy_setup_test.go
```diff
@@ -0,0 +1,84 @@
...
+	"os/exec"
+	"testing"
+
+	"github.com/mongodb/mongodb-atlas-cli/test/e2e"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	atlasv2 "go.mongodb.org/atlas-sdk/v20230201004/admin"
+)
+
+func TestSetupCompliancePolicy(t *testing.T) {
```
gssbzn, line 31:
```suggestion
func TestBackupCompliancePolicySetup(t *testing.T) {
```

test/e2e/atlas/compliancepolicy_test.go
```diff
@@ -0,0 +1,57 @@
...
+		projectID,
+		"--authorizedEmail",
+		authorizedEmail,
+		"-o=json",
+		"--watch",
+	)
+	cmd.Env = os.Environ()
+	_, outputErr := cmd.CombinedOutput()
+	return outputErr
+}
```
gssbzn, line 57: move these to the helper file

test/e2e/atlas/compliancepolicy_test.go
```diff
@@ -0,0 +1,57 @@
...
+		"--projectId",
+		projectID,
+		"--authorizedEmail",
+		authorizedEmail,
+		"-o=json",
+		"--watch",
+	)
+	cmd.Env = os.Environ()
+	_, outputErr := cmd.CombinedOutput()
+	return outputErr
```
gssbzn, line 56: heads up, if this starts failing is  going to be hard to debug since you swallow the output

test/e2e/atlas/compliancepolicy_copyprotection_test.go
```diff
@@ -0,0 +1,92 @@
...
+	t.Run("copyprotection invalid nr of arguments", func(t *testing.T) {
+		cmd := exec.Command(
+			cliPath,
+			backupsEntity,
+			compliancepolicyEntity,
+			"copyprotection")
+		cmd.Env = os.Environ()
+		_, err = cmd.CombinedOutput()
+		r.Error(err)
+	})
```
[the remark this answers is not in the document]
VincentPikand: Will address unit tests in a separate PR, but removed

test/e2e/atlas/compliancepolicy_test.go
VincentPikand: I'm going with `r` and `a`, since that's the pattern used in the other E2E tests.

test/e2e/atlas/compliancepolicy_copyprotection_test.go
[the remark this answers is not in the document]
VincentPikand: I improved the error message a bit, but not sure if that is what you mean't. Let me know if it works for you.

test/e2e/atlas/compliancepolicy_enable_test.go
VincentPikand: They can, but I thought that for the sake of consistency it should be separated.

test/e2e/atlas/compliancepolicy_copyprotection_test.go
VincentPikand: I'm going with `r` and `a`, since that's the pattern used in the other E2E tests.

test/e2e/atlas/compliancepolicy_test.go
VincentPikand: I wrote a wrapper around it, not sure if that is what you meant, let me know if that works

test/e2e/atlas/compliancepolicy_copyprotection_test.go
gssbzn: keep in mind what I said above, prefer
`assert.True(t)` over `a := assert.New(t); a.True` if there's only one use of `a`, most test follow this

test/e2e/atlas/backup_compliancepolicy_copyprotection_test.go
```diff
@@ -0,0 +1,59 @@
...
+	)
+	cmd.Env = os.Environ()
+	resp, outputErr := cmd.CombinedOutput()
+	r.NoError(outputErr, string(resp))
+
+	var compliancepolicy atlasv2.DataProtectionSettings
+	r.NoError(json.Unmarshal(resp, &compliancepolicy), string(resp))
+
+	a := assert.New(t)
+	a.True(*compliancepolicy.CopyProtectionEnabled)
```
gssbzn, line 58:
```suggestion
	assert.True(t, *compliancepolicy.CopyProtectionEnabled)
```

test/e2e/atlas/helper_test.go
```diff
@@ -770,3 +776,47 @@ func compareStingsWithHiddenPart(expectedSting, actualString string, specialChar
...
+		"--projectId",
+		projectID,
+		"--authorizedEmail",
+		authorizedEmail,
+		"-o=json",
+		"--watch", // avoiding HTTP 400 Bad Request "CANNOT_UPDATE_BACKUP_COMPLIANCE_POLICY_SETTINGS_WITH_PENDING_ACTION".
+	)
+	cmd.Env = os.Environ()
+	_, outputErr := cmd.CombinedOutput()
+	return fmt.Errorf("executing command unsuccessful: %w", outputErr)
```
gssbzn, line 821:
```suggestion
	output, outputErr := cmd.CombinedOutput()
	if outputErr != nil {
		return fmt.Errorf("%w\n %s", outputErr, string(output))
	}
	return nil
```

VincentPikand: Realised this myself 2 minutes ago, had hard tunnel vision

test/e2e/atlas/compliancepolicy_enable_test.go
gssbzn: the pattern we have followed for e2e has usually been `<root_parent>_test.go` with usually `TestRootParentTest()` then all runnable siblings as `t.Run("runnable command", ...)` enable, describe and setup seem like siblings that could follow this pattern, maybe set up has the precondition it can't be enable in which case is another pattern of `TestRootParentSpecialTest` if same file or not would depend on how complex (Lines of code/[Cyclomatic complexity](https://en.wikipedia.org/wiki/Cyclomatic_complexity)) and right now they don't seem that long/complex

PR #2228, opened 2023-09-11
CLOUDP-199262: fix local dev containers stop/start

tibulca:
_Jira ticket:_ CLOUDP-199262


internal/cli/atlas/deployments/setup.go
```diff
@@ -177,15 +177,18 @@ func (opts *SetupOpts) configureMongod(ctx context.Context, keyFileContents stri
...
-				`echo $KEYFILECONTENTS > $KEYFILE && \
-                   chmod 400 $KEYFILE && \
-                   python3 /usr/local/bin/docker-entrypoint.py \
-                                          --transitionToAuth \
-                                          --dbpath $DBPATH \
-                                          --keyFile $KEYFILE \
-                                          --replSet $REPLSETNAME \
-                                          --setParameter "mongotHost=$MONGOTHOST" \
-                                          --setParameter "searchIndexManagementHostAndPort=$MONGOTHOST"`,
+				`if [ ! -f $KEYFILE ]
```
tibulca, line 180: without this check, stopping and starting back mongod will fail (as keyfile is already there)

internal/cli/atlas/deployments/setup.go
```diff
@@ -177,15 +177,18 @@ func (opts *SetupOpts) configureMongod(ctx context.Context, keyFileContents stri
...
+                 then
+                     echo $KEYFILECONTENTS > $KEYFILE
+                     chmod 400 $KEYFILE
+                 fi
+                 python3 /usr/local/bin/docker-entrypoint.py \
```
blva, line 185: [q] do we need to capture the python3 dependency as well somewhere or do all OS come with it by default?

internal/cli/atlas/deployments/setup.go
blva: nice!

blva (approved):

internal/cli/atlas/deployments/setup.go
gssbzn: these file are becoming a bit gnarly, you should consider moving them to dedicates `.sh` files so they are checked with shellcheck and include them via file embedding, see https://github.com/mongodb/mongodb-atlas-cli/blob/fdb8b0cf8b9a87be29095b2d4edfc56fccdd652d/tools/chocolateypkg/chocolateypkg.go#L68C1-L69

internal/cli/atlas/deployments/setup.go
blva: ah nvm me :D

internal/cli/atlas/deployments/setup.go
tibulca: good point, done

blva (approved):

gssbzn (approved):

PR #2245, opened 2023-09-13
task: update unversioned libs

gssbzn:
These are used for evg testing and don't have a stable release so dependabot doesn't pick them up 

Task generation still works as excepted for shrub

Test config whcih uses the simulated terminal still works as expected


wtrocki:
Also another reason why dependabot < renovate :)

PR #2260, opened 2023-09-15
fix: fix deb10 publishing

gssbzn:
Fix deb publishing to publish the correct bin

### DIFF
--- a/tools/genevergreen/generate/publish.go
+++ b/tools/genevergreen/generate/publish.go
@@ -53,7 +53,7 @@ var distros = map[string]Platform{
 		architectures: []string{x86_64, aarch64},
 	},
 	"debian10": {
-		extension:     rpm,
+		extension:     deb,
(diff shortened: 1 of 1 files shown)


tools/genevergreen/generate/publish.go
```diff
@@ -53,7 +53,7 @@ var distros = map[string]Platform{
 		architectures: []string{x86_64, aarch64},
 	},
 	"debian10": {
-		extension:     rpm,
```
gssbzn, line 56: 🙃

andreaangiolillo (approved):
LGTM

tools/genevergreen/generate/publish.go
fmenezes: oh, was that always wrong?

gssbzn: I think so

PR #2273, opened 2023-09-18
CLOUDP-200168: [stability] Update the default podman settings to avoid issues

andreaangiolillo:
## Proposed changes

_Jira ticket:_ CLOUDP-200168

Closes #[issue number]

## Checklist

### DIFF
--- a/internal/podman/client.go
+++ b/internal/podman/client.go
@@ -29,8 +29,10 @@ import (
 )
 
 var (
-	ErrPodmanNotFound  = errors.New("podman not found in your system, check requirements at http://docpage")
-	ErrNetworkNotFound = errors.New("network ip range was not found")
+	ErrPodmanNotFound    = errors.New("podman not found in your system, check requirements at http://docpage")
+	ErrNetworkNotFound   = errors.New("network ip range was not found")
+	defaultMachineCPUs   = "2"
+	defaultMachineMemory = "2048"
 )
 
 type Diagnostic struct {
@@ -201,7 +203,11 @@ func (o *client) machineInit(ctx context.Context) error {
 		return nil
 	}
 
-	_, err = o.runPodman(ctx, "machine", "init")
+	if _, err = o.runPodman(ctx, "machine", "init"); err != nil {
+		return err
+	}
+
+	_, err = o.runPodman(ctx, "machine", "set", "--cpus", defaultMachineCPUs, "--memory", defaultMachineMemory)
 	return err
 }
 


internal/podman/client.go
```diff
@@ -29,8 +29,10 @@ import (
 )
 
 var (
-	ErrPodmanNotFound  = errors.New("podman not found in your system, check requirements at http://docpage")
-	ErrNetworkNotFound = errors.New("network ip range was not found")
+	ErrPodmanNotFound    = errors.New("podman not found in your system, check requirements at http://docpage")
+	ErrNetworkNotFound   = errors.New("network ip range was not found")
+	defaultMachineCPUs   = "2"
+	defaultMachineMemory = "2048"
```
fmenezes, line 35: how do you know if this machine has enough memory or CPU cores?

andreaangiolillo: oh good point, I will add a check. Thanks

internal/podman/client.go
```diff
@@ -201,8 +203,18 @@ func (o *client) machineInit(ctx context.Context) error {
 		return nil
 	}
 
-	_, err = o.runPodman(ctx, "machine", "init")
-	return err
+	if _, err = o.runPodman(ctx, "machine", "init"); err != nil {
```
fmenezes, line 206:
```suggestion
	if _, err = o.runPodman(ctx, "machine", "init", "--cpus", defaultMachineCPUs, "--memory", defaultMachineMemory); err != nil {
```

fmenezes: you don't need to call `podman` again

internal/podman/client.go
andreaangiolillo: updated

internal/podman/client.go
andreaangiolillo: updated

internal/podman/client.go
andreaangiolillo: this actually need a bit more love, I will update you as soon as it is ready for review. Sorry for the ping

internal/podman/client.go
```diff
@@ -201,8 +217,51 @@ func (o *client) machineInit(ctx context.Context) error {
...
+		_, _ = log.Warning(err)
+	}
+
+	if memory.Available > defaultMachineMemoryUint64 {
+		args = append(args, "--memory", defaultMachineMemory)
+	} else {
+		_, _ = log.Warningf(notEnoughMemoryAvailable, memory.Available, defaultMachineMemory)
+	}
+
+	cores, err := cpu.Counts(false)
```
fmenezes, line 240: just for my curiosity what is the difference between this and https://pkg.go.dev/runtime#NumCPU ?

andreaangiolillo: > NumCPU returns the number of logical CPUs usable by the current process.
sure, my understanding is that the CPU usable by the current GO process can be different than the physical system cpus

internal/podman/client.go
andreaangiolillo: updated

internal/podman/client.go
fmenezes: @andreaangiolillo do you mind double checking if you have different results? my understanding is that usable for Go is the same as present on System, it should be all virtual CPUs

internal/podman/client.go
```diff
@@ -201,8 +217,54 @@ func (o *client) machineInit(ctx context.Context) error {
...
+	if cores >= defaultCPUs {
+		args = append(args, "--cpus", defaultMachineCPUs)
+	} else {
+		_, _ = log.Warningf(notEnoughCPUsAvailable, cores, defaultMachineCPUs)
+	}
+
+	return args
+}
+
+func (o *client) machineSet(ctx context.Context) error {
```
fmenezes, line 262: do you need this?

fmenezes: I think you're setting it on machine init

andreaangiolillo: ops you are right. Removing it

internal/podman/client.go
andreaangiolillo: I was not able to find a scenario where `NumCPU` or `cpu.Counts` have different results. Defaulting back to NumCPU

andreaangiolillo: updated

fmenezes (approved):

PR #2296, opened 2023-09-20
CLOUDP-200481: skip podman machine init for linux

tibulca:
## Proposed changes
skip podman machine init when running in linux
tested on Mac and RHEL8

_Jira ticket:_ CLOUDP-200481


colm-quinn (approved):
LGTM!

PR #2299, opened 2023-09-21
CLOUDP-200646: Add network to diagnostics commands

tibulca:
## Proposed changes
Add network to diagnostics commands
_Jira ticket:_ CLOUDP-200646

### DIFF
--- a/internal/podman/client.go
+++ b/internal/podman/client.go
@@ -152,7 +152,7 @@ type Client interface {
 	ListImages(ctx context.Context, nameFilter string) ([]*Image, error)
 	PullImage(ctx context.Context, name string) ([]byte, error)
 	Version(ctx context.Context) (*Version, error)
-	Logs(ctx context.Context) ([]interface{}, error)
+	Logs(ctx context.Context) (map[string]interface{}, []error)
 	ContainerLogs(ctx context.Context, name string) ([]string, error)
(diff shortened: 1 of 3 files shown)


internal/podman/client.go
```diff
@@ -457,20 +464,33 @@ func (o *client) Version(ctx context.Context) (version *Version, err error) {
 	return v, err
 }
 
-func (o *client) Logs(ctx context.Context) ([]interface{}, error) {
-	output, err := o.runPodman(ctx, "ps", "--format", "json")
+func (o *client) Logs(ctx context.Context) (map[string]interface{}, []error) {
+	l := map[string]interface{}{}
+	errs := []error{}
+
+	output, err := o.runPodman(ctx, "ps", "--all", "--format", "json", "--filter", "name=mongo")
```
tibulca, line 471: @blva I added the filter because I think the intention here was to get the conf of MongoDB containers only, or we want the entire conf?

andreaangiolillo (approved):
LGTM

internal/podman/client.go
blva: hello! yes, makes sense, here I added `podman ps` from @fmenezes suggestion, but I may have missed the filter part. FYI@fmenezes just in case you want the whole output for some reason.

blva (approved):

PR #2300, opened 2023-09-21
CLOUDP-200475 Leftover containers when setup/delete fails

fmenezes:
## Proposed changes
 Leftover containers when setup/delete fails

_Jira ticket:_ CLOUDP-200475

## Checklist


internal/cli/atlas/deployments/delete.go
```diff
@@ -42,11 +42,7 @@ func (opts *DeleteOpts) Run(ctx context.Context) error {
 	}
 
 	telemetry.AppendOption(telemetry.WithDeploymentType(options.LocalCluster)) // always local
-	if opts.DeploymentName != "" {
-		if err := opts.DeploymentOpts.CheckIfDeploymentExists(ctx); err != nil {
-			return err
-		}
-	} else {
+	if opts.DeploymentName == "" {
```
fmenezes, line 45: removing the check because we can have left overs

internal/cli/atlas/deployments/options/deployment_opts_connect_string.go
```diff
@@ -17,23 +17,26 @@ package options
...
+func (opts *DeploymentOpts) updateFields(c *define.InspectContainerData) {
 	opts.DeploymentType = deploymentTypeLocal
-	opts.MdbVersion = c.Labels["version"]
-	if len(c.Ports) > 0 {
-		opts.Port = c.Ports[0].HostPort
+	opts.MdbVersion = c.Config.Labels["version"]
+	if len(c.HostConfig.PortBindings) > 0 {
+		for _, port := range c.HostConfig.PortBindings {
+			opts.Port, _ = strconv.Atoi(port[0].HostPort)
+		}
```
tibulca, line 33: what's in c.HostConfig.PortBindings collection for mongo*, do we have multiple entries? (in which case we're overwriting the opts.Port in this for loop).

internal/cli/atlas/deployments/options/deployment_opts_remove.go
```diff
@@ -16,16 +16,27 @@ package options
 import "context"
 
 func (opts *DeploymentOpts) Remove(ctx context.Context) error {
-	if _, err := opts.PodmanClient.RemoveContainers(ctx, opts.LocalMongodHostname(), opts.LocalMongotHostname()); err != nil {
-		return err
+	volumes := []string{opts.LocalMongodDataVolume(), opts.LocalMongotDataVolume(), opts.LocalMongoMetricsVolume()}
```
tibulca, line 19: isn't  `opts.LocalMongodDataVolume()` found by container inspect 2 lines below?
If yes, maybe this collection should include only the last 2 volumes (used by mongot)

internal/cli/atlas/deployments/options/deployment_opts_connect_string.go
fmenezes: in some cases (not ours) we can expose multiple ports for the same container, that is why this struct is in this way

internal/cli/atlas/deployments/options/deployment_opts_remove.go
fmenezes: so what if the container is not present? then we would miss it

internal/cli/atlas/deployments/delete.go
blva: if we find no left overs, shouldn't we still say "deployment not found? "

internal/cli/atlas/deployments/options/deployment_opts_connect_string.go
tibulca: but in this case the code above will set opts.Port with the last HostPort.
not blocking as our container has only one

internal/cli/atlas/deployments/options/deployment_opts_remove.go
tibulca: is podman ok with duplicate names for the delete command (`[non-blocking if podman is ok with duplicate volume names]`)

internal/cli/atlas/deployments/options/deployment_opts_connect_string.go
tibulca: Probably it should look like this:
```suggestion
		for internalPort, port := range c.HostConfig.PortBindings {
			if strings.HasPrefix(internalPort, strconv.Itoa(InternalMongodPort)) {
				opts.Port, _ = strconv.Atoi(port[0].HostPort)
			}
		}
```

fmenezes: I've addressed it, unfortunately reusing the internalMongodPort would require some more coding, so I decided to leave it for a future PR

tibulca (approved):

fmenezes:
failures are from old sha

PR #2324, opened 2023-09-28
CLOUDP-199629: atlas deployments delete --type ATLAS

tibulca:
## Proposed changes
implement atlas deployments delete --type ATLAS

_Jira ticket:_ CLOUDP-199629

### DIFF
--- a/internal/cli/atlas/deployments/delete.go
+++ b/internal/cli/atlas/deployments/delete.go
@@ -16,43 +16,100 @@ package deployments
 
 import (
 	"context"
+	"fmt"
+	"time"
 
 	"github.com/mongodb/mongodb-atlas-cli/internal/cli"
 	"github.com/mongodb/mongodb-atlas-cli/internal/cli/atlas/deployments/options"
 	"github.com/mongodb/mongodb-atlas-cli/internal/cli/require"
+	"github.com/mongodb/mongodb-atlas-cli/internal/config"
 	"github.com/mongodb/mongodb-atlas-cli/internal/flag"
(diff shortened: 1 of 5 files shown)


internal/cli/atlas/deployments/delete.go
```diff
@@ -61,16 +118,52 @@ func (opts *DeleteOpts) Run(ctx context.Context) error {
...
-		Use:     "delete [deploymentName]",
-		Short:   "Delete a deployment.",
+		Use:   "delete [deploymentName]",
+		Short: "Delete a deployment.",
+		Long: `The command prompts you to confirm the operation when you run the command without the --force option. 
+		
+Deleting an Atlas deployment also deletes any backup snapshots for that cluster.
+Deleting a Local deployment also deletes any local data volumes.
+
+` + fmt.Sprintf(usage.RequiredRole, "Project Owner"),
```
blva, line 156: we should probably remove this as it's only true for local

blva (approved):

zach-carr (approved):
LGTM! Nice use of active voice - that's a core MongoDB style thing ([page](https://www.mongodb.com/docs/meta/style-guide/writing/use-active-voice/))

internal/cli/atlas/deployments/delete.go
tibulca: this translates to: `To use this command, you must authenticate with a user account or an API key that has the Project Owner role.` which is still correct for Atlas deployments. Your suggestion is to remove it entirely or to adjust it to something like: `To use this command with Atlas deployments, you must authenticate with a user account or an API key that has the Project Owner role.`

PR #2356, opened 2023-10-11
CLOUDP-205691: support go 1.21

wtrocki:
## Proposed changes

Add support for latest golang

_Jira ticket: CLOUDP-205691

### DIFF
--- a/.tool-versions
+++ b/.tool-versions
@@ -1,1 +1,1 @@
-golang 1.20
+golang 1.21.3
--- a/go.mod
+++ b/go.mod
@@ -2,8 +2,6 @@ module github.com/mongodb/mongodb-atlas-cli
 
 go 1.21
 
-toolchain go1.21.3
-
 require (
 	cloud.google.com/go/kms v1.15.2
 	github.com/AlecAivazis/survey/v2 v2.3.7
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,7 @@
 module github.com/mongodb/mongodb-atlas-cli
 
-go 1.21.3
+go 1.21
+toolchain go1.21.3
 
 require (
 	cloud.google.com/go/kms v1.15.2
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/mongodb/mongodb-atlas-cli
 
-go 1.20
+go 1.21.3
 
 require (
 	cloud.google.com/go/kms v1.15.2
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,7 @@
 module github.com/mongodb/mongodb-atlas-cli
 
 go 1.21
+
 toolchain go1.21.3
 
 require (
--- a/internal/kubernetes/operator/version/provider.go
+++ b/internal/kubernetes/operator/version/provider.go
@@ -25,7 +25,8 @@ import (
 )
 
 const (
-	operatorRepositoryOrg     = "mongodb"
+	operatorRepositoryOrg = "mongodb"
+	//nolint:gosec
 	operatorRepository        = "mongodb-atlas-kubernetes"
 	maxMajorVersionsSupported = 3
 )
--- a/.github/workflows/code-health.yml
+++ b/.github/workflows/code-health.yml
@@ -18,7 +18,7 @@ jobs:
       - name: golangci-lint
         uses: golangci/golangci-lint-action@v3.7.0
         with:
-          version: v1.53.3
+          version: v1.54.2
   unit-tests:
     env:
       COVERAGE: coverage.out
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -63,7 +63,7 @@ To make the contribution process as seamless as possible, we ask for the followi
 #### Prerequisite Tools
 
 - [Git](https://git-scm.com/)
-- [Go (at least Go 1.20)](https://golang.org/dl/)
+- [Go (at least Go 1.21.3)](https://golang.org/dl/)
 
 #### Environment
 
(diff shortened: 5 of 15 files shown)


wtrocki:
evergreen keep definitions

go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/mongodb/mongodb-atlas-cli
 
-go 1.20
+go 1.21.3
```
helderjs, line 3: Should we be tied to a minor version?

wtrocki: No strong opinion. What do you do in kubernetes operator?

For context 
https://go.dev/blog/toolchain

wtrocki: I have used toolchain keyword and downgraded to 1.20 without patch

helderjs: I see... I think there is no concern here.

internal/kubernetes/operator/version/provider.go
```diff
@@ -25,7 +25,8 @@ import (
 )
 
 const (
-	operatorRepositoryOrg     = "mongodb"
+	operatorRepositoryOrg = "mongodb"
+	//nolint:gosec
 	operatorRepository        = "mongodb-atlas-kubernetes"
```
wtrocki, line 30: Apparently this string looks like leaked credentials :)

.tool-versions
```diff
@@ -1 +1 @@
-golang 1.20
+golang 1.21.3
```
wtrocki, line 1: Specifically, here we want to ensure the same toolchain is used

go.mod
```diff
@@ -1,6 +1,8 @@
 module github.com/mongodb/mongodb-atlas-cli
 
-go 1.20
+go 1.21
+
+toolchain go1.21.3
```
andreaangiolillo, line 5: Are there any reasons why users cannot use the toolchain 1.21.0 / 1.21.1 / 1.21.2?  if not, I would remove this

andreaangiolillo (commented):
Left one comment

fmenezes (dismissed):

go.mod
fmenezes: According to https://go.dev/doc/toolchain if you run using 1.21.0-2 or older like 1.20 the toolchain will automatically download go 1.21.3 compile it and continue it's flow, if running 1.21.4+ or even 1.22 it will continue believing the compatibility promise.

I think this is a good change.

internal/kubernetes/operator/version/provider.go
fmenezes: nit: I prefer the inline style, I think it's clearer to tell which line was affected
```suggestion
	operatorRepository        = "mongodb-atlas-kubernetes" //nolint:gosec // false positive
```

go.mod
wtrocki: This is a good pattern to align toolchain versions across asdf, evergreen, and github actions. 
Go toolchains add an additional layer of go versioning so I wanted to be more explicit to ensure alignment.

Context: 
https://go.dev/blog/toolchain#:~:text=If%20you%20are,the%20module%20cache.

I have no strong opinion to keep it. This way we can also align other dependencies like Kubernetes operator etc.

Why:

With 1.21 we cannot anymore use 1.20 as a universal binary that always points to latest and tools like ads need to specifically specify the patch version:

Non Patch
```
asdf install golang 1.21
URL: https://dl.google.com/go/go1.21.darwin-arm64.tar.gz returned status 404
```
With patch:
```
▶ asdf install golang 1.21.3            
golang 1.21.3 is already installed
```

andreaangiolillo: > I think this is a good change.
Does this force us to update the toolversion for each new patch version of 1.21? If there is no reason we cannot run the CLI with previous patch versions of 1.21, I don't understand why we need this 🤔

wtrocki:
@andreaangiolillo Removing toolchain from variables

go.mod
fmenezes: @andreaangiolillo this is not blocking old versions (like 1.21.1) it is actually improving the scenario (if you run on a very old version you will still compile on a newer compiler).

.tool-versions
andreaangiolillo: According to https://go.dev/doc/toolchain

> If the toolchain line is omitted, the module or workspace is considered to have an implicit toolchain goV line, where V is the Go version from the go line.

So I would set this version to be the same as the go version in go mod if we don't use the toolchain

go.mod
andreaangiolillo: > @andreaangiolillo this is not blocking old versions (like 1.21.1) it is actually improving the scenario (if you run on a very old go version you will still compile on a newer compiler).

I understood was not blocking older versions. I am wondering if we will need to keep updating the toolchain version for each new patch release of 1.21. 

Btw my comment is not blocking, I am just trying to understand how we want to maintain the toolchain version going forward and if there was a specific reason to select 1.21.3 besides being the latest patch release.

fmenezes: since toolchain was removed we can leave it for now

fmenezes: My assumption was to update it together with go versions

.tool-versions
wtrocki: See: https://github.com/mongodb/mongodb-atlas-cli/pull/2356#discussion_r1356620224

go.mod
gssbzn: >This is a good pattern to align toolchain versions across asdf, evergreen, and github actions.

Not necessarily true, for example right now packaging happens in evg (we won't change that due to notarization/signing), up until this change we were using the latest patch version when building distributables, which we won't do anymore, also evergreen has usually discouraged from pining to a patch for this same reason

IMHO the new toolchain feature replaces asdf and we should consider deprecating from this repo, we also need to decide what to do about evg, do we need to keep a reminder/task to keep updating patch versions to make sure we package with included security fixes or we should stick to the current evg recommended pattern to not pin to a minor version

wtrocki:
Follow up. Let's try to change the version to Evergreen to let it autosimlink.

PR #2379, opened 2023-10-20
CLOUDP-202471: add support to track GH Actions telemetry

matt-condon:
## Proposed changes

_Jira ticket:_ CLOUDP-202471

Adding telemetry support to GH Actions (API Telemetry and CLI Telemetry)

Setting hostname via the same mechanism as docker container (env var) 
Small change to how - using new shared env var `MONGODB_ATLAS_HOSTNAME` with explicit hostname allow-list

## Checklist

### DIFF
--- a/internal/oauth/oauth.go
+++ b/internal/oauth/oauth.go
@@ -47,6 +47,12 @@ var defaultTransport = &http.Transport{
 	ExpectContinueTimeout: expectContinueTimeout,
 }
 
+var validHostnames = map[string]bool{
+	config.NativeHostName:    true,
+	config.ContainerHostName: true,
+	config.ActionsHostName:   true,
+}
+
 type ServiceGetter interface {
 	Service() string
 	OpsManagerURL() string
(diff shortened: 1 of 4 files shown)


internal/oauth/oauth.go
```diff
@@ -58,10 +64,16 @@ const (
...
 )
 
-func changeHostNameToContainer() {
-	if !strings.Contains(config.UserAgent, config.ContainerHostName) {
-		config.UserAgent = strings.ReplaceAll(config.UserAgent, config.HostName, config.ContainerHostName)
-		config.HostName = config.ContainerHostName
+func patchConfigHostname() {
+	hostnameEnvVal := os.Getenv(config.HostnameEnv)
+
+	if hostnameEnvVal == "" && (os.Getenv("MONGODB_ATLAS_IS_CONTAINERIZED") == "true") {
```
andreaangiolillo, line 70: Does it make sense to consider the `MONGODB_ATLAS_IS_CONTAINERIZED` as true when is defined instead of explicitly check for its value? I am wondering if there is a scenario where the var is defined with a different value than true.

internal/oauth/oauth.go
```diff
@@ -47,6 +47,12 @@ var defaultTransport = &http.Transport{
 	ExpectContinueTimeout: expectContinueTimeout,
 }
 
+var validHostnames = map[string]bool{
+	config.NativeHostName:    true,
+	config.ContainerHostName: true,
+	config.ActionsHostName:   true,
```
andreaangiolillo, line 53: Could you add a comment beside the configs to explain the differences between this hostname?

andreaangiolillo, line 53: Action is quite generic and if someone does not have context, it may not be obvious that we are taking about github action. 
```suggestion
	config.GitHubActionsHostName:   true,
```
Another name could be CIPipeline

PR #2383, opened 2023-10-23
task: enable testifylint

gssbzn:
Update `golangci-lint`and enable the new `testifylint` so our test are more consistent if we use that lib 

This also fixes a lot of the inconsistencies, the main one the linter is opinionated on `require.Error` over `assert.Error` which we have a bit lax on


wtrocki (approved):
LGTM

wtrocki (approved):

PR #2440, opened 2023-11-10
CLOUDP-209403 Update error message regarding credentials for atlas in local dev

fmenezes:
## Proposed changes
Update error message regarding credentials for atlas in local dev

_Jira ticket:_ CLOUDP-209403

## Checklist

### DIFF
--- a/internal/cli/refresher_opts.go
+++ b/internal/cli/refresher_opts.go
@@ -25,6 +25,8 @@ import (
 	atlas "go.mongodb.org/atlas/mongodbatlas"
 )
 
+var TokenRefreshed bool
+
 type RefresherOpts struct {
 	flow Refresher
 }
@@ -58,6 +60,7 @@ func (opts *RefresherOpts) RefreshAccessToken(ctx context.Context) error {
 		return err
 	}
 	if current.Valid() {
+		TokenRefreshed = true
 		return nil
 	}
 	t, _, err := opts.flow.RefreshToken(ctx, config.RefreshToken())
@@ -73,7 +76,11 @@ func (opts *RefresherOpts) RefreshAccessToken(ctx context.Context) error {
 	}
 	config.SetAccessToken(t.AccessToken)
 	config.SetRefreshToken(t.RefreshToken)
-	return config.Save()
+	if err := config.Save(); err != nil {
+		return err
+	}
+	TokenRefreshed = true
+	return nil
 }
 
 func (opts *RefresherOpts) PollToken(c context.Context, d *atlasauth.DeviceCode) (*atlasauth.Token, *atlas.Response, error) {
--- a/internal/cli/atlas/deployments/list.go
+++ b/internal/cli/atlas/deployments/list.go
@@ -17,6 +17,7 @@ package deployments
 import (
 	"context"
 	"errors"
+	"fmt"
 
 	"github.com/mongodb/mongodb-atlas-cli/internal/cli"
 	"github.com/mongodb/mongodb-atlas-cli/internal/cli/atlas/deployments/options"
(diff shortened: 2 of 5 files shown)


tibulca (approved):
LGTM if LGTEV

gssbzn:
in this branch
```
./bin/atlas deployments ls -P dev
NAME        TYPE    MDB VER   STATE
local7886   LOCAL   7.0.2     STOPPED
Error: failed to retrieve Atlas deployments: https://cloud-dev.mongodb.com/api/atlas/v2/groups/5e429f2e06822c6eac4d59c9/clusters GET: HTTP 500 Internal Server Error (Error code: "") Detail:  Reason: Server Error. Params: []
```

I was under the impression the goal was to stop doing the API call

gssbzn:
this is on the prod build
```
atlas deployments ls -P dev
Could not refresh access token: session expired

To login, run: atlas auth login
NAME        TYPE    MDB VER   STATE
local7886   LOCAL   7.0.2     STOPPED
Error: failed to retrieve Atlas deployments with: https://cloud-dev.mongodb.com/api/atlas/v2/groups/5e429f2e06822c6eac4d59c9/clusters GET: HTTP 500 Internal Server Error (Error code: "") Detail:  Reason: Server Error. Params: []
```

fmenezes:
> in this branch
> 
> ```
> ./bin/atlas deployments ls -P dev
> NAME        TYPE    MDB VER   STATE
> local7886   LOCAL   7.0.2     STOPPED
> Error: failed to retrieve Atlas deployments: https://cloud-dev.mongodb.com/api/atlas/v2/groups/5e429f2e06822c6eac4d59c9/clusters GET: HTTP 500 Internal Server Error (Error code: "") Detail:  Reason: Server Error. Params: []
> ```
> 
> I was under the impression the goal was to stop doing the API call

I decided to not avoid doing API call, there is a bit of state management involved that we would need to change, the way we deal with auth basically happens at the builder/persistentprerun as opposed to in each command which made things complicated.

I decided to catch the 401 error and make a better message, in your original post you were suffering from auth issues, this time you won't get this confusion. On your current example you are not having any auth issues so the message is not the same.

gssbzn (changes requested):
In the current state I don't think this really address the issue, for example in multiple occasions I got errors from podman but I the double error on API and podman makes the experience worst 

The issue here is not asking users to login (which I believe we wanted to keep) but the extraneous warning when nothing really went wrong

fmenezes:
@gssbzn I've updated the approach now you will see
```
atlas deployments ls -P dev
Could not refresh access token: session expired

To login, run: atlas auth login
NAME        TYPE    MDB VER   STATE
local7886   LOCAL   7.0.2     STOPPED
```
instead of
```
atlas deployments ls -P dev
Could not refresh access token: session expired

To login, run: atlas auth login
NAME        TYPE    MDB VER   STATE
local7886   LOCAL   7.0.2     STOPPED
Error: failed to retrieve Atlas deployments with: https://cloud-dev.mongodb.com/api/atlas/v2/groups/5e429f2e06822c6eac4d59c9/clusters GET: HTTP 500 Internal Server Error (Error code: "") Detail:  Reason: Server Error. Params: []
```

internal/cli/refresher_opts.go
```diff
@@ -25,6 +25,8 @@ import (
 	atlas "go.mongodb.org/atlas/mongodbatlas"
 )
 
+var TokenRefreshed bool
```
gssbzn, line 28: [nit] let's make it readonly just in case  
```suggestion
var tokenRefreshed bool
func TokenRefreshed() bool {
  return tokenRefreshed
}
```

gssbzn (commented):
thanks for taking some time to look more into this, one small nit then good to merge

internal/cli/refresher_opts.go
fmenezes: Unfortunately I can't because I need to set it on tests, across different packages

gssbzn (approved):

PR #2458, opened 2023-11-17
CLOUDP-211540 Delete protected clusters on clean up

fmenezes:
## Proposed changes
Delete protected clusters on clean up

_Jira ticket:_ CLOUDP-211540

## Checklist

### DIFF
--- a/test/e2e/atlas/helper_test.go
+++ b/test/e2e/atlas/helper_test.go
@@ -879,6 +879,25 @@ func listDataFederationsByProject(t *testing.T, cliPath, projectID string) []atl
 	return dataFederations
 }
 
+func listServerlessByProject(t *testing.T, cliPath, projectID string) []atlasv2.ServerlessInstanceDescription {
+	t.Helper()
+
+	cmd := exec.Command(cliPath,
+		serverlessEntity,
+		"list",
+		"--projectId", projectID,
+		"-o=json")
+	cmd.Env = os.Environ()
+	resp, err := cmd.CombinedOutput()
+	require.NoError(t, err)
+
+	var serverlessInstances []atlasv2.ServerlessInstanceDescription
+	err = json.Unmarshal(resp, &serverlessInstances)
+	require.NoError(t, err)
+
+	return serverlessInstances
+}
+
 func deleteAllDataFederations(t *testing.T, cliPath, projectID string) {
 	t.Helper()
 
@@ -890,6 +909,19 @@ func deleteAllDataFederations(t *testing.T, cliPath, projectID string) {
 	}
 
 	t.Log("all datafederations successfully deleted")
+}
+
+func deleteAllServerlessInstances(t *testing.T, cliPath, projectID string) {
+	t.Helper()
+
+	serverlessInstances := listServerlessByProject(t, cliPath, projectID)
+
+	for _, instance := range serverlessInstances {
+		err := deleteServerlessInstanceForProject(projectID, instance.GetName())
+		require.NoError(t, err)
+	}
+
+	t.Log("all serverless instances successfully deleted")
 }
 
 func deleteDataFederationForProject(projectID, dataFedName string) error {
--- a/test/e2e/atlas/helper_test.go
+++ b/test/e2e/atlas/helper_test.go
@@ -197,11 +197,9 @@ func watchServerlessInstanceForProject(projectID, clusterName string) error {
 	return watchCmd.Run()
 }
 
-func deleteServerlessInstanceForProject(projectID, clusterName string) error {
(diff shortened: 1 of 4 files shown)


fmenezes:
I've tested but creating project id 6557a52c638cb10274fc1c94 with a project with termination protection and saw it being removed on the latest run. Other projects were not removed due to be creating other resources like serverless.

test/e2e/atlas/helper_test.go
```diff
@@ -302,9 +313,43 @@ func deleteClusterForProject(projectID, clusterName string) error {
...
+	if resp, err := updateCmd.CombinedOutput(); err != nil {
+		return fmt.Errorf("error updating cluster %w: %s", err, string(resp))
+	}
+
+	return watchCluster(projectID, clusterName)
+}
+
+func deleteClusterForProject(projectID, clusterName string) error {
+	if err := internalDeleteClusterForProject(projectID, clusterName); err != nil {
+		if !strings.Contains(err.Error(), "CANNOT_TERMINATE_CLUSTER_WHEN_TERMINATION_PROTECTION_ENABLED") {
```
tibulca, line 344: is this err generated immediately or only after the watchCluster timeout?
I wonder if it's not better to do a getCluster, check the `terminationProtectionEnabled` flag, then removeTerminationProtectionFromCluster (if this is enabled) and then internalDeleteClusterForProject

test/e2e/atlas/helper_test.go
```diff
@@ -292,6 +292,17 @@ func deleteClusterForProject(projectID, clusterName string) error {
 		return fmt.Errorf("error deleting cluster %w: %s", err, string(resp))
 	}
 
+	// this command will fail with 404 once the cluster is deleted
+	// we just need to wait for this to close the project
+	_ = watchCluster(projectID, clusterName)
+	return nil
```
tibulca, line 298: wdyt about something like?
```suggestion
	if watchCluster(projectID, clusterName) != nil {
	      return nil
	}
	return errors.New("cluster is still active after delete attempt")
```

test/e2e/atlas/helper_test.go
fmenezes: immediately, watch happens only if the resource passes validation.

I personally think getCluster is worse, only a subset of clusters are not being deleted because of termination protection if we add getCluster we will impact every single one of the clusters as opposed to only the small subset, leading to a much slower flow.

test/e2e/atlas/helper_test.go
fmenezes: I'm not too sure how to reproduce the situation you're suggesting, do we have that scenario already?

tibulca: I was looking at the watch command and there's no timeout for it (this was one of the possible scenarios to reproduce this), and also the watch status is printed for both success / error cases (so no need for additional logs here).

test/e2e/atlas/helper_test.go
tibulca: if it's immediate I'm ok with this flow, thanks for checking

tibulca (approved):

tibulca (commented):
code LGTM, but it seems there are some cleanup cases that are not handled correctly.
Check [evg](https://spruce.mongodb.com/task/mongodb_atlas_cli_master_e2e_cleanup_atlas_cleanup_e2e_patch_70be4c482e07d1c44aeef17e9b015e83952c8a17_6557a25f306615f267dce504_23_11_17_17_26_57?execution=1&sortBy=STATUS&sortDir=ASC): `HTTP 409 Conflict (Error code: "CANNOT_CLOSE_GROUP_ACTIVE_ATLAS_CLUSTERS")`

fmenezes:
> code LGTM, but it seems there are some cleanup cases that are not handled correctly. Check [evg](https://spruce.mongodb.com/task/mongodb_atlas_cli_master_e2e_cleanup_atlas_cleanup_e2e_patch_70be4c482e07d1c44aeef17e9b015e83952c8a17_6557a25f306615f267dce504_23_11_17_17_26_57?execution=1&sortBy=STATUS&sortDir=ASC): `HTTP 409 Conflict (Error code: "CANNOT_CLOSE_GROUP_ACTIVE_ATLAS_CLUSTERS")`

hard to investigate given it's past, but I'm submitting a change to delete all serverless instances to make it even more reliable

fmenezes:
@tibulca So far from all that I can see the error originates from clusters or serverless instances being in `DELETING` state, we would skip those and try to delete the project afterward.


I've added watch in those scenarios to try to delete project after all resources are guaranteed to be deleted.

tibulca (approved):
LGTM, thanks for the great improvements with performance too

PR #2500, opened 2023-12-07
fix: remove usages of cluster labels

gssbzn:
Labels are deprecated and will be removed, doing this clean up in the CLI now


gssbzn:
>why not remove in-line w/ SDK update/API update? it's a breaking change to the CLI as well right?

This feature is not exposed to users, it was a (poor) attempt to capture CLI usage via labels but at the same time we never had a way to check those labels, this is just cleaning those labels we never really used anyway

blva (approved):
LGTM

PR #2516, opened 2023-12-15
CLOUDP-218285: Failing upgrade tests

gssbzn:
## Proposed changes

_Jira ticket:_ CLOUDP-218285

Small clean up on how the upgrade tests works

### DIFF
--- a/test/e2e/atlas/clusters_upgrade_test.go
+++ b/test/e2e/atlas/clusters_upgrade_test.go
@@ -20,7 +20,6 @@ import (
 	"os"
 	"os/exec"
 	"testing"
-	"time"
 
 	"github.com/mongodb/mongodb-atlas-cli/test/e2e"
 	"github.com/stretchr/testify/assert"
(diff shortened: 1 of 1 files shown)


test/e2e/atlas/clusters_upgrade_test.go
```diff
@@ -102,35 +65,27 @@ func TestSharedClusterUpgradeToDedicatedTier(t *testing.T) {
 			"-o=json")
 		cmd.Env = os.Environ()
 		resp, err := cmd.CombinedOutput()
-		req.NoError(err, string(resp))
+		require.NoError(t, err, string(resp))
 	})
+	require.NoError(t, watchCluster(g.projectID, g.clusterName))
+	ensureClusterTier(t, cliPath, g.projectID, g.clusterName, tierM10)
```
tibulca, line 71: as this upgrade also changes the disk size, mdbVersion, and the tags, can we validate all these changes, not only the tier? (unless covered by another test)

gssbzn: the version change is a lie, the current default on shared tiers, probably at the time of writing this test the default of shared was 5.0 but it changes with time. I removed that flag from the update and added back the checks for disk size and implemented a check for tags

Test are still running so we'll wait for the test to pass

tibulca (approved):

tibulca (approved):

PR #2526, opened 2023-12-19
CLOUDP-216213: [AtlasCLI] device_id is missing from some events in CLI telemetry

gssbzn:
CLOUDP-216213: some telemetry fixes
- track errors for missing device_id
- correctly track authentication type for un-auth 
- start tracking disabling update alerts

### DIFF
--- a/internal/telemetry/event.go
+++ b/internal/telemetry/event.go
@@ -154,10 +154,11 @@ func withAnonymousID() EventOpt {
 	return func(event Event) {
 		id, err := machineid.ProtectedID(config.ToolName)
 		if err != nil {
+			event.Properties["device_id_err"] = err.Error()
 			_, _ = log.Debugf("error generating machine id: %v\n", err)
-		} else {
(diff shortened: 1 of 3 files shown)


internal/telemetry/event.go
```diff
@@ -201,16 +202,17 @@ func withOS() EventOpt {
 type Authenticator interface {
 	PublicAPIKey() string
 	PrivateAPIKey() string
+	AccessToken() string
 }
 
 func withAuthMethod(c Authenticator) EventOpt {
```
blva, line 208: are we defaulting to auth_method = nil if unauth or do we set this value to `unauth` somewhere else?

blva (approved):
LGTM 🚀

internal/telemetry/event.go
gssbzn: yes, default to not set which seemed to be the common pattern rather than setting to unauth

PR #2530, opened 2023-12-21
CLOUDP-219151: fix template for create azure/gcp privateEndpoints

tibulca:
## Proposed changes
fix template for create azure/gcp privateEndpoints

_Jira ticket:_ CLOUDP-219151

### DIFF
--- a/internal/cli/atlas/privateendpoints/azure/create.go
+++ b/internal/cli/atlas/privateendpoints/azure/create.go
@@ -43,7 +43,7 @@ func (opts *CreateOpts) initStore(ctx context.Context) func() error {
 	}
 }
 
-var createTemplate = "Private endpoint '{{.ID}}' created.\n"
+var createTemplate = "Private endpoint '{{.Id}}' created.\n"
 
 func (opts *CreateOpts) Run() error {
 	createRequest := opts.newPrivateEndpointConnection()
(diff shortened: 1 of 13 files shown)


internal/cli/atlas/privateendpoints/azure/create.go
```diff
@@ -43,7 +43,7 @@ func (opts *CreateOpts) initStore(ctx context.Context) func() error {
 	}
 }
 
-var createTemplate = "Private endpoint '{{.ID}}' created.\n"
+var createTemplate = "Private endpoint '{{.Id}}' created.\n"
```
tibulca, line 46: the fix

andreaangiolillo (dismissed):
LGTM

andreaangiolillo (approved):

internal/cli/atlas/privateendpoints/gcp/create.go
```diff
@@ -43,7 +43,7 @@ func (opts *CreateOpts) initStore(ctx context.Context) func() error {
 	}
 }
 
-var createTemplate = "GCP private endpoint '{{.ID}}' created.\n"
+var createTemplate = "GCP private endpoint '{{.Id}}' created.\n"
```
jwilliams-mongo, line 46: it doesn't appear as if we include the cloud provider in the message for the other providers. we should probably omit it here to remain consistent.
```suggestion
var createTemplate = "Private endpoint '{{.Id}}' created.\n"
```

jwilliams-mongo (approved):
@tibulca LGTM % one minor comment. We don't specify a cloud provider in the templates for AWS and Azure, so I think we can omit it for GCP. Conversely we can add it for the other two

andreaangiolillo (approved):

PR #2550, opened 2024-01-12
fix: handle pagination properly for events

gssbzn:
Some tests for events were flaking and seems like we were asking for no limits when paging so fixing that


colm-quinn (approved):
Nice improvement. LGTM

tibulca (approved):
LGTM

PR #2557, opened 2024-01-17
CLOUDP-202471: add support to track GH Actions telemetry

matt-condon:
## Proposed changes

_Jira ticket:_ CLOUDP-202471

Adding telemetry support to GH Actions (API Telemetry and CLI Telemetry)

Setting hostname via the same mechanism as docker container (env var) 
We now have 3 env vars that determine the composition of the hostname:

| Env Var      | Represents       | 
| ------------- |:--------------:|
| `ATLAS_GITHUB_ACTION`  | use of our predefined workflows set up in [atlas-github-action](https://github.com/mongodb/atlas-github-action/blob/main/action.yml)   |
| `GITHUB_ACTIONS`  | usage in any GitHub Action   | 
| `MONGODB_ATLAS_IS_CONTAINERIZED` | use of the [atlasCLI Docker image](https://hub.docker.com/r/mongodb/atlas) | 

## Checklist

## Further comments

Following up on https://github.com/mongodb/mongodb-atlas-cli/pull/2379, which was closed after some time


fmenezes (approved):
code lgtm

fmenezes:
> but will have a limitation on telemetry data for custom scripts using the GitHub Action

@matt-condon can you clarify this from your PR description?

matt-condon:
> > but will have a limitation on telemetry data for custom scripts using the GitHub Action
> 
> @matt-condon can you clarify this from your PR description?
> 
> Note: GitHub actions always set the env var `GITHUB_ACTIONS` no matter what you run (https://docs.github.com/en/actions/learn-github-actions/variables#default-environment-variables)

Sure, I was testing setting custom env vars from the base GH Action, but once the CLI was installed, any consequential actions that aren't `using` the base Action did not have the env var set, so we were missing that context

Using the `GITHUB_ACTIONS` env var seems like a nice workaround if it's available from the .env context, so I'll hold off on merging this one to test that change

internal/oauth/oauth.go
```diff
@@ -58,10 +65,17 @@ const (
...
 )
 
-func changeHostNameToContainer() {
-	if !strings.Contains(config.UserAgent, config.ContainerHostName) {
-		config.UserAgent = strings.ReplaceAll(config.UserAgent, config.HostName, config.ContainerHostName)
-		config.HostName = config.ContainerHostName
+// Patches the agent hostname based on set env vars.
+func patchConfigHostnameFromEnv() {
+	hostnameEnvVal := os.Getenv(config.HostNameEnv)
+	legacyContainerizedEnvIsPopulated := os.Getenv(config.LegacyContainerizedHostNameEnv) != ""
```
gssbzn, line 71:
```suggestion
	hostnameEnvVal := os.Getenv(config.HostNameEnv)
	_, ok := os.LookupEnv(config.LegacyContainerizedHostNameEnv)
```

matt-condon:
@fmenezes included the suggested change to use the native `GITHUB_ACTIONS` env var instead. Thanks for the suggestion!

The tradeoff is that we'll see traffic for all GH actions usage rather than specifically the GH Actions that we provide

I've pinged Jakub to confirm which option he prefers, but I'd imagine he'll have a preference for that over the also limited visibility of our Action's predefined workflows

matt-condon:
The tests are failing here and passing locally because the `GITHUB_ACTIONS` env var is set when running the tests on GH 🤣  
huh

matt-condon:
Discussed with Jakub and he has a preference for including both, so I've added checks for each of the env vars individually, to build the hostname in the useragent

internal/oauth/oauth.go
```diff
@@ -58,10 +58,32 @@ const (
...
+		config.HostName = hostnameEnvVal
+	}
+}
+
+func appendToHostName(builder *strings.Builder, configVal string) {
+	if configVal != "" {
+		if builder.Len() > 0 {
+			builder.WriteString("|")
+		}
+		builder.WriteString(configVal)
```
andreaangiolillo, line 86: [nit] removing a layer
```suggestion
	if configVal == "" {
	   return
	}
	if builder.Len() > 0 {
		builder.WriteString("|")
	}
	
	builder.WriteString(configVal)
```

internal/oauth/oauth_test.go
```diff
@@ -0,0 +1,108 @@
+// Copyright 2023 MongoDB Inc
```
andreaangiolillo, line 1:
```suggestion
// Copyright 2024 MongoDB Inc
```

internal/oauth/oauth.go
```diff
@@ -58,10 +58,32 @@ const (
 	GovClientID = "0oabtyfelbTBdoucy297" // GovClientID for production
 )
 
-func changeHostNameToContainer() {
-	if !strings.Contains(config.UserAgent, config.ContainerHostName) {
-		config.UserAgent = strings.ReplaceAll(config.UserAgent, config.HostName, config.ContainerHostName)
-		config.HostName = config.ContainerHostName
+// Patches the agent hostname based on set env vars.
```
andreaangiolillo, line 61:
```suggestion
// patchConfigHostnameFromEnvs patches the agent hostname based on set env vars.
```

andreaangiolillo (approved):
LGTM

internal/oauth/oauth.go
matt-condon: Updated, thanks for the suggestion!

internal/oauth/oauth.go
```diff
@@ -58,13 +58,36 @@ const (
...
+	var builder strings.Builder
+	if checker.IsPopulated(config.AtlasActionHostNameEnv) {
+		appendToHostName(&builder, config.AtlasActionHostName)
+	}
+	if checker.IsPopulated(config.GitHubActionsHostNameEnv) {
+		appendToHostName(&builder, config.GitHubActionsHostName)
+	}
+	if checker.IsPopulated(config.ContainerizedHostNameEnv) {
+		appendToHostName(&builder, config.DockerContainerHostName)
+	}
```
tibulca, line 72: [optional] I wonder if we should always append, either the value itself, or a `-`, if value not populated.
I'm thinking about the scenario when only one of the 3 values is populated, looking at the hostname, can we know which env var it represents?
wdyt?

internal/cli/envs/env_checker.go
gssbzn, line 1: is this helper necessary? and if yes this doesn't belong in the cli package since not CLI specific

internal/cli/envs/env_checker.go
```diff
@@ -0,0 +1,29 @@
...
+
+package envs
+
+import "os"
+
+type EnvChecker interface {
+	IsPopulated(env string) bool
+}
+
+type NewEnvChecker struct{}
```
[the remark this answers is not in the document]
matt-condon: Yep because when running the tests without mocking the env var `ATLAS_GITHUB_ACTION` is set which causes tests to fail when run in the GitHub CI

I couldn't see another way around it and imo it makes sense not to set/check env vars directly in tests

andreaangiolillo (dismissed):

internal/oauth/oauth.go
matt-condon: So the hostName if the CLI is run natively would be "-|-|-"?
It's included in the userAgent so at that point I wonder whether we would just consider extending the UserAgent for each possible hostname

ie: 
`UserAgent      = fmt.Sprintf("%s/%s (%s;%s;%s)", ToolName, version.Version, runtime.GOOS, runtime.GOARCH, HostName)`

changes to
`UserAgent      = fmt.Sprintf("%s/%s (%s;%s;%s;%s;%s)", ToolName, version.Version, runtime.GOOS, runtime.GOARCH, MaybeAtlasActionHostName, MaybeGitHubActionsHostName, MaybeDockerContainerHostName)`

I think it starts to get a bit messy either way so I'd slightly favour keeping it as is, but don't feel particularly strongly if you'd like a change

internal/oauth/oauth_test.go
```diff
@@ -0,0 +1,108 @@
...
+	}
+}
+
+type envCheckerMock struct {
+	MockValues map[string]bool
+}
+
+func (e envCheckerMock) IsPopulated(env string) bool {
+	return e.MockValues[env]
+}
```
gssbzn, line 108: what's the advantage og this abstraction over using `t.Setenv`?
Quoting go proverbs:
[Clear is better than clever.](https://www.youtube.com/watch?v=PAAkCSZUG1c&t=14m35s)

matt-condon: Same as above
Tests are failing in GH CI if using the actual env vars
There's no t.UnsetEnv method to avail of so mocking seems like the way

gssbzn: [t.Setenv](https://pkg.go.dev/testing#T.Setenv) should self clean, this smells of something wrong with your tests

matt-condon: We check if the env var exists or not, and there's no way to set an env var to not exist

We could pivot to check on specific values and set `ATLAS_GITHUB_ACTION` to `""` where we want it unpopulated if you think that's cleaner, but I don't think overriding the `ATLAS_GITHUB_ACTION` is ideal from a clarity perspective either

internal/oauth/oauth.go
tibulca: indeed, when all 3 values are not populated, the "-|-|-" is not helpful, so in this case the hostname should not be set to this value. 
I was thinking more about the scenario one at least one is, like "-|-|sameValue". Might be easier to parse / interpret this value rather than "someValue" (which might be AtlasActionHostName OR GitHubActionsHostName OR  DockerContainerHostName).
This is a [nit], so I'm ok with keeping it as it is, or extending the userAgent, as long as this won't have side effects or create confusion when interpreting the telemetry data

matt-condon: Sure I'll make this change

internal/config/profile.go
```diff
@@ -175,6 +180,51 @@ func Exists(name string) bool {
...
+
+	if isDefaultHostName(configHostName) {
+		return NativeHostName
+	}
+	return configHostName
+}
+
+func envIsPopulated(env string) bool {
+	return os.Getenv(env) == "true"
+}
```
gssbzn, line 213: sometimes people will use 1|0 so 
```suggestion
func envIsTrue(env string) bool {
	return IsTrue(os.Getenv(env))
}
```

gssbzn (dismissed):
LGTM

internal/config/profile_test.go
```diff
@@ -159,3 +161,77 @@ func TestConfig_IsTrue(t *testing.T) {
...
+	}
+	for _, tt := range tests {
+		fields := tt.fields
+		expectedHostName := tt.expectedHostName
+		t.Run(tt.name, func(t *testing.T) {
+			t.Setenv(AtlasActionHostNameEnv, fields.atlasActionEnv)
+			t.Setenv(GitHubActionsHostNameEnv, fields.ghActionsEnv)
+			t.Setenv(ContainerizedHostNameEnv, fields.containerizedEnv)
+
+			HostName = NativeHostName
```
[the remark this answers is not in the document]
matt-condon: Nope, leftover from moving the test, I'll remove

internal/config/profile.go
matt-condon: Thanks for the tip! Added

internal/oauth/oauth.go
matt-condon: Updated the tests to describe the different scenarios
Let me know if that's good with you

gssbzn (approved):
LGTM

PR #2601, opened 2024-02-07
CLOUDP-228406: [cli] Rename teams

gssbzn:
## Proposed changes

_Jira ticket:_ CLOUDP-228406

```sh
./bin/atlas teams  rename test-test --teamId <>                        
Team 'test-test' updated.
```


internal/cli/atlas/teams/rename.go
```diff
@@ -0,0 +1,100 @@
...
+		Name: opts.name,
+	}
+}
+
+// RenameBuilder atlas team(s) update <newName> [--orgId orgId].
+func RenameBuilder() *cobra.Command {
+	opts := &renameOpts{}
+	cmd := &cobra.Command{
+		Use:     "rename <newName>",
+		Aliases: []string{"update"},
```
tibulca, line 68: just a thought (no change needed), after looking at the `atlas teams create <team> --username=...`, I was thinking initially of not using this alias for rename, as it will block in the future the creation of an `atlas teams update --username=...` command. But saw that for users management we have `atlas teams users` sub-command

tibulca (approved):
LGTM

internal/cli/atlas/teams/rename.go
gssbzn: yes, I did a pass also on the API and this is the update operation but only names are available for update, I'm not sure if this will change in the future and why going with rename as a command but update should also be valid

docs/atlascli/command/atlas-teams-rename.txt
```diff
@@ -0,0 +1,105 @@
...
+
+.. contents:: On this page
+   :local:
+   :backlinks: none
+   :depth: 1
+   :class: singlecol
+
+Rename a team in your organization.
+
+To use this command, you must authenticate with a user account or an API key that has the Organization Owner role.
```
kyuan-mongodb, line 17:
```suggestion
To use this command, you must authenticate with a user account or have an API key with the Organization Owner role.
```

docs/atlascli/command/atlas-teams-rename.txt
```diff
@@ -0,0 +1,105 @@
...
+     - false
+     - help for rename
+   * - --orgId
+     - string
+     - false
+     - Organization ID to use. This option overrides the settings in the configuration file or environment variable.
+   * - -o, --output
+     - string
+     - false
+     - Output format. Valid values are json, json-path, go-template, or go-template-file. To see full output, use the -o json option.
```
kyuan-mongodb, line 67:
```suggestion
     - Output format. Valid values are json, json-path, go-template, or go-template-file. To see the full output, use the -o json option.
```

kyuan-mongodb (approved):
LGTM % 2 minor wording fixes.

gssbzn:
@kyuan-mongodb I'm going to merge as is since the nits are on some copy which is globally defined and will impact probably every command making this PR harder to review but I do think your recommendations make sense so I'll do as a follow up

PR #2617, opened 2024-02-09
task: fix unused args for anon functions

gssbzn:
Linting update now flags unused params on anonymous functions, first of a couple of PRs fixing this (splitting to make the review easy), this one targets the test folder

### DIFF
--- a/test/e2e/atlas/access_lists_test.go
+++ b/test/e2e/atlas/access_lists_test.go
@@ -34,7 +34,6 @@ func TestAccessList(t *testing.T) {
 	g.generateProject("accessList")
 
 	n, err := e2e.RandInt(255)
-	a := assert.New(t)
 	req := require.New(t)
 	req.NoError(err)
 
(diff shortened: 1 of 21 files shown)


test/e2e/atlas/access_lists_test.go
```diff
@@ -55,11 +54,10 @@ func TestAccessList(t *testing.T) {
 			"-o=json")
 		cmd.Env = os.Environ()
 		resp, err := cmd.CombinedOutput()
-		req.NoError(err)
+		require.NoError(t, err)
```
lantoli, line 57: don't you delete req in line 38?

tibulca (approved):

test/e2e/atlas/access_lists_test.go
gssbzn: it's used outside of the anon function for `t.Run`, the main focus here was to avoid relying on `require` or `assert` from the closure, which then the linter was complaining

PR #2626, opened 2024-02-12
task: update golangci-lint to 1.56.1

gssbzn:
Update to the latest golangci-lint
Some major changes include it now detects unused params in  anon funcs and better detection of ID vs Id

Additionally it includes support for go1.22


blva (approved):
LGTM if LGTEV

PR #2629, opened 2024-02-12
[CLOUDP-229226] Search node management - implement list command

jeroenvervaeke:
## Proposed changes

Added `atlas cluster search nodes list --clusterName clusterName [--projectId projectId]` command.

_Jira ticket:_ CLOUDP-229226

PR #2644, opened 2024-02-14
Remarks missed in PR 2629

jeroenvervaeke:
Remarks I've missed in #2629

### DIFF
--- a/internal/cli/atlas/search/nodes/list.go
+++ b/internal/cli/atlas/search/nodes/list.go
@@ -1,4 +1,4 @@
-// Copyright 2020 MongoDB Inc
+// Copyright 2024 MongoDB Inc
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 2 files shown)


tibulca (approved):

blva (approved):

internal/cli/atlas/search/nodes/list.go
```diff
@@ -1,4 +1,4 @@
-// Copyright 2020 MongoDB Inc
+// Copyright 2024 MongoDB Inc
```
lantoli, line 1: in Amazon the year in copyright was "from" when it was created the file, it was not needed to changed it, don't know here, i would assume there is no need either

jeroenvervaeke: The file was created this week. I copied the copyright from somewhere else. That's why it's updated to 2024

PR #2643, opened 2024-02-13
[Atlas CLI] search node management - implement create command

jeroenvervaeke:
## Proposed changes

Added `atlas cluster search nodes list --clusterName clusterName [--projectId projectId]` command.

_Jira ticket:_ CLOUDP-229227

PR #2647, opened 2024-02-15
CLOUDP-229229: search node management - implement delete command

jeroenvervaeke:
## Proposed changes

Added `atlas cluster search nodes delete --clusterName clusterName [--projectId projectId]` command.

_Jira ticket:_ CLOUDP-229229


internal/cli/atlas/search/nodes/delete.go
```diff
@@ -0,0 +1,129 @@
...
+	"github.com/mongodb/mongodb-atlas-cli/internal/usage"
+	"github.com/spf13/afero"
+	"github.com/spf13/cobra"
+)
+
+type DeleteOpts struct {
+	cli.GlobalOpts
+	cli.WatchOpts
+	*cli.DeleteOpts
+	fs    afero.Fs
```
gssbzn, line 36: unused?

gssbzn (commented):
having a delete is usually a good point to add e2e or are we planning e2e as a separate ticket?

jeroenvervaeke:
> having a delete is usually a good point to add e2e or are we planning e2e as a separate ticket?

I've got a separate ticket for that. I'm also waiting for a docs review for #2643. Can't delete anything in my e2e tests yet because I can't create anything yet

internal/cli/atlas/search/nodes/delete.go
jeroenvervaeke: Good catch, I've removed it

lmkerbey-mdb (dismissed):
LGTM.

internal/cli/atlas/search/nodes/delete.go
```diff
@@ -0,0 +1,126 @@
...
+
+	if strings.Contains(err.Error(), "ATLAS_FTS_DEPLOYMENT_DOES_NOT_EXIST") {
+		return true, nil
+	}
+
+	return false, err
+}
+
+// atlas clusters search nodes delete [--clusterName clusterName].
+func DeleteBuilder() *cobra.Command {
```
blva, line 89: could we add unit tests for delete?

internal/cli/atlas/search/nodes/delete.go
```diff
@@ -0,0 +1,126 @@
...
+	}
+
+	if opts.EnableWatch {
+		if err := opts.Watch(opts.watcher); err != nil {
+			return err
+		}
+		opts.Template = deleteWatchTemplate
+	} else {
+		opts.Template = deleteTemplate
+	}
```
tibulca, line 69: I think this is redundant, right? (it's set in this line `opts.InitOutput(cmd.OutOrStdout(), deleteTemplate),`)
```suggestion
	}```

**Review by blva — CHANGES_REQUESTED** [review]

File: internal/cli/atlas/search/nodes/delete.go [hunk 4]
```diff
@@ -0,0 +1,126 @@
...
+
+	return opts.Print(opts.Entry)
+}
+
+func (opts *DeleteOpts) watcher() (bool, error) {
+	_, err := opts.store.SearchNodes(opts.ConfigProjectID(), opts.Entry)
+	// searchNodes returns an error when there are no search nodes available
+	if err == nil {
+		return false, nil
+	}
```
blva, line 79:
```suggestion
```

this can be removed and you can update your check below to err != nil && ...

internal/cli/atlas/search/nodes/delete.go
```diff
@@ -0,0 +1,126 @@
...
+
+	if strings.Contains(err.Error(), "ATLAS_FTS_DEPLOYMENT_DOES_NOT_EXIST") {
+		return true, nil
+	}
```
blva, line 83: nit: you can simplify this a bit more

```golang
	if _, err := opts.store.SearchNodes(opts.ConfigProjectID(), opts.Entry); err != nil {
		if strings.Contains(err.Error(), "ATLAS_FTS_DEPLOYMENT_DOES_NOT_EXIST") {
		return true, nil
	         }
	}

```

internal/cli/atlas/search/nodes/delete.go
```diff
@@ -0,0 +1,126 @@
...
+
```
tibulca, line 84: what do you think about adding a 404 NOT FOUND check to safeguard for future updates on this API? (when the API might return 404, instead of 400 as today)?
```suggestion
		var atlasErr *atlas.ErrorResponse
		if errors.As(err, &atlasErr) && atlasErr.HTTPCode == 404 {
			return true, nil
		}
```

jeroenvervaeke: Added

internal/cli/atlas/search/nodes/delete.go
jeroenvervaeke: Done

internal/cli/atlas/search/nodes/delete.go
jeroenvervaeke: Removed

internal/cli/atlas/search/nodes/delete.go
gssbzn: jumping here, `strings.Contains(error.Error())` is an anti-pattern
please use `admin.AsError(err)` like
```
target, ok := atlasv2.AsError(err)
	if ok && target.GetErrorCode() == "ATLAS_FTS_DEPLOYMENT_DOES_NOT_EXIST" {
		...
	} 
```

internal/cli/atlas/search/nodes/delete.go
```diff
@@ -0,0 +1,132 @@
...
+	// Fallback case in case the backend starts returning 404 instead of the 400 it's returning currently
+	var atlasErr *atlas.ErrorResponse
+	if errors.As(err, &atlasErr) && atlasErr.HTTPCode == 404 {
```
gssbzn, line 87: Please avoid introducing dependencies to the deprecated sdk 
```suggestion
	target, ok := atlasv2.AsError(err)
	if errors.As(err, &target) && atlasErr.GetHTTPCode() == 404 {
```

jeroenvervaeke: @tibulca, are you ok with me removing this check?

tibulca: let's keep the check, but with the changes proposed by Gus (using the new SDK `atlasv2` which I missed in my original suggestion)

jeroenvervaeke: Fixed it

internal/cli/atlas/search/nodes/delete.go
```diff
@@ -0,0 +1,128 @@
...
+var deleteWatchTemplate = "Search nodes for cluster '{{.}}' are deleted.\n"
+
+func (opts *DeleteOpts) Run() error {
+	if err := opts.Prompt(); err != nil {
+		return nil
+	}
+
+	if !opts.Confirm {
+		fmt.Println(opts.FailMessage())
+		return nil
```
blva, line 58:
```suggestion
		return err
```

internal/cli/atlas/search/nodes/delete.go
```diff
@@ -0,0 +1,128 @@
...
+
+	return opts.Print(opts.Entry)
+}
+
+func (opts *DeleteOpts) watcher() (bool, error) {
+	_, err := opts.store.SearchNodes(opts.ConfigProjectID(), opts.Entry)
+	// searchNodes returns an error when there are no search nodes available
+	if err == nil {
+		return false, nil
+	}
```
blva, line 79:
```suggestion
```

[nit] I think this is redundant with your last return and you could remove it

Can you double-check `atlasv2.AsError(err)` handles nil errors (I believe it does as per `errors.As`
```
func As(err error, target any) bool {
	if err == nil {
		return false
	}
```

internal/cli/atlas/search/nodes/delete.go
jeroenvervaeke: There's no error to return here. `err` is undefined

internal/cli/atlas/search/nodes/delete.go
jeroenvervaeke: Checked and removed the nil check

tibulca (dismissed):

internal/cli/atlas/search/nodes/delete.go
```diff
@@ -0,0 +1,123 @@
...
+		return err
+	}
+}
+
+var deleteTemplate = "Started deleting search nodes for cluster '{{.}}'.\n"
+var deleteWatchTemplate = "Search nodes for cluster '{{.}}' are deleted.\n"
+
+func (opts *DeleteOpts) Run() error {
+	if err := opts.Prompt(); err != nil {
+		return nil
```
tibulca, line 53:
```suggestion
		return err
```

internal/cli/atlas/search/nodes/delete.go
blva: sorry i think I suggested the wrong place, it should be in line 53

internal/cli/atlas/search/nodes/delete.go
jeroenvervaeke: Fixed

tibulca (approved):

blva (approved):
LGTM, thanks for all the changes!

PR #2666, opened 2024-02-21
CLOUDP-230402: Implemented project tagging

jeroenvervaeke:
## Proposed changes
Implemented project tagging.

_Jira ticket:_ CLOUDP-230402


docs/atlascli/command/atlas-projects-create.txt
```diff
@@ -75,6 +75,10 @@ Options
      - string
      - false
      - Unique 24-digit string that identifies the Atlas user to be granted the Project Owner role on the specified project. If unspecified, this value defaults to the user ID of the oldest Organization Owner.
+   * - --tag
+     - stringToString
```
gssbzn, line 79: for @mongodb/docs-cloud-team I think we should update https://github.com/mongodb-labs/cobra2snooty/blob/main/flags.go to better handle these

test/e2e/atlas/projects_test.go
gssbzn, line 1: I belive the test you are looking for is https://github.com/mongodb/mongodb-atlas-cli/blob/cee6a16ebc0068ad0207f810f8c64b8af0cb91d3/test/e2e/iam/atlas_projects_test.go

jeroenvervaeke: Moved the test

test/e2e/iam/projects_test.go
```diff
@@ -90,6 +93,33 @@ func TestProjects(t *testing.T) {
...
+			t.Fatalf("unexpected error: %v", err)
+		}
+
+		require.Len(t, project.GetTags(), 2)
+
+		expectedTags := map[string]string{"env": "e2e", "prod": "false"}
+		for _, tag := range project.GetTags() {
+			expectedValue, ok := expectedTags[tag.GetKey()]
+			if !ok {
+				t.Errorf("expected %v, to have key %s\n", project.Tags, tag.GetKey())
```
tibulca, line 116: I think this is more like: `unexpected tag key %s` as the `tag` var is actually part of the projects.Tags

test/e2e/iam/projects_test.go
```diff
@@ -90,6 +93,33 @@ func TestProjects(t *testing.T) {
...
+			}
+
+			require.Equal(t, expectedValue, tag.GetValue())
+		}
```
tibulca, line 120: what about: 
```
		expectedTags := []atlasv2.ResourceTag{
			{Key: pointer.Get("env"), Value: pointer.Get("e2e")},
			{Key: pointer.Get("prod"), Value: pointer.Get("false")},
		}
		require.Equal(t, expectedTags, project.GetTags())
```

test/e2e/atlas/projects_test.go
jeroenvervaeke: I've actually reverted this change.
The test you tagged invokes `atlas iam project [subcommands]`, what I need is `atlas project [subcommands]`

test/e2e/iam/projects_test.go
jeroenvervaeke: There's no guarantee that the server will return the tags in the same order, so I don't think we can use this approach.

test/e2e/atlas/projects_test.go
gssbzn: > The test you tagged invokes atlas iam project [subcommands]

No it doesn't :)

docs/atlascli/command/atlas-projects-create.txt
sarahsimpers: Agreed. In the meantime, is there a way we can prevent this for now? Would `StringVar` not work in this scenario?

docs/atlascli/command/atlas-projects-create.txt
```diff
@@ -75,6 +75,10 @@ Options
...
+     - false
+     - List that contains key-value pairs between 1 to 255 characters in length for tagging and categorizing the project. This value defaults to [].
```
sarahsimpers, line 81: I haven't seen this null default before, should there be a default value here? Otherwise, shouldn't this line automatically be omitted?

test/e2e/iam/projects_test.go
jeroenvervaeke: Fixed

docs/atlascli/command/atlas-projects-create.txt
jeroenvervaeke: I've checked some other commands which use tags. They do have the sentence "This value defaults to [].", but the weird formatting makes it look like that line doesn't belong to the tag. Example:

>   * - --tag
>     - stringToString
>     - false
>     - List that contains key-value pairs between 1 to 255 characters in length for tagging and categorizing the cluster.
>
>       Mutually exclusive with --file. This value defaults to [].

https://github.com/mongodb/mongodb-atlas-cli/blob/master/docs/atlascli/command/atlas-clusters-create.txt#L132

In this command there's no `--file` flag to be mutually exclusive with, that's why it's on the same line.
I've checked some other commands which have the `--tag` flag and the behavior seems to be consistent, these are some I've checked:

- [clusters create](https://github.com/mongodb/mongodb-atlas-cli/blob/master/docs/atlascli/command/atlas-clusters-create.txt#L132)
- [clusters update](https://github.com/mongodb/mongodb-atlas-cli/blob/master/docs/atlascli/command/atlas-clusters-update.txt#L106)
- [clusters upgrade](https://github.com/mongodb/mongodb-atlas-cli/blob/master/docs/atlascli/command/atlas-clusters-upgrade.txt#L105)
- [deployments setup](https://github.com/mongodb/mongodb-atlas-cli/blob/master/docs/atlascli/command/atlas-deployments-setup.txt#L121), not mutually exclusive, same like this commands

Same with `stringToString`, it seems to be the same with other commands.

test/e2e/atlas/projects_test.go
jeroenvervaeke: Oh, my bad. In my change I've moved my changes to `projects_test.go` instead of `atlas_projects_test.go`

jeroenvervaeke: Moved the tests

docs/atlascli/command/atlas-projects-create.txt
jeroenvervaeke: `StringVar` will not work in this scenario since we accept multiple tags.

`atlas project create MyProject --tag environment=testing --tag prod=false` is a valid command.
With `StringVar` you would only be able to provide one tag.

docs/atlascli/command/atlas-projects-create.txt
jeroenvervaeke: Created a PR for the default value: https://github.com/mongodb-labs/cobra2snooty/pull/50

sarahsimpers (approved):
Copy LGTM with m2snooty changes as discussed

PR #2668, opened 2024-02-23
fix: use pagination defaults across commands

gssbzn:
Set the defaults to all pagination options to avoi

### DIFF
--- a/internal/cli/atlas/alerts/list.go
+++ b/internal/cli/atlas/alerts/list.go
@@ -91,8 +91,8 @@ func ListBuilder() *cobra.Command {
 			return opts.Run()
 		},
 	}
-	cmd.Flags().IntVar(&opts.PageNum, flag.Page, 0, usage.Page)
-	cmd.Flags().IntVar(&opts.ItemsPerPage, flag.Limit, 0, usage.Limit)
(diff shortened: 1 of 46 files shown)


internal/cli/atlas/alerts/list.go
```diff
@@ -91,8 +91,8 @@ func ListBuilder() *cobra.Command {
 			return opts.Run()
 		},
 	}
-	cmd.Flags().IntVar(&opts.PageNum, flag.Page, 0, usage.Page)
-	cmd.Flags().IntVar(&opts.ItemsPerPage, flag.Limit, 0, usage.Limit)
+	cmd.Flags().IntVar(&opts.PageNum, flag.Page, cli.DefaultPage, usage.Page)
+	cmd.Flags().IntVar(&opts.ItemsPerPage, flag.Limit, cli.DefaultPageLimit, usage.Limit)
```
tibulca, line 95: with default value 0, how was it working, were we returning all records or the backend was using it's default 1 and 100?

gssbzn: this was a regression introduced by the SDK migration, the original SDK would ignore zero values, and this change is actually no-op for mongocli since still old SDK

tibulca (approved):

PR #2669, opened 2024-02-23
CLOUDP-210224: More detailed message when deleting an instance

gssbzn:
## Proposed changes

_Jira ticket:_ CLOUDP-210224

Cluster
```bash
/bin/atlas clusters delete Cluster0 
? Confirm your backup settings before terminating your cluster.
Are you sure you want to terminate Cluster0, this action cannot be undone? (y/N) 
```

Serverless
```bash
./bin/atlas serverless delete Cluster0
? This operation will delete the instance, all of its data, and any associated backups.
Are you sure you want to terminate Cluster0, this action cannot be undone? (y/N) 
```

### DIFF
--- a/internal/cli/atlas/clusters/delete.go
+++ b/internal/cli/atlas/clusters/delete.go
@@ -104,7 +104,7 @@ Deleting a cluster also deletes any backup snapshots for that cluster.
 				return err
 			}
 			opts.Entry = args[0]
-			return opts.PromptWithMessage("Confirm your backup settings before terminating your cluster.\nAre you sure you want to terminate %s, this action cannot be undone?")
+			return opts.PromptWithMessage("Confirm your backup settings before terminating your cluster. This action cannot be undone.\nAre you sure you want to terminate %s?")
 		},
 		RunE: func(_ *cobra.Command, _ []string) error {
 			return opts.Run()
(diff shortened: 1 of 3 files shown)


colm-quinn (dismissed):
LGTM from eng side - let's get a docs review, but I think this really helps as a proposed copy

internal/cli/atlas/clusters/delete.go
```diff
@@ -104,7 +104,7 @@ Deleting a cluster also deletes any backup snapshots for that cluster.
 				return err
 			}
 			opts.Entry = args[0]
-			return opts.Prompt()
+			return opts.PromptWithMessage("Confirm your backup settings before terminating your cluster.\nAre you sure you want to terminate %s, this action cannot be undone?")
```
tibulca, line 107: can you please do the same for the `atlas deployments delete` command

gssbzn: do you have a suggestion for the message since this same copy doesn't apply for local

sarahsimpers (approved):
A few optional suggestions for you, thank you!

internal/cli/atlas/clusters/delete.go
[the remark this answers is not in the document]
gssbzn: we can't because it'll depend on the cluster settings, there's a setting to keep backup, I took both sentences from the GUI

PR #2673, opened 2024-02-26
CLOUDP-233923: fix orgs selection when profile is configured

tibulca:
## Proposed changes

fix orgs selection when profile is configured

_Jira ticket:_ CLOUDP-233923

### DIFF
--- a/internal/cli/default_setter_opts.go
+++ b/internal/cli/default_setter_opts.go
@@ -157,7 +157,7 @@ func (opts *DefaultSetterOpts) orgs(filter string) (results interface{}, err err
 		if r.GetTotalCount() > resultsLimit {
 			return nil, errTooManyResults
 		}
-		results = r.Results
+		results = *r.Results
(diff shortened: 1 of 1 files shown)


internal/cli/default_setter_opts.go
```diff
@@ -157,7 +157,7 @@ func (opts *DefaultSetterOpts) orgs(filter string) (results interface{}, err err
 		if r.GetTotalCount() > resultsLimit {
 			return nil, errTooManyResults
 		}
-		results = r.Results
+		results = *r.Results
```
tibulca, line 160: current function returns (results interface{}, err error) and the compiler was not catching that we were returning either an array or an array pointer.
With the mongocli fork, this issue will be properly solved (as we'll return a well-know type instead of `any`)

wtrocki (approved):

gssbzn (approved):

PR #2707, opened 2024-03-05
task: remove some ops manager only functions

gssbzn:
Delete some Ops Manager/Cloud manager only things related to custom http transport


fmenezes:
I've done exact same code locally together with so much more changes on the store

gssbzn:
This was kind of a quick win while I as there

PR #2726, opened 2024-03-07
test: reduce events flake

gssbzn:
Add omit count to the test so it doesn't error 

Started by adding the flag but this discovered we didn't have it for orgs so adding it


internal/cli/atlas/events/orgs_list.go
```diff
@@ -111,6 +111,7 @@ func OrgListBuilder() *cobra.Command {
 	cmd.Flags().StringSliceVar(&opts.EventType, flag.TypeFlag, nil, usage.Event)
 	cmd.Flags().StringVar(&opts.MaxDate, flag.MaxDate, "", usage.MaxDate)
 	cmd.Flags().StringVar(&opts.MinDate, flag.MinDate, "", usage.MinDate)
+	cmd.Flags().BoolVar(&opts.OmitCount, flag.OmitCount, false, usage.OmitCount)
```
andreaangiolillo, line 114: you need to update line 75 and include 
```
	if opts.OmitCount {
		p.IncludeCount = pointer.Get(false)
	}
```

PR #2734, opened 2024-03-07
task: unification of the stores part 3

gssbzn:
Moving now backup compliance and api access list to the general store


internal/cli/atlas/backup/compliancepolicy/copyprotection/disable_test.go
```diff
@@ -43,13 +42,6 @@ func TestDisableBuilder(t *testing.T) {
 	)
 }
 
-func TestDisableOpts_InitStore(t *testing.T) {
-	opts := &DisableOpts{}
-
-	require.NoError(t, opts.initStore(context.TODO())())
-	assert.NotNil(t, opts.store)
-}
```
gssbzn, line 51: we added a validation on the general store for the service and this now fails, deiced to remove since we never test this method

gssbzn: same for all removals of this test in backup compliance

fmenezes (approved):

PR #2743, opened 2024-03-11
task: delete streams from mongocli

gssbzn:
The streams feature is atlas only, deleting code


internal/mocks/mock_streams.go
```diff
@@ -1,390 +0,0 @@
-// Code generated by MockGen. DO NOT EDIT.
-// Source: github.com/mongodb/mongodb-atlas-cli/mongocli/v2/internal/store (interfaces: StreamsLister,StreamsDescriber,StreamsCreator,StreamsDeleter,StreamsUpdater,ConnectionCreator,ConnectionDeleter,ConnectionUpdater,StreamsConnectionDescriber,StreamsConnectionLister)
-
-// Package mocks is a generated GoMock package.
```
andreaangiolillo, line 4: could you check that the command is not added in the mongocli builder?

andreaangiolillo (approved):
LGTM

internal/mocks/mock_streams.go
gssbzn: if it was the compilation would fail :P the command was deleted but the store kept which I'm getting rid off here

PR #2745, opened 2024-03-11
task: reduce use of interfaces

gssbzn:
Reduce some of the interfaces on the store now that we don't support atlas on this branch


andreaangiolillo (approved):
LGTM. We waited years for this PR 🎉 🥳 🍾

PR #2767, opened 2024-03-13
CLOUDP-237474: AtlasCLI fails to support collection names with dots for custom roles

gssbzn:
CLOUDP-237474: properly handle db names with dots


internal/convert/custom_db_role.go
```diff
@@ -56,7 +56,7 @@ func BuildAtlasActions(a []string) []atlasv2.DatabasePrivilegeAction {
 			resource := strings.Split(action[1], resourceSep)
 			resourceStruct.Db = resource[0]
```
tibulca, line 57: this works if db names don't have dots, which is indeed a MongoDB restriction  https://www.mongodb.com/docs/manual/reference/limits/#naming-restrictions

I'm going to update the task and PR titles (support collection names with dots)

tibulca (approved):
LGTM

PR #2779, opened 2024-03-19
task: add go toolchain

gssbzn:
See https://github.com/golang/go/issues/62278


Makefile
```diff
@@ -28,6 +28,7 @@ export PATH := $(shell go env GOPATH)/bin:$(PATH)
 export PATH := ./bin:$(PATH)
 export TERM := linux-m
 export GO111MODULE := on
+export GOTOOLCHAIN := local
```
gssbzn, line 31: I found the "auto" behavior with asdf to be dangerous as it installs versions within the version manager so avoiding issue shere

.tool-versions
gssbzn, line 1: since go can be its won version manager let's forget of asdf

go.mod
```diff
@@ -2,6 +2,8 @@ module github.com/mongodb/mongodb-atlas-cli/atlascli
 
 go 1.22
```
blva, line 3: checked out to the branch and got:
- do we need to do any setup to be able to update master?
```
❯ make deps
==> Installing go.mod dependencies...
go mod download
go: go.mod requires go >= 1.22 (running go 1.21.1; GOTOOLCHAIN=local)
make: *** [deps] Error 1
```

gssbzn: see https://github.com/mongodb/mongodb-atlas-cli/pull/2779#discussion_r1530657881

you need to install/set go1.22 however you are managing your go versions, I have swapped from asdf to using the homebrew version (`asdf global golang system`) but you can keep using asdf if that works for you

blva, line 3:
```suggestion
go 1.22.1
```

gssbzn: >, I still get the error,

I can't reproduce before doing that change can you confirm you have updated your env correctly, as in reload the shell env, via something like `source ~/.zshrc`, or restarting your term, and if you still get an error what's the output of `go env`?

Please consider that 
```
go 1.22.1
```

should be equivalent to 
```
go 1.22
toolchain go1.22.1
```

blva: I think if it's not reproducible I can workaround to fix my machine, so I'll leave my stamp.

blva (approved):

PR #2787, opened 2024-03-19
task: go back to rhel80 for since we don't run podman tests anymore

gssbzn:
EVG is stuck on the 8.9 hosts and we can go back to 80 since we don't need podman


blva (approved):
LGTM

note: Can we add a note to whatever ticket we have tracking adding back those tests (if we ever do)

gssbzn:
>note: Can we add a note to whatever ticket we have tracking adding back those tests (if we ever do)

done

PR #2790, opened 2024-03-20
CLOUDP-237254: Set deploymet_type telemetry value in missing atlas deployments commands

blva:
## Proposed changes

_Jira ticket:_ [CLOUDP-237254](https://jira.mongodb.org/browse/CLOUDP-237254)

We were sending telemetry events on selection and some commands, this is an approach to try to unify and append deployment_type value at 3 steps:
1. pre-run (this covers if customer provides a flag but fails the command)
2. on selection (this covers if customer goes through interactive flow but fails the command)
3. post-run (this covers all cases where the command is successful)

I'll add a follow-up PR adding the telemetry to all commands, but want to get a review on this approach first.

## Checklist


internal/cli/atlas/deployments/list.go
```diff
@@ -71,6 +71,7 @@ func (opts *ListOpts) Run(ctx context.Context) error {
 }
 
 func (opts *ListOpts) PostRun() error {
+	opts.UpdateDeploymentTelemetry()
```
fmenezes, line 74: what is the expectation for list here?

blva: it will only set it if customer provides --type local/atlas, otherwise it will not set the type

internal/cli/atlas/deployments/delete_test.go
```diff
@@ -159,3 +159,27 @@ func TestDeleteBuilder(t *testing.T) {
...
+		},
+		DeleteOpts: cli.NewDeleteOpts(deleteSuccessMessage, deleteFailMessage),
+	}
+
+	deploymentsTest.MockDeploymentTelemetry.EXPECT().AppendDeploymentType().Times(1)
+
+	if err := opts.PostRun(); err != nil {
+		t.Fatalf("PostRun() unexpected error: %v", err)
+	}
+	t.Log(buf.String())
```
gssbzn, line 184: doesn't do much logging it if you are not comparing it with something
```suggestion
```

internal/cli/atlas/deployments/options/deployment_opts_telemetry.go
```diff
@@ -0,0 +1,36 @@
...
+import (
+	"github.com/mongodb/mongodb-atlas-cli/atlascli/internal/telemetry"
+)
+
+//go:generate mockgen -destination=../../../../mocks/mock_deployment_opts_telemetry.go -package=mocks github.com/mongodb/mongodb-atlas-cli/atlascli/internal/cli/atlas/deployments/options DeploymentTelemetry
+type DeploymentTelemetry interface {
+	AppendDeploymentType()
+}
+
+func NewDeploymentTypeTelemetry(opts *DeploymentOpts) DeploymentTelemetry {
```
gssbzn, line 26: [nit] in go the r[ecommendation is to accept interfaces and return structs](https://tutorialedge.net/golang/accept-interfaces-return-structs/), not the other way around

internal/cli/atlas/deployments/options/deployment_opts_telemetry.go
```diff
@@ -0,0 +1,36 @@
...
+	return opts
+}
+
+func (opts *DeploymentOpts) AppendDeploymentType() {
+	if opts.IsLocalDeploymentType() {
+		telemetry.AppendOption(telemetry.WithDeploymentType(LocalCluster))
+	} else if opts.IsAtlasDeploymentType() {
+		telemetry.AppendOption(telemetry.WithDeploymentType(AtlasCluster))
+	}
```
gssbzn, line 35: do the thing once
```suggestion
       var deploymentType string
	if opts.IsLocalDeploymentType() {
		deploymentType = LocalCluster
	} else if opts.IsAtlasDeploymentType() {
		deploymentType = AtlasCluster
	}
	if deploymentType != "" {
		telemetry.AppendOption(telemetry.WithDeploymentType())
	}
```

I would also say maybe is better to do a switch over `opts.DeploymentType`

internal/telemetry/event.go
```diff
@@ -164,7 +164,7 @@ func withAnonymousID() EventOpt {
 
 func WithDeploymentType(t string) EventOpt {
 	return func(event Event) {
-		event.Properties["deployment_type"] = t
+		event.Properties["deployment_type"] = strings.ToLower(t)
```
gssbzn, line 167: I would day the commands should normalize it so `opts.DeploymentType` is always the correct value, since I see we don't guarantee it

blva: good point, will update!

internal/cli/atlas/deployments/options/deployment_opts_telemetry.go
blva: we can do that it may look a bit weird for `DeploymentOpts` though as there we need to use `DeploymentType` we'd need to cast with something like this (which works fine):

```shell
func NewDeploymentTypeTelemetry(opts DeploymentTelemetry) *DeploymentOpts {
	return opts.(*DeploymentOpts)
}
```

internal/cli/atlas/deployments/options/deployment_opts_telemetry.go
blva: i prefer the suggestion you gave above as it uses the standard methods to check if it's atlas or local

internal/cli/atlas/deployments/options/deployment_opts_telemetry.go
gssbzn: it's a bit weird that you need to cast it to an interface if it accepts an interface but let's leave this for now

gssbzn: I think the problem may be this is not really a "new" which should return a new object but this is more of a "cast as interface" thing

gssbzn (approved):
LGTM

PR #2802, opened 2024-03-25
task: remove the atlas folder for commands

gssbzn:
deep nesting makes certain commands like `go fmt` or `go test` slower now that everything is atlas we can simplify some of our folder structure a little

Flatten the folder structure since everything is atlas now so no need for a `cli/atlas` folder to differentiate the from the old ops manager/cloud manager commands 

This change is only file moves and import fixes, no other changes


cmd/atlas/atlas.go
```diff
@@ -24,7 +24,7 @@ import (
 
 	"github.com/AlecAivazis/survey/v2"
 	"github.com/AlecAivazis/survey/v2/core"
-	"github.com/mongodb/mongodb-atlas-cli/atlascli/internal/cli/root/atlas"
+	"github.com/mongodb/mongodb-atlas-cli/atlascli/internal/cli/root"
```
blva, line 27: won't this affect `go:generate mockgen ` references as well? 

both destination=. and package reference, ok if we want to do a follow-up PR but mock generation may be broken meanwhile so may be worth to get next PR ready

gssbzn: >both destination=. and package reference, ok if we want to do a follow-up PR but mock generation may be broken meanwhile so may be worth to get next PR ready

No, I fixed the mocks and ran `make gen-mock` after my commit to confirm there are no changes as expected

blva (approved):

jeroenvervaeke (approved):

PR #2813, opened 2024-03-27
CLOUDP-239849: remove unused podman struct fields

tibulca:
## Proposed changes
remove unused podman struct fields

_Jira ticket:_ CLOUDP-239849


internal/podman/container_inspect.go
tibulca, line 1: [podman 5.0.0](https://github.com/containers/podman/releases/tag/v5.0.0) introduced a series of breaking changes which was failing the unmarshalling. In this PR I removed all unused fields (among which we also had the fields updated in 5.0.0) to reduce the risk of other breaking changes. 
For GA we should consider the version of podman when doing unmarshalling

gssbzn (approved):
LGTM if LGTEVG

PR #2865, opened 2024-04-16
CLOUDP-242974: Infer org id from project

josvazg:
## Proposed changes

- Org ID is optional for the Kubernetes config exported, used by **generate** and **apply**.
- The Org ID to be used when a project ID is given is always from the Org owner of such project.
- If the Org & Project IDs are passed in, the Org ID must match the owner of a project, otherwise an error will be returned:

`the project {PROJECT} ({PROJECT_ID}) is not part of the organization "{ORG_ID}".
Please confirm the arguments provided to the command or you are using the correct profile.`

_Jira ticket:_ CLOUDP-242974

## Checklist

- [X] I have signed the [MongoDB CLA](https://www.mongodb.com/legal/contributor-agreement)
- [X] I have added tests that prove my fix is effective or that my feature works
- [X] I have run `make fmt` and formatted my code

### DIFF
--- a/internal/kubernetes/operator/config_exporter.go
+++ b/internal/kubernetes/operator/config_exporter.go
@@ -26,7 +26,6 @@ import (
 	"github.com/mongodb/mongodb-atlas-cli/atlascli/internal/kubernetes/operator/features"
 	"github.com/mongodb/mongodb-atlas-cli/atlascli/internal/kubernetes/operator/project"
 	"github.com/mongodb/mongodb-atlas-cli/atlascli/internal/kubernetes/operator/resources"
-	"github.com/mongodb/mongodb-atlas-cli/atlascli/internal/log"
 	"github.com/mongodb/mongodb-atlas-cli/atlascli/internal/store"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/serializer/json"
@@ -63,7 +62,7 @@ var (
 	ErrNoCloudManagerClusters = errors.New("can not get 'advanced clusters' object")
 )
 
-func NewConfigExporter(dataProvider store.OperatorGenericStore, credsProvider store.CredentialsGetter, projectID, orgID string) *ConfigExporter {
+func NewConfigExporter(dataProvider store.OperatorGenericStore, credsProvider store.CredentialsGetter, projectID string) *ConfigExporter {
 	return &ConfigExporter{
 		dataProvider:            dataProvider,
 		credsProvider:           credsProvider,
@@ -72,7 +71,6 @@ func NewConfigExporter(dataProvider store.OperatorGenericStore, credsProvider st
 		dataFederationNames:     []string{},
 		targetNamespace:         "",
 		includeSecretsData:      false,
-		orgID:                   orgID,
 		dictionaryForAtlasNames: resources.AtlasNameToKubernetesName(),
 	}
 }
@@ -165,10 +163,7 @@ func (e *ConfigExporter) exportProject() ([]runtime.Object, string, error) {
 	if err != nil {
 		return nil, "", err
 	}
-	if atlasProject.OrgId != e.orgID {
-		_, _ = log.Warningf("Re-setting org ID to %q, owner of exported project\n", atlasProject.OrgId)
-		e.orgID = atlasProject.OrgId
-	}
+	e.orgID = atlasProject.OrgId
 
 	// Project
 	projectData, err := project.BuildAtlasProject(
--- a/internal/kubernetes/operator/config_exporter.go
+++ b/internal/kubernetes/operator/config_exporter.go
@@ -165,8 +165,9 @@ func (e *ConfigExporter) exportProject() ([]runtime.Object, string, error) {
 		return nil, "", err
 	}
 	if e.orgID != "" && e.orgID != atlasProject.OrgId {
-		return nil, "", fmt.Errorf("%q is not the owner of project %s, org id %s is",
-			e.orgID, atlasProject.GetId(), atlasProject.OrgId)
+		return nil, "", fmt.Errorf("the project %s (%s) is not part of the organization %q,"+
+			"please confirm the arguments provided to the command or you are using the correct profile",
+			atlasProject.GetName(), atlasProject.GetId(), e.orgID)
 	}
 	e.orgID = atlasProject.OrgId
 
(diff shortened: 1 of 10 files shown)


josvazg:
Not only this needs a test but also I believe this changes the semantics of the org id we pass in to the config exporter:
- The passed in org id is superfluous we could remove it already and even drop the CLI param as we can always use the org id of the project owner.
- But, we cannot do that for backward compatibility, we cannot change the CLI public interface suddenly.

Another way would be to mark the CLI org id param for the config exporter calls (generate and apply) as deprecated, saying it will have no effect at all and it should not be used going forward. Then place the org id on a separate place in the struct, maybe called `hintOrgId` or something like that, which is set but never used. Or better yet, it could be ignored completely even dropped from the exported config internal API.

The org id to be used will always be the owner of the project id that was passed in.

WDYT?
cc @s-urbaniak @gssbzn @helderjs @igor-karpukhin @roothorp

gssbzn:
>Another way would be to mark the org id param ... as deprecated

This is usually our recommended approach when we "remove" flags

josvazg:
> > Another way would be to mark the org id param ... as deprecated
> 
> This is usually our recommended approach when we "remove" flags

Anything special to mark it deprecated? or just say so on the param help CLI output for users to read?
Any example I should follow perhaps?

josvazg:
I see: `Flags().MarkDeprecated`

internal/kubernetes/operator/install.go
```diff
@@ -146,10 +146,6 @@ func (i *Install) ensureProject(orgID, projectName string) (*admin.Group, error)
 		return project, nil
 	}
 
-	if err != nil {
-		return nil, fmt.Errorf("failed to retrieve project: %w", err)
-	}
-
```
josvazg, line 152: Not really related by this If seems to no longer have sense here.

josvazg:
> @josvazg @s-urbaniak I wonder if we have any corner cases when we won't rely on the project. Like, for future AKO features like handling resources separately. For example, exporting federated auth doesn't require project ID but does require an org ID

In such case maybe it is probably preferable NOT to deprecate the org id but instead make sure it is ONLy taken into account when the project id is not given/required.

Still, for compatibility is is probably best NOT to fail if both are passed in, but instead just use the org id from the project.

WDYT @gssbzn ?

igor-karpukhin (approved):

internal/kubernetes/operator/project/project.go
```diff
@@ -70,12 +70,7 @@ type AtlasProjectResult struct {
 	Teams   []*akov2.AtlasTeam
 }
 
-func BuildAtlasProject(projectStore store.OperatorProjectStore, validator features.FeatureValidator, orgID, projectID, targetNamespace string, includeSecret bool, dictionary map[string]string, version string) (*AtlasProjectResult, error) { //nolint:gocyclo
-	project, err := projectStore.Project(projectID)
-	if err != nil {
-		return nil, err
-	}
-
+func BuildAtlasProject(projectStore store.OperatorProjectStore, project *atlasv2.Group, validator features.FeatureValidator, orgID, projectID, targetNamespace string, includeSecret bool, dictionary map[string]string, version string) (*AtlasProjectResult, error) { //nolint:gocyclo
```
helderjs, line 73: if you agree, it's a good candidate for a request struct instead of a lot of parameters

internal/kubernetes/operator/config_exporter.go
```diff
@@ -160,9 +160,20 @@ func (e *ConfigExporter) Run() (string, error) {
...
 func (e *ConfigExporter) exportProject() ([]runtime.Object, string, error) {
+	atlasProject, err := e.dataProvider.Project(e.projectID)
+	if err != nil {
+		return nil, "", err
+	}
+	if e.orgID != "" && e.orgID != atlasProject.OrgId {
+		return nil, "", fmt.Errorf("%q is not the owner of project %s, org id %s is",
+			e.orgID, atlasProject.GetId(), atlasProject.OrgId)
+	}
+	e.orgID = atlasProject.OrgId
```
helderjs, line 171: If the check above is to ensure orgID is not empty and equal to the value in the project, why still assign it here?

internal/kubernetes/operator/config_exporter.go
```diff
@@ -160,9 +160,20 @@ func (e *ConfigExporter) Run() (string, error) {
 }
 
 func (e *ConfigExporter) exportProject() ([]runtime.Object, string, error) {
+	atlasProject, err := e.dataProvider.Project(e.projectID)
+	if err != nil {
+		return nil, "", err
+	}
+	if e.orgID != "" && e.orgID != atlasProject.OrgId {
+		return nil, "", fmt.Errorf("%q is not the owner of project %s, org id %s is",
+			e.orgID, atlasProject.GetId(), atlasProject.OrgId)
```
gssbzn, line 169:
```suggestion
		return nil, "", fmt.Errorf(`the project %s (%s) is not part of the organization %q.
Please confirm the arguments provided to the command or you are using the correct profile.",
			atlasProject.GetName(), atlasProject.GetId(), e.orgID)
```

helderjs (approved):
LGTM, just some suggestions

test/e2e/atlas/kubernetes_config_generate_test.go
```diff
@@ -131,6 +131,28 @@ func InitialSetup(t *testing.T) KubernetesConfigGenerateProjectSuite {
 	return s
 }
 
+func TestProjectWithWrongOrgID(t *testing.T) {
+	s := InitialSetup(t)
+	cliPath := s.cliPath
+
+	t.Run("should fail to generate when the org id does not match the owner of the project", func(t *testing.T) {
```
gssbzn, line 138: this feels like it could be a unit and not an e2e whcih should be faster to run

internal/kubernetes/operator/config_exporter.go
josvazg: yes, this error message is much better, thanks!

internal/kubernetes/operator/project/project.go
josvazg: it might make sense, yes

test/e2e/atlas/kubernetes_config_generate_test.go
josvazg: Done

internal/kubernetes/operator/project/project.go
josvazg: Done

gssbzn (approved):
LGTM

PR #2912, opened 2024-05-01
task: enable perfsprint

gssbzn:
## Proposed changes

Enable perfsprint so we stop doing `fmt.Errorf` when we can do `errors.New`


internal/cli/kubernetes/config/apply.go
```diff
@@ -173,7 +173,7 @@ func ApplyBuilder() *cobra.Command {
 }
 
 func getOperatorMajorVersion(image string) string {
-	version := strings.Trim(image, fmt.Sprintf("%s:", containerImage))
+	version := strings.TrimPrefix(image, containerImage+":")
```
gssbzn, line 176: @mongodb/atlas_kubernetes_team this was an interesting one, removing fmt flagged an incorrect usage of Trim and that we should use TrimPrefix, can you confirm?

s-urbaniak: technically this behaves differently but indeed the `TrimPrefix` makes much more sense. LGTM from my side but leaving the review to have another set of eyes to @josvazg @helderjs @roothorp

helderjs (approved):
LGTM

internal/cli/kubernetes/config/apply.go
josvazg: yeah LGTM as well, we are only interesting on removing a prefix, that should not happen also at the end.

PR #2936, opened 2024-05-07
task: drop config file delegation

gssbzn:
Remove the method delegation since this was only used when there was an `if` depending on the binary, so we can remove now that mongocli is forked

### DIFF
--- a/cmd/atlas/atlas.go
+++ b/cmd/atlas/atlas.go
@@ -83,7 +83,7 @@ func shouldCopyConfig(atlasConfigPath string) bool {
 
 // createConfigFromMongoCLIConfig creates the atlasCLI config file from the mongocli config file.
 func createConfigFromMongoCLIConfig() {
-	atlasConfigHomePath, err := config.AtlasCLIConfigHome()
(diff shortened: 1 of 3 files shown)


cmd/atlas/atlas.go
```diff
@@ -145,7 +145,7 @@ func mongoCLIConfigFilePath() (configPath string, err error) {
 	}
 
 	if configDir, err := config.OldMongoCLIConfigHome(); err == nil { //nolint:staticcheck // Deprecated before fully removing support in the future
-		configPath = fmt.Sprintf("%s/mongocli.toml", configDir)
+		configPath = path.Join(configDir, "mongocli.toml")
 	}
```
andreaangiolillo, line 149: can't we remove this piece of code?

andreaangiolillo (approved):

cmd/atlas/atlas.go
gssbzn: I though about it but we may want to confirm with product I'm going to create a ticket

gssbzn: https://jira.mongodb.org/browse/CLOUDP-248054

PR #2939, opened 2024-05-07
task: split kube test into two tasks

gssbzn:
Try to improve kube tests times by splitting into two tasks

Before 1 task ~30 min
Now 2 tasks in parallel where the slowest takes ~18min


gssbzn:
evergreen refresh

s-urbaniak (approved):

PR #2968, opened 2024-05-22
CLOUD-250122: [CLI] NPE when org list is empty

andreaangiolillo:
## Proposed changes

_Jira ticket:_ CLOUDP-250122

A user of MongoDB University is getting the following error when using `org ls` without an organization. 
```bash
Error: template: output:1:33: executing "output" at <.Results>: nil pointer evaluating *admin.PaginatedOrganization.Results
```

I wasn't able to reproduce the error with a user without an org. This PR updates the unit test to include additional scenarios to include the endpoint response without results.


internal/cli/organizations/list_test.go
```diff
@@ -22,28 +22,83 @@ import (
...
+					},
+				},
+			},
+			returnErr: nil,
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			ctrl := gomock.NewController(t)
```
gssbzn, line 78: can you parallelize this?
```suggestion
			t.Parallel()
			ctrl := gomock.NewController(t)
```

internal/cli/organizations/list_test.go
```diff
@@ -22,28 +22,83 @@ import (
...
+		name      string
+		expected  *atlasv2.PaginatedOrganization
+		returnErr error
+	}{
+		{
+			name: "non-nil result",
+			expected: &atlasv2.PaginatedOrganization{
+				Results: &[]atlasv2.AtlasOrganization{{}, {}, {}},
+			},
+			returnErr: nil,
```
gssbzn, line 41: this is always nil, do you need it as a prop?

gssbzn (approved):
LGTM

PR #2988, opened 2024-06-04
task: last references to atlas go client

gssbzn:
Last reference in ops manager code to the atlas go client

### DIFF
--- a/internal/cli/mongocli/alerts/settings/list_test.go
+++ b/internal/cli/mongocli/alerts/settings/list_test.go
@@ -25,7 +25,6 @@ import (
 	"github.com/mongodb/mongodb-atlas-cli/mongocli/v2/internal/mocks"
 	"github.com/mongodb/mongodb-atlas-cli/mongocli/v2/internal/pointer"
 	"github.com/stretchr/testify/assert"
-	"go.mongodb.org/atlas/mongodbatlas"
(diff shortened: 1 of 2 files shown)


internal/cli/mongocli/alerts/settings/list_test.go
```diff
@@ -25,7 +25,6 @@ import (
 	"github.com/mongodb/mongodb-atlas-cli/mongocli/v2/internal/mocks"
 	"github.com/mongodb/mongodb-atlas-cli/mongocli/v2/internal/pointer"
 	"github.com/stretchr/testify/assert"
-	"go.mongodb.org/atlas/mongodbatlas"
```
lantoli, line 28: does it help to do `go mod tidy` if not used anymore?

lantoli (approved):

blva (approved):

internal/cli/mongocli/alerts/settings/list_test.go
gssbzn: we have a test check to make sure is needed but in this case mocks for some reason still use the atlas version (ops manager defines an alias) and `mongocli auth login` still needs this lib, I think the new team that will take over mongocli will be cleanign some of that and why this PR

husniMDB (approved):

PR #2993, opened 2024-06-07
CLOUDP-250877: E2E tests fail when error level is set to debug

jeroenvervaeke:
## Proposed changes
_Jira ticket:_ CLOUDP-250877

Fixed e2e tests when you change the default log level from `WarningLevel` to `DebugLevel` on line 109:
https://github.com/mongodb/mongodb-atlas-cli/blob/3ba0033fe8392dbdb692f31a78ba40169f907d47/internal/log/log.go#L109

What I've changed:
I've added a helper method called `e2e.RunAndGetStdOut` which does exactly the same as `cmd.CombinedOutput()` (copy pasted most of the code), except that it writes `stderr` output to `stderr` and only captures `stdout`.

I've tested this here:
https://spruce.mongodb.com/version/66630cc3c73c360007999b6f/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC

As you can see some tests under kubernetes and backup fail.
These e2e tests also fail in our master branch: https://spruce.mongodb.com/version/6662dd9b46b7c80007caf13c/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC
I've opened a slack threat in the apix2 channel to follow up on the failing e2e tests on master.

### DIFF
--- a/test/e2e/atlas/access_lists_test.go
+++ b/test/e2e/atlas/access_lists_test.go
@@ -53,7 +53,7 @@ func TestAccessList(t *testing.T) {
 			g.projectID,
 			"-o=json")
 		cmd.Env = os.Environ()
-		resp, err := cmd.CombinedOutput()
+		resp, err := e2e.RunAndGetStdOut(cmd)
 		require.NoError(t, err)
 
 		var entries *atlasv2.PaginatedNetworkAccess
@@ -78,7 +78,7 @@ func TestAccessList(t *testing.T) {
 			g.projectID,
 			"-o=json")
 		cmd.Env = os.Environ()
-		resp, err := cmd.CombinedOutput()
+		resp, err := e2e.RunAndGetStdOut(cmd)
 		require.NoError(t, err, string(resp))
 		var entries *atlasv2.PaginatedNetworkAccess
 		require.NoError(t, json.Unmarshal(resp, &entries))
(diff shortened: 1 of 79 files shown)


test/e2e/atlas/access_lists_test.go
```diff
@@ -53,7 +53,7 @@ func TestAccessList(t *testing.T) {
 			g.projectID,
 			"-o=json")
 		cmd.Env = os.Environ()
-		resp, err := cmd.CombinedOutput()
+		resp, err := e2e.RunAndGetStdOut(cmd)
```
andreaangiolillo, line 56: Should we check that the `os.Stderr` is empty in all the tests? I am wondering if checking for `require.NoError(t, err)` (next line) may not be enough since the command errors will be shown only in `Stderr`

andreaangiolillo (approved):
Left a comment to make sure tests will still fail since we are not checking stderr. LGTM once the comment is resolved. Thanks!

test/e2e/atlas/access_lists_test.go
jeroenvervaeke: `os.Stderr` is used to print our logs; it doesn't mean an error has occurred.

Some places do use `os.Stderr` to check if some log has been printed; I've checked those places manually and made sure that the test still works.

PR #2994, opened 2024-06-10
CLOUDP-253083: Refactored setup command

jeroenvervaeke:
## Proposed changes

_Jira ticket:_ CLOUDP-253083

Refactored setup command to use the new unified image:
- don't create a network anymore
- remove references to MongoT
- simplify overall setup, most things are done in the unified docker image

Also, fixed collateral damage in other commands, which were impacted by removing references to MongoT and removing some networking functionality. Other commands will be fully refactored in separate PRs.

## Testing
Ran `atlas deploy setup` and connected with compass
- With default settings and no authentication
- With custom settings and no authentication
- With default settings and authentication


internal/cli/.DS_Store
blva, line 1: oops! can we remove this

internal/cli/deployments/setup.go
```diff
@@ -16,21 +16,17 @@ package deployments
 
 import (
 	"context"
-	_ "embed"
```
blva, line 19: [q]I remember at some point we removed the local dev tests/marked them to spot running as they're flaky. I strongly recommend we add them back sooner rather than later so that we can see easily how our changes may affect the functionality, do we have a ticket for that?

internal/cli/deployments/options/deployment_opts_remove.go
```diff
@@ -31,10 +31,6 @@ func (opts *DeploymentOpts) RemoveLocal(ctx context.Context) error {
 		return errRemove
 	}
 
-	if _, errRemove := opts.PodmanClient.RemoveNetworks(ctx, opts.LocalNetworkName()); errRemove != nil {
```
blva, line 34: as we clean up things, I left a comment in the [scope](https://docs.google.com/document/d/1c5xU5trWBMf7H0IOcPMi_Fj2gyl_JnJpe2QAOzrcIOY/edit?disco=AAABLbO9g8s).

- will these changes be released in to a major version?
- as customers move to a new version of the CLI, how do we make sure the leftover instances from podman are removed? 
- is it OK to add this code here and perform a release this week?

internal/cli/.DS_Store
jeroenvervaeke: done

internal/cli/deployments/setup.go
jeroenvervaeke: Do you know which tests so I can create a ticket?

PR #3014, opened 2024-06-13
CLOUDP-253925: labeler action is not sending the notification

andreaangiolillo:
## Proposed changes

_Jira ticket:_ CLOUDP-253925

## BUG
`github.event.pull_request.labels.*.name` doesn't sometimes contain the PR labels if the label was added in the same action. When the action runs the second time it always contains the labels

## Fix
This PR calls the GH API to retrieve the label added to the PR. I tested the changes in my [fork](https://github.com/andreaangiolillo/mongocli-test/pull/65) but I will need to create a PR in mongocli after this is merged to make sure it works as expected.


blva (approved):
LGTM, thanks for fixing this!!!

PR #3020, opened 2024-06-14
CLOUDP-253914: Remove volumes

jeroenvervaeke:
## Proposed changes

_Jira ticket:_ CLOUDP-253914#

Remove all volumes in setup.

Verified that everything still works (reasons why volumes were introduced):
- Close laptop lid
- Let laptop go in standby by not using it
- Restart laptop (we could also force push off)

After testing, it seems like everything seems to work perfectly in these situations.


fmenezes:
LGTM

PR #3021, opened 2024-06-17
CLOUDP-252933: Add telemetry indicator for university users

andreaangiolillo:
## Proposed changes
_Jira ticket:_ CLOUDP-252933

### Context:
Analytics need to be able to separate university users from real users in the CLI telemetry data.

### Proposed Solution
Add a `CLI_USER_TYPE ` environment variable for AtlasCLI. When this variable is set to `university`, AtlasCLI will set the property`cli_user_type` to `university` for telemetry. If not set, `cli_user_type` will be set to `default.

# Testing
I tested the changes by running a local instance of Atlas and checked that the telemetry endpoint received the `cli_user_type` property with and without `CLI_USER_TYPE `. I also checked segment and the new property was there for cloud-dev

- the env CLI_USER_TYPE was defined
Screenshot 2024-06-19 at 15 43 11


internal/config/profile.go
```diff
@@ -65,6 +65,9 @@ const (
 	ContainerizedHostNameEnv = "MONGODB_ATLAS_IS_CONTAINERIZED"
 	GitHubActionsHostNameEnv = "GITHUB_ACTIONS"
 	AtlasActionHostNameEnv   = "ATLAS_GITHUB_ACTION"
+	MongoDBUniversityUserEnv = "MONGODB_UNIVERSITY_USER" // MongoDBUniversityUserEnv is used to separate MongoDB University users from default users
```
fmenezes, line 68: who sets this var?

fmenezes: are we inline with university folks about this?

andreaangiolillo: > who sets this var?

The MongoDB university team will set this var when running the CLI in their environment

> are we inline with university folks about this?

The Jira ticket doesn't provide context to answer this question. I was planning to reach out once the PR is merged so that they are aware that this is coming and to set the env var.

andreaangiolillo: I tagged the university team in the atlascli slack channel for awareness.

braz: Thanks @andreaangiolillo and @fmenezes - delighted to see this change and happy to ensure it's rolled out to all of labs.

internal/telemetry/event.go
```diff
@@ -259,6 +259,16 @@ func withProjectID(cmd CmdFlags, c ProjectIDGetter) EventOpt {
 	}
 }
 
+type SourceGetter interface {
```
fmenezes, line 262: do we need this SourceGetter?

internal/telemetry/event.go
```diff
@@ -259,6 +259,16 @@ func withProjectID(cmd CmdFlags, c ProjectIDGetter) EventOpt {
...
+	Source() string
+}
+
+func withUserType() EventOpt {
+	return func(event Event) {
+		event.Properties["user_type"] = config.UserType
```
fmenezes, line 268: the pattern we have for env vars in the same as `withCI` where we check if the env var is present and set the property to true, I don't think we need the concept of `user_type`

internal/telemetry/event.go
andreaangiolillo: leftover. removed. Thanks!

internal/telemetry/event.go
andreaangiolillo: Are you suggesting setting the CI property of telemetry to `university`?

tibulca: I think that regardless of the env var we ask mdb university to use, the telemetry property should be an enum, so we can add more values in the future. So either `cli_source` or `user_type`.
For the university we have 2 options:
1. MONGODB_UNIVERSITY_USER=true env var (and translate to cli_source or user_type = university)
2. CLI_SOURCE or USER_TYPE=university and just pass to the telemetry as a property with the same name
@fmenezes was your suggestion more in line with second option based on current telemetry implementation?

fmenezes: I'm suggesting to use the env var key as the property

For instance mongodb-university-user=true/false

andreaangiolillo: mh.. This solution will force us to add a new telemetry property instead of updating just the enum if we find another user type in the future. @tibulca WDYT?

fmenezes: @andreaangiolillo are you expecting any other user types?

fmenezes: I _personally_ think user_type is quite confusing especially because it does not make me remember the original env var.

I feel telemetry should be as close as possible to the running environment.

If our env var is called `MONGODB_UNIVERSITY_USER` _I think_ we should forward _exactly that_ to telemetry.

andreaangiolillo: understood. I am wondering if renaming `MONGODB_UNIVERSITY_USER_TYPE` remove the confusion 🤔 I don't have a strong opinion here. I am happy to add `MONGODB_UNIVERSITY_USER` as a telemetry prop. I will wait for Ciprian to confirm and then I will address the comment. Thanks for the feedback! 🙇

tibulca: If it's keeping the same name for the env var and then option 2 works, right?
```
CLI_SOURCE or USER_TYPE=university and just pass to the telemetry as a property with the same name
@fmenezes was your suggestion more in line with second option based on current telemetry implementation?
```
There's no other source currently (as this is the first use case), but I don't think we should exclude that for the future and end up with another MONGODB_SOME_OTHER_TYPE and so on

fmenezes: @tibulca it works for me

fmenezes: in telemetry there is a source property currently at https://github.com/mongodb/mongodb-atlas-cli/blob/594c4fe2ddfc1f0586a5d8d007076c76efe053fd/internal/telemetry/event.go#L332

fmenezes: I wouldn't touch it as it would require mms code changes

andreaangiolillo: Decision:
Use `CLI_USER_TYPE =university` and add a new telemetry prop with the same name as the env variable

andreaangiolillo: I updated the PR with the new env variable and telemetry property.

> the telemetry property should be an enum

Regarding this comment, golang doesn't support enums. I don't think that simulating an enum with iota here would improve the usability of the code. WDYT?

internal/config/profile.go
```diff
@@ -196,6 +200,15 @@ func getConfigHostnameFromEnvs() string {
 	return configHostName
 }
 
+// newCLIUserTypeFromEnvs patches the user type information based on set env vars.
+func newCLIUserTypeFromEnvs() string {
+	if os.Getenv(CLIUserTypeEnv) == UniversityUser {
+		return UniversityUser
+	}
+
+	return DefaultUser
```
tibulca, line 209: to avoid adding another check when other user types are added, wdyt about:
```suggestion        
	if value, ok := os.LookupEnv(CLIUserTypeEnv); ok {
		return value
	}

	return DefaultUser
```

andreaangiolillo: oh nice 💯

andreaangiolillo: accepted! 👍  xD

tibulca (approved):

PR #3036, opened 2024-06-18
CLOUDP-136708: AtlasCLI: Support Get/Update Audit configs for projects

andreaangiolillo:
## Proposed changes

_Jira ticket:_ CLOUDP-136708


andreaangiolillo:
```bash
./bin/atlas auditing update --help                                                      4s  16.18.0
To use this command, you must authenticate with a user account or an API key with the Project Owner role.

Usage:
  atlas auditing update [flags]

Examples:
  # Audit all authentication events for known users:
  atlas auditing update --auditFilter "{"atype": "authenticate"}"

  # Audit all authentication events for known user via a configuration file:
  atlas auditing update -f filter.json


Flags:
      --auditAuthorizationSuccess   Flag that indicates whether someone set auditing to track successful authentications. This only applies to the "atype" : "authCheck" audit filter. Setting this parameter to true degrades cluster performance.
      --auditFilter string          JSON document that specifies which events to record. Escape any characters that may prevent parsing, such as single or double quotes, using a backslash (\).
      --enabled                     Flag that indicates whether someone enabled database auditing for the specified project.
  -f, --file string                 Path to an optional JSON configuration file that defines auditing filters. To learn more about auditing configuration files for the Atlas CLI, see https://www.mongodb.com/docs/atlas/database-auditing/#example-auditing-filters
  -h, --help                        help for update
  -o, --output string               Output format. Valid values are json, json-path, go-template, or go-template-file. To see the full output, use the -o json option.
      --projectId string            Hexadecimal string that identifies the project to use. This option overrides the settings in the configuration file or environment variable.

Global Flags:
  -P, --profile string   Name of the profile to use from your configuration file. To learn about profiles for the Atlas CLI, see https://dochub.mongodb.org/core/atlas-cli-save-connection-settings.
```


```bash
./bin/atlas auditing update --auditFilter "{"atype": "authenticate"}" --enabled -o json  
{
  "auditAuthorizationSuccess": false,
  "auditFilter": "{atype: authenticate}",
  "configurationType": "FILTER_JSON",
  "enabled": true
}
```

erabil-mdb (approved):
LGTM!

internal/cli/auditing/update.go
```diff
@@ -0,0 +1,126 @@
...
+	cmd := &cobra.Command{
+		Use:   "update",
+		Short: "Updates the auditing configuration for the specified project",
+		Long:  fmt.Sprintf(usage.RequiredRole, "Project Owner"),
+		Args:  require.NoArgs,
+		Annotations: map[string]string{
+			"output": updateTemplate,
+		},
+		Example: `  # Audit all authentication events for known users:
+  atlas auditing update --auditFilter "{"atype": "authenticate"}"
```
tibulca, line 96: does this work or the quotes have to be escaped?

internal/cli/auditing/update.go
```diff
@@ -0,0 +1,126 @@
...
+	cmd.Flags().StringVar(&opts.auditFilter, flag.AuditFilter, "", usage.AuditFilter)
+	cmd.Flags().BoolVar(&opts.enabled, flag.Enabled, false, usage.EnabledAuditing)
+	cmd.Flags().StringVarP(&opts.filename, flag.File, flag.FileShort, "", usage.AuditingFilename)
+	cmd.Flags().BoolVar(&opts.auditAuthorizationSuccess, flag.AuditAuthorizationSuccess, false, usage.AuditAuthorizationSuccess)
+	cmd.Flags().StringVarP(&opts.Output, flag.Output, flag.OutputShort, "", usage.FormatOut)
+	_ = cmd.RegisterFlagCompletionFunc(flag.Output, opts.AutoCompleteOutputFlag())
+
+	_ = cmd.MarkFlagFilename(flag.File)
+
+	cmd.MarkFlagsMutuallyExclusive(flag.File, flag.AuditFilter)
```
tibulca, line 123: the flags are mutually exclusive but I don't see an `opts.validateFlags` to check if one of the 2 is set

internal/cli/auditing/update.go
andreaangiolillo: during my testing, I didn't have to escape the quotes

internal/cli/auditing/update.go
andreaangiolillo: good point 👍

internal/cli/auditing/update.go
tibulca: in terminal, without escaping `"{"atype": "authenticate"}"` ends in the CLI as `{atype: authenticate}` which is not a valid JSON.
try for example `echo "{"atype": "authenticate"}"`

Also, this format (using single quote) works too: `atlas auditing update --auditFilter '{"atype": "authenticate"}'`

internal/cli/auditing/update.go
andreaangiolillo: updated

internal/cli/auditing/update.go
andreaangiolillo: Here is an example. The API response had the quote 🤔 . I will update anyway the example to include `'`
```bash
./bin/atlas auditing update --auditFilter "{"atype": "authenticate"}" --enabled -o json  
{
  "auditAuthorizationSuccess": false,
  "auditFilter": "{atype: authenticate}",
  "configurationType": "FILTER_JSON",
  "enabled": true
}
```

andreaangiolillo: nvm mi comment I see the answer of the API was wrong indeed 🤦 I am wondering why I didn't get an error

andreaangiolillo: here is the reply with `'`. Valid JSON indeed 👍 

./bin/atlas auditing update  --auditFilter '{"atype": "authenticate"}' --enabled -o json
{
  "auditAuthorizationSuccess": false,
  "auditFilter": "{\"atype\": \"authenticate\"}",
  "configurationType": "FILTER_JSON",
  "enabled": true
}

andreaangiolillo: updated

tibulca (approved):

PR #3040, opened 2024-06-21
CLOUDP-256305: SETUP: Wait for container to become healthy

jeroenvervaeke:
## Proposed changes
_Jira ticket:_ CLOUDP-256305

- Fix the issue where podman ignores the healthcheck defined in our unified docker image
- Wait for the healthcheck to be ready when running the setup command (this is the same behavior like we have in the master branch right now)

### DIFF
--- a/internal/container/podman.go
+++ b/internal/container/podman.go
@@ -16,6 +16,7 @@ package container
 
 import (
 	"context"
+	"fmt"
 	"strings"
 
 	"github.com/mongodb/mongodb-atlas-cli/atlascli/internal/podman"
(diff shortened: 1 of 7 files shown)


internal/container/podman.go
```diff
@@ -77,6 +79,21 @@ func (e *podmanImpl) ContainerRun(ctx context.Context, image string, flags *RunF
 				podmanOpts.Ports[entry.HostPort] = entry.ContainerPort
 			}
 		}
+		if flags.HealthCmd != nil {
```
fmenezes, line 82: do we need to do this for docker as well?

jeroenvervaeke: I wanted to avoid writing something among the lines of `if podman`. Since our abstractions are `startContainer` and not `startDeployment`, I couldn't move this code to the postman abstraction.

This code will also run for docker, but docker will find the healthcheck, so nothing will happen (I'm implementing the docker code right now, that's why I put the PR back in draft)

fmenezes (approved):

PR #3063, opened 2024-06-28
chore: update tj actions

blva:
## Proposed changes

Trying to update this to resolve https://github.com/mongodb/mongodb-atlas-cli/security/dependabot

## Checklist

- [x] I have signed the [MongoDB CLA](https://www.mongodb.com/legal/contributor-agreement)
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] I have added any necessary documentation in document requirements section listed in [CONTRIBUTING.md](https://github.com/mongodb/mongodb-atlas-cli/blob/master/CONTRIBUTING.md) (if appropriate)
- [ ] I have addressed the @mongodb/docs-cloud-team comments (if appropriate)
- [ ] I have updated [test/README.md](https://github.com/mongodb/mongodb-atlas-cli/blob/master/test/README.md) (if an e2e test has been added)
- [x] I have run `make fmt` and formatted my code

PR #3078, opened 2024-07-04
chore: update code_health third parties action to their shas

andreaangiolillo:
## Proposed changes
Update the code health third parties action to their sha


gssbzn (changes requested):
not the same binary but the same rules, does the binary matter?

gssbzn:
If you see the code they use it as a library, https://github.com/golangci/golangci-lint/blob/8f348db7bbd6366c528a369ad0b02505afd265c8/pkg/golinters/staticcheck/staticcheck.go#L4C22-L4C33 the note is just because of beef between the two projects 

https://github.com/golangci/golangci-lint/pull/2894

andreaangiolillo:
ah okay, I was tricked by `It's not the same thing as the staticcheck binary.`. I will reuse this PR to update the action to ping third parties to sha then

.github/workflows/code-health.yml
```diff
@@ -10,17 +10,17 @@ jobs:
     name: lint
     runs-on: ubuntu-latest
     steps:
-      - uses: GitHubSecurityLab/actions-permissions/monitor@v1
+      - uses: GitHubSecurityLab/actions-permissions/monitor@e83639073897b68de235d81a783b2221cc13def7
         with:
           config: ${{ vars.PERMISSIONS_CONFIG }}
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332
```
gssbzn, line 16: my understanding was that actions supported by GH could remain on numbers and not shas, has this changed?

gssbzn: using shas is actually really bad for security scanning tools as you may be able to  see from https://github.com/mongodb/mongodb-atlas-cli/pull/3063

andreaangiolillo: I was unaware of this, could you share where it was decided?

andreaangiolillo: no problem if you don't find it. It makes sense that GH actions can be trusted

PR #3081, opened 2024-07-08
CLOUDP-237038: Enable e2e_local_deployments variant

jeroenvervaeke:
## Proposed changes
Enable e2e_local_deployments variant

_Jira ticket:_ CLOUDP-237038

Fix flaky e2e tests:
- Podman does not always run healthchecks
- Podman healthchecks stay in the "starting" after the container exited
- Docker impl wasn't passing environment variables properly
- Docker impl wasn't parsing port map properly when ipv6 is enabled
- Mark every linux as supported
- Fix diagnostics when container is not running
- Added additional debug logging


internal/podman/client.go
```diff
@@ -383,6 +394,29 @@ func (o *client) ContainerLogs(ctx context.Context, name string) ([]string, erro
 	return logs, nil
 }
 
+func (o *client) ContainerUptime(ctx context.Context, name string) (time.Duration, error) {
```
fmenezes, line 397: Instead of creating these functions can we reuse ContainerInspect?

internal/cli/deployments/setup.go
```diff
@@ -242,7 +241,8 @@ func (opts *SetupOpts) configureMongod(ctx context.Context) error {
 		return err
 	}
 
-	const healthyDeploymentTimeout = 2 * time.Minute
+	// This can be a high number because the container will become unhealthy before the 10 minutes is reached
+	const healthyDeploymentTimeout = 10 * time.Minute
```
fmenezes, line 245: this seems very high, is it for both podman and docker?

internal/podman/client.go
jeroenvervaeke: We could, but we don't need it for docker. I wanted to avoid having to implement things in multiple places if we don't use them in both.

internal/cli/deployments/setup.go
jeroenvervaeke: Yes, but we will never reach this timeout because the timeout in our container is set to 2 minutes.
The only case where we will reach this timeout is when podman/docker doesn't run the healthcheck, which should never happen.

internal/podman/client.go
fmenezes: Well container inspect is implemented for podman and it does not impact docker, check https://github.com/mongodb/mongodb-atlas-cli/pull/3081/files#diff-b2a128f7fb3a56f3cca07064286b8b276642d57d2b75dfa8b1924448b1888a15R152 it is just a matter of adding more fields to the podman only internal struct

internal/cli/deployments/setup.go
fmenezes: how about setting this to 3 minutes?

jeroenvervaeke: sure, will update

internal/podman/client.go
jeroenvervaeke: Container inspects only deserializes the JSON into a struct. But doesn't do any parsing.

If I add the uptime to the inspect struct, I would have to do one of the following:
- Parse the created at the level of the caller
- Change the current logic of the `ContainerInspect` function and use two separate structs, one for the raw JSON we get back and one for the parsed JSON, and do the conversion from a string to a DateTime in the inspect method.

Both options are worse than what we have now in my opinion.

fmenezes: you can do `o.runPodman(ctx, "inspect", "--format", "[\"{{.State.Status}}\", \"{{.State.StartedAt}}\"]", name)` for instance

jeroenvervaeke: updated

fmenezes (approved):

PR #3098, opened 2024-07-11
CLOUDP-259871: Plugin discovery logic and plugin list command

stefan4h:
## Proposed changes

- Plugin discovery logic
- Add `plugin` sub command
- Add `plugin list` sub command

_Jira ticket:_ CLOUDP-259871


tools/cli-generator/generate.go
```diff
@@ -237,7 +237,8 @@ func cleanupFile(filePath string) error {
 	if err != nil {
 		return err
 	}
-	return os.WriteFile(filePath, r, os.ModePerm)
+	const filePerm = 0600
```
jeroenvervaeke, line 240: Do we need this change as part of this PR?

jeroenvervaeke (changes requested):

tools/cli-generator/generate.go
jeroenvervaeke: I think you also changed this here: #3101

internal/plugin/plugin_manifest.go
```diff
@@ -0,0 +1,216 @@
...
+	}
+	if p.Description == "" {
+		errorsList = append(errorsList, fmt.Errorf(errorMessage, "description"))
+	}
+	if p.Binary == "" {
+		errorsList = append(errorsList, fmt.Errorf(errorMessage, "binary"))
+	}
+	if p.Version == "" {
+		errorsList = append(errorsList, fmt.Errorf(errorMessage, "version"))
+	} else if valid, _ := regexp.MatchString(`^\d+\.\d+\.\d+$`, p.Version); !valid {
```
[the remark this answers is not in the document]
stefan4h: FYI we are using [this semver package](https://pkg.go.dev/github.com/Masterminds/semver/v3@v3.2.1) and not the one you posted so the validation looks slightly different:

```go
	} else if _, err := semver.NewVersion(p.Version); err != nil {
```

gssbzn:
bad merge?

sarahsimpers (approved):
LGTM

tools/cli-generator/generate.go
stefan4h: Yes I did, removed it from this PR now

jeroenvervaeke (approved):
LGTM

PR #3099, opened 2024-07-11
CLOUDP-200390: [ML3] atlas deployments setup unit test

jeroenvervaeke:
## Proposed changes
- Added unit tests for `atlas deployment setup`
- Fixed an issue I've discovered while implementing the unit tests. When a deployment already exists we print "deployment already exists" and then delete the deployment we found and exit.

_Jira ticket:_ CLOUDP-200390

### DIFF
--- a/internal/cli/deployments/setup.go
+++ b/internal/cli/deployments/setup.go
@@ -59,6 +59,7 @@ const (
 
 var (
 	errCancel                   = errors.New("the setup was cancelled")
+	ErrDeploymentExists         = errors.New("deployment already exists")
 	errMustBeInt                = errors.New("you must specify an integer")
 	errPortOutOfRange           = errors.New("you must specify a port within the range 1..65535")
 	errPortNotAvailable         = errors.New("the port is unavailable")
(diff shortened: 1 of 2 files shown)


fmenezes (approved):

internal/cli/deployments/setup.go
```diff
@@ -598,7 +598,10 @@ func (opts *SetupOpts) runLocal(ctx context.Context) error {
 	}
 
 	if err := opts.createLocalDeployment(ctx); err != nil {
-		_ = opts.RemoveLocal(ctx)
+		// in case the deployment already exists we shouldn't delete it
+		if !strings.Contains(err.Error(), "deployment already exists and is currently in") {
```
fmenezes, line 602: I wonder if we there is a better error checking approach? like `errors.Is`?

jeroenvervaeke: I was wondering the same thing. But the error contains some dynamic parts, hence the contains.

jeroenvervaeke: Updated using your suggestion

tibulca: was late with my comment :)

tibulca (approved):

PR #3102, opened 2024-07-12
CLOUDP-261264: remove podman as a recommondation on debian based distros

jeroenvervaeke:
## Proposed changes
- Make docker a debendency on debian based systems.
- Add docker to the suggestions on fedora based systems.

_Jira ticket:_ CLOUDP-261264

### DIFF
--- a/build/package/.goreleaser.yml
+++ b/build/package/.goreleaser.yml
@@ -110,11 +110,13 @@ nfpms:
   overrides:
     deb:
       dependencies:
-        - docker
+        - docker-ce
+        - docker-ce-cli
     rpm:
       recommends:
         - podman
-        - docker
+        - docker-ce
+        - docker-ce-cli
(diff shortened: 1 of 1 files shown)


build/package/.goreleaser.yml
```diff
@@ -106,9 +106,15 @@ nfpms:
...
-    - podman
   meta: true
+  overrides:
+    deb:
+      dependencies:
+        - docker
+    rpm:
+      recommends:
+        - podman
+        - docker
```
fmenezes, line 117: I don't think we can do docker the official package is docker-ce on linux right? (ex. https://docs.docker.com/engine/install/debian/)

jeroenvervaeke: good catch, I've updated it

fmenezes (approved):

PR #3105, opened 2024-07-12
CLOUDP-207842: [AtlasCLI] default new deployments to pick latest MongoDB version

jeroenvervaeke:
## Proposed changes
Always pull the latest version of the unified docker image

_Jira ticket:_ CLOUDP-207842

### DIFF
--- a/internal/cli/deployments/setup.go
+++ b/internal/cli/deployments/setup.go
@@ -54,7 +54,7 @@ const (
 	cancelSettings     = "cancel"
 	skipConnect        = "skip"
 	spinnerSpeed       = 100 * time.Millisecond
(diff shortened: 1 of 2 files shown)


fmenezes (approved):

fmenezes:
one edge case scenario that we might consider including is if there is an image already but pull fails we could proceed (scenarios like wiki off)

jeroenvervaeke:
> one edge case scenario that we might consider including is if there is an image already but pull fails we could proceed (scenarios like wifi off)

I'll take a look at that edge case

fmenezes (approved):
thanks for the changes 🎉

PR #3121, opened 2024-07-19
CLOUDP-262694: Fix dockerNotFound error message.

jeroenvervaeke:
## Proposed changes
- Fixed docker error message
- Fixed post_run_opts to:
    - use container engine, get rid of reference to podman
    - show message if docker or podman is not installed, not only podman
- Fix broken tests

_Jira ticket:_ CLOUDP-262694

### DIFF
--- a/internal/container/docker.go
+++ b/internal/container/docker.go
@@ -26,7 +26,7 @@ import (
 	"time"
 )
 
-var ErrDockerNotFound = errors.New("podman not found in your system, check requirements at https://dochub.mongodb.org/core/atlas-cli-deploy-local-reqs")
+var ErrDockerNotFound = fmt.Errorf("%w: docker not found in your system, check requirements at https://dochub.mongodb.org/core/atlas-cli-deploy-local-reqs", ErrContainerEngineNotFound)
 
 type dockerImpl struct {
 }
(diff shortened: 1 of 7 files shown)


internal/container/docker.go
```diff
@@ -26,7 +26,7 @@ import (
 	"time"
 )
 
-var ErrDockerNotFound = errors.New("podman not found in your system, check requirements at https://dochub.mongodb.org/core/atlas-cli-deploy-local-reqs")
+var ErrDockerNotFound = fmt.Errorf("%w: docker not found in your system, check requirements at https://dochub.mongodb.org/core/atlas-cli-deploy-local-reqs", ErrContainerEngineNotFound)
```
fmenezes, line 29: [q] instead of joining the errors here how about doing it on `container_new_linux.go` and `container_new.go`?

jeroenvervaeke: What would be the advantage of doing that?
Keep in mind that on some linux repos we'll have both linux/podman

fmenezes: That is exactly why, one place to wrap the error, I haven't seen the pattern of fmt.Errorf in a package level yet.

It is normally done on functions.

fmenezes (approved):
I'm ok with the change, my comments are nits

PR #3130, opened 2024-07-23
CLOUDP-262686: In windows we can't find Compass hence cli does not launch it

jeroenvervaeke:
## Proposed changes
Use the windows register set by the compass installer.
If the registry value is not set, try the path too.

_Jira ticket:_ CLOUDP-262686

### DIFF
--- a/internal/compass/compass.go
+++ b/internal/compass/compass.go
@@ -18,7 +18,6 @@ import (
 	"context"
 	"errors"
 	"fmt"
-	"os/exec"
 	"time"
 )
 
@@ -26,22 +25,12 @@ const waitForRunningStateDuration = 10 * time.Second
 
 var errCompassExited = errors.New("MongoDB Compass process has exited")
 
-func binPath() string {
-	if p, err := exec.LookPath(compassBin); err == nil {
-		return p
-	}
-	return ""
-}
-
 func Detect() bool {
 	return binPath() != ""
 }
 
 func Run(username, password, mongoURI string) error {
 	path := binPath()
-	if path != compassBin {
-		path += compassBin
-	}
 
 	cmd := compassCmd(path, username, password, mongoURI)
 	if err := cmd.Start(); err != nil {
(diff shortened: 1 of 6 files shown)


internal/compass/compass.go
```diff
@@ -18,30 +18,19 @@ import (
...
 	return binPath() != ""
 }
 
 func Run(username, password, mongoURI string) error {
 	path := binPath()
-	if path != compassBin {
-		path += compassBin
-	}
 
 	cmd := compassCmd(path, username, password, mongoURI)
```
blva, line 35: [nit] can we add an unit test on what happens when path == ""? hoping we can capture as we have windows unit tests running

blva (approved):

internal/compass/compass_bin_windows.go
```diff
@@ -14,4 +14,47 @@
...
 const compassBin = "MongoDBCompass.exe"
+
+func binPath() string {
+	// Find the path to the MongoDB Compass binary using the registry
+	directory, err := compassDirectory()
+	if directory == "" || err != nil {
+		// If the directory is not found in the registry, search in the PATH environment variable
+		p, err := exec.LookPath(compassBin)
+		if err != nil {
+			return ""
```
blva, line 34: [nit] should we print a log debug here as we are ignoring the error or are we checking the value later?

jeroenvervaeke: It's handled later, this is the message when compass is not found.
```
Error: did not find MongoDB Compass, install: https://dochub.mongodb.org/core/install-compass
```

internal/compass/compass.go
jeroenvervaeke: It will fail on line 36, because `cmd.Start` will fail.

We can add a unit test but we would have to introduce mocks and replace a lot of code. I rather do this in a separate PR.

blva (approved):

PR #3131, opened 2024-07-23
CLOUDP-262687: in windows we cannot launch mongosh

jeroenvervaeke:
## Proposed changes
Use `exec.Command` on windows.

_Jira ticket:_ CLOUDP-262687

### DIFF
--- a/internal/mongosh/mongosh_bin_windows.go
+++ b/internal/mongosh/mongosh_bin_windows.go
@@ -14,4 +14,18 @@
 
 package mongosh
 
+import (
+	"os"
+	"os/exec"
+)
+
 const mongoshBin = "mongosh.exe"
+
+func execCommand(args ...string) error {
+	cmd := exec.Command(mongoshBin, args...)
+	cmd.Stdin = os.Stdin
+	cmd.Stdout = os.Stdout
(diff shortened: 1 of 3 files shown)


internal/mongosh/mongosh_bin_windows.go
```diff
@@ -14,4 +14,18 @@
...
 
+import (
+	"os"
+	"os/exec"
+)
+
 const mongoshBin = "mongosh.exe"
+
+func execCommand(args ...string) error {
+	cmd := exec.Command(mongoshBin, args...)
```
fmenezes, line 25: [q] why not use `exec.Command` on every os?

jeroenvervaeke: I'm assuming that the person who originally implemented this had a good reason to use `syscall.Exec` over `exec.Command`. So I opted to leave the existing code as is and implemented a working version for windows.

I can change it if we're positive that there was no special reason to use `syscall.Exec`

jeroenvervaeke: I asked, there's no specific reason. Will update

fmenezes (approved):

PR #3140, opened 2024-07-29
CLOUDP-264043: Integration tests for plugin install command

stefan4h:
## Proposed changes

- Add integration tests for `plugin install`

_Jira ticket:_ CLOUDP-264043

### DIFF
--- a/internal/cli/plugin/install.go
+++ b/internal/cli/plugin/install.go
@@ -35,8 +35,8 @@ type InstallOpts struct {
 }
 
 func (opts *InstallOpts) checkForDuplicatePlugins() error {
-	for _, plugin := range opts.plugins {
-		if plugin.Github != nil && plugin.Github.Equals(opts.githubRelease.owner, opts.githubRelease.name) {
+	for _, p := range opts.plugins {
+		if p.Github != nil && p.Github.Equals(opts.githubRelease.owner, opts.githubRelease.name) {
 			return fmt.Errorf("a plugin from the repository %s is already installed.\nTo update the plugin run: \n\tatlas plugin update %s", opts.repository(), opts.repository())
 		}
 	}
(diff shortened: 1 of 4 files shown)


internal/cli/plugin/install.go
```diff
@@ -35,8 +35,8 @@ type InstallOpts struct {
 }
 
 func (opts *InstallOpts) checkForDuplicatePlugins() error {
-	for _, plugin := range opts.plugins {
-		if plugin.Github != nil && plugin.Github.Equals(opts.githubRelease.owner, opts.githubRelease.name) {
+	for _, p := range opts.plugins {
```
jeroenvervaeke, line 38: Why did you rename this? `plugin` is more readable than `p`

internal/cli/plugin/plugin_asset.go
```diff
@@ -226,7 +225,7 @@ func extractTarGz(src string, dest string) error {
 					return errCopyFileContentExtraction
 				}
 			}
-
+			outFile.Close()
```
jeroenvervaeke, line 228: Wouldn't it be better to put this in a `defer` statement?
You will not close the file when you throw `errCopyFileContentExtraction`

jeroenvervaeke (changes requested):

internal/cli/plugin/install.go
stefan4h: I got an error from the linter saying that I cannot use "plugin" because it is the name of a package

internal/cli/plugin/plugin_asset.go
stefan4h: The linter threw me an error saying that I should not use defer in a loop

jeroenvervaeke: Can you close the file before returning the error too?

jeroenvervaeke (changes requested):

jeroenvervaeke (approved):
LGTM

PR #3147, opened 2024-07-30
CLOUDP-260774: Plugin run functionality

stefan4h:
## Proposed changes
- Set plugin subcommands to `OptionalAuth` to avoid error message for require authentication
- reduce nesting in PersistentPreRunE

_Jira ticket:_ CLOUDP-260774


internal/cli/root/builder.go
```diff
@@ -247,6 +207,57 @@ Use the --help flag with any command for more info on that command.`,
 		federatedauthentication.Builder(),
 	)
 
+	// create copy of the subcommands slice before adding the commands from the plugins
+	// That allows us to differentiate between normal subcommands and plugin subcommands
+	// and we can use it to set all normal subcommands to RequireAuth
+	// and all plugin subcommands to OptionalAuth in shouldCheckCredentials
+	subCommandsWithoutPlugins := make([]*cobra.Command, len(rootCmd.Commands()))
```
tibulca, line 214: Are these changes needed only for identifying commands from plugins vs commands from cli and use this info in `shouldCheckCredentials`?
If yes, wdyt about using annotations? e.g. when registering the plugin commands you can add cmd.Annotations["sourceType"] = "plugin" and cmd.Annotations["sourceName"] = "k8s".
Then in shouldCheckCredentials you can check cmd.Annotations["sourceType"]

stefan4h: done

jeroenvervaeke (approved):

internal/cli/root/builder.go
```diff
@@ -310,6 +317,11 @@ func shouldCheckCredentials(cmd *cobra.Command) AuthRequirements {
 			return r
 		}
 	}
+
+	if sourceType, ok := cmd.Annotations["sourceType"]; ok && sourceType == plugin.SourceType {
```
tibulca, line 321: [optional recommendation] move this logic to a plugin.IsPluginCmd(cmd *cobra.Command).
This way the entire logic (annotate/check annotation) is encapsulated

internal/cli/root/builder_test.go
```diff
@@ -25,16 +25,25 @@ import (
...
 	"github.com/mongodb/mongodb-atlas-cli/atlascli/internal/version"
 	"github.com/spf13/afero"
 )
 
 func TestBuilder(t *testing.T) {
+	rootCmd := Builder()
+
+	pluginCommandCount := 0
+	for _, cmd := range rootCmd.Commands() {
+		if sourceType, ok := cmd.Annotations["sourceType"]; ok && sourceType == plugin.SourceType {
```
tibulca, line 39: here you could use plugin.IsPluginCmd(cmd *cobra.Command)

internal/plugin/plugin.go
```diff
@@ -132,7 +134,10 @@ func (p *Plugin) GetCobraCommands() []*cobra.Command {
 		command := &cobra.Command{
 			Use:   command.Name,
```
tibulca, line 135: [nit] 
```
for _, command := range p.Commands {
           command := &cobra.Command{
``` 
can be :
```
for _, pluginCmd := range p.Commands {
           command := &cobra.Command{
```
It's confusing to redefine a var as something else

internal/plugin/plugin.go
```diff
@@ -132,7 +134,10 @@ func (p *Plugin) GetCobraCommands() []*cobra.Command {
...
 			Short: command.Description,
-			RunE:  p.Run,
+			Annotations: map[string]string{
+				"sourceType": SourceType,
+			},
```
tibulca, line 139: we should not reset the Annotations (can be used for help for example)
suggestion: Annotations: setSourceType(command.Annotations)
and setSourceType:
if cmd.Annotations == nil {
		cmd.Annotations = map[string]string{}
	}
	cmd.Annotations["test"] = "test"

jvincent-mongodb (approved):
LGTM!

internal/plugin/plugin.go
stefan4h: In this case we would not be reseting the annotations because the Command is just being created which means that there cannot be any annotations before that. Or am I missing something?

tibulca: I was thinking about carrying over the annotations from pluginCmd, but that a simple name/description struct, not a cobra command, so you can ignore my suggestion

tibulca (approved):

PR #3151, opened 2024-08-01
CLOUDP-260775: Plugin uninstall command

stefan4h:
## Proposed changes

- Create plugin uninstall command
- e2e tests for uninstall command
- Add validation to plugin discovery functionality to make sure there is all loaded plugins names are unique
- ensure a plugin can only be installed if there is not any plugin with the same name

_Jira ticket:_ CLOUDP-260775

### DIFF
--- a/internal/cli/plugin/plugin.go
+++ b/internal/cli/plugin/plugin.go
@@ -15,10 +15,34 @@
 package plugin
 
 import (
+	"fmt"
+
 	"github.com/mongodb/mongodb-atlas-cli/atlascli/internal/cli"
 	"github.com/mongodb/mongodb-atlas-cli/atlascli/internal/plugin"
 	"github.com/spf13/cobra"
 )
+
+type Opts struct {
+	plugins []*plugin.Plugin
+}
+
+func (opts *Opts) findPluginByGithubValues(owner string, name string) (*plugin.Plugin, error) {
+	for _, p := range opts.plugins {
+		if p.Github != nil && p.Github.Equals(owner, name) {
+			return p, nil
+		}
+	}
+	return nil, fmt.Errorf(`could not find plugin with github values %s/%s`, owner, name)
+}
+
+func (opts *Opts) findPluginByName(name string) (*plugin.Plugin, error) {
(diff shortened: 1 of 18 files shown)


internal/cli/plugin/uninstall.go
```diff
@@ -0,0 +1,89 @@
...
+You can specify a plugin to uninstall using either the "<github-owner>/<github-repository-name>" format or the plugin name.
+`,
+		Args: require.ExactArgs(1),
+		Example: `  # Uninstall a plugin:
+  atlas plugin uninstall mongodb/atlas-cli-plugin-example
+  atlas plugin uninstall atlas-cli-plugin-example`,
+		PreRunE: func(_ *cobra.Command, arg []string) error {
+			// try to parse input to github values
+			// if parsing fails it will be assumed that the input is the plugin name
+			githubValues, err := parseGithubReleaseValues(arg[0])
```
jeroenvervaeke, line 66: I wonder if it would make more sense to just get `arg[0]` here, assign it to something like `opts.pluginName` and move the rest of the code in `run`.

jeroenvervaeke (changes requested):

internal/cli/plugin/plugin.go
```diff
@@ -15,11 +15,35 @@
...
 
 import (
+	"fmt"
+
 	"github.com/mongodb/mongodb-atlas-cli/atlascli/internal/cli"
 	"github.com/mongodb/mongodb-atlas-cli/atlascli/internal/plugin"
 	"github.com/spf13/cobra"
 )
 
+type Opts struct {
```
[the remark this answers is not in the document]
stefan4h: It was called `PluginOpts` initially but the golinter didn't like that as it "stutters" when used in another package: `plugin.PluginOpts`

internal/cli/plugin/uninstall.go
stefan4h: yeah that makes sense.

jeroenvervaeke (approved):

corryroot (approved):
LGTM!

PR #3162, opened 2024-08-06
CLOUDP-260776: Plugin update command

stefan4h:
## Proposed changes
- Create `plugin update` command
- Refactoring of `AssetOpts` to `GithubAssetManager`

_Jira ticket:_ CLOUDP-260776

### DIFF
--- a/docs/command/atlas-plugin-update.txt
+++ b/docs/command/atlas-plugin-update.txt
@@ -1,0 +1,99 @@
+.. _atlas-plugin-update:
+
+===================
+atlas plugin update
+===================
+
+.. default-domain:: mongodb
+
+.. contents:: On this page
+   :local:
+   :backlinks: none
+   :depth: 1
+   :class: singlecol
+
+Update Atlas CLI plugin.
+
+Update an Atlas CLI plugin.
+You can specify a plugin to update using either the "<github-owner>/<github-repository-name>" format or the plugin name.
+Additionally, you can use the "--all" flag to update all plugins.
+
+
+Syntax
+------
+
+.. code-block::
+   :caption: Command Syntax
+
+   atlas plugin update [plugin] [options]
+
+.. Code end marker, please don't delete this comment
+
+Arguments
+---------
+
+.. list-table::
+   :header-rows: 1
+   :widths: 20 10 10 60
+
+   * - Name
+     - Type
+     - Required
+     - Description
+   * - plugin
+     - string
+     - false
+     - Plugin identifier.
+
+Options
+-------
+
+.. list-table::
+   :header-rows: 1
+   :widths: 20 10 10 60
+
+   * - Name
+     - Type
+     - Required
+     - Description
+   * - --all
+     - 
+     - false
+     - Update all plugins.
+   * - -h, --help
+     - 
+     - false
+     - help for update
+
+Inherited Options
+-----------------
+
+.. list-table::
+   :header-rows: 1
+   :widths: 20 10 10 60
+
+   * - Name
+     - Type
+     - Required
+     - Description
+   * - -P, --profile
+     - string
+     - false
+     - Name of the profile to use from your configuration file. To learn about profiles for the Atlas CLI, see https://dochub.mongodb.org/core/atlas-cli-save-connection-settings.
+
+Examples
+--------
+
+.. code-block::
+   :copyable: false
+
+   # Update a plugin:
+   atlas plugin update mongodb/atlas-cli-plugin-example
+   atlas plugin update atlas-cli-plugin-example
+   
+   
+.. code-block::
+   :copyable: false
+
+   # Update all plugins
+   atlas plugin update --all
--- a/docs/command/atlas-plugin-update.txt
+++ b/docs/command/atlas-plugin-update.txt
@@ -59,7 +59,7 @@ Options
    * - --all
      - 
      - false
-     - Update all plugins.
+     - update all plugins
    * - -h, --help
      - 
      - false
--- a/internal/cli/plugin/install.go
+++ b/internal/cli/plugin/install.go
@@ -15,10 +15,8 @@
 package plugin
 
 import (
-	"errors"
 	"fmt"
 	"os"
-	"strings"
 
 	"github.com/google/go-github/v61/github"
 	"github.com/mongodb/mongodb-atlas-cli/atlascli/internal/cli"
@@ -31,16 +29,16 @@ type InstallOpts struct {
 	cli.GlobalOpts
 	cli.OutputOpts
 	Opts
-	AssetOpts
+	githubAssetManager *GithubAssetManager
 }
 
 func (opts *InstallOpts) checkForDuplicatePlugins() error {
-	_, err := opts.findPluginWithGithubValues(opts.githubRelease.owner, opts.githubRelease.name)
+	_, err := opts.findPluginWithGithubValues(opts.githubAssetManager.owner, opts.githubAssetManager.name)
 	if err != nil {
 		return nil
 	}
 
-	return fmt.Errorf("a plugin from the repository %s is already installed.\nTo update the plugin run: \n\tatlas plugin update %s", opts.repository(), opts.repository())
+	return fmt.Errorf("a plugin from the repository %s is already installed.\nTo update the plugin run: \n\tatlas plugin update %s", opts.githubAssetManager.repository(), opts.githubAssetManager.repository())
 }
 
 // checks if the plugin manifest is valid and that the plugin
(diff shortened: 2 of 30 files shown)


docs/command/atlas-plugin-update.txt
```diff
@@ -0,0 +1,99 @@
...
+
+.. contents:: On this page
+   :local:
+   :backlinks: none
+   :depth: 1
+   :class: singlecol
+
+Update Atlas CLI plugin.
+
+Update an Atlas CLI plugin.
```
kyuan-mongodb, line 17: Was it intentional to have this repeat twice?

docs/command/atlas-plugin-update.txt
```diff
@@ -0,0 +1,99 @@
...
+   :widths: 20 10 10 60
+
+   * - Name
+     - Type
+     - Required
+     - Description
+   * - --all
+     - 
+     - false
+     - Update all plugins.
```
kyuan-mongodb, line 62:
```suggestion
     - update all plugins
```
Nit/Suggestion: keeping the formatting uniform to the help option below

kyuan-mongodb (commented):
Looks good overall, but a few things to consider before I approve!

docs/command/atlas-plugin-update.txt
stefan4h: Yes that is intentional. Line 15 is the short description shown in the 'help' output and line 17 is the first line of the long message shown in the 'help <this-command>' output.

internal/cli/plugin/install.go
```diff
@@ -31,16 +29,16 @@ type InstallOpts struct {
 	cli.GlobalOpts
 	cli.OutputOpts
 	Opts
-	AssetOpts
+	githubAssetManager *GithubAssetManager
```
jeroenvervaeke, line 32: Could we consider dropping the `Manager` part?

I think this makes more sense:
```go
githubAsset *GithubAsset
```

internal/cli/plugin/plugin.go
```diff
@@ -15,15 +15,37 @@
...
 	"github.com/spf13/cobra"
 )
 
 type Opts struct {
-	plugins []*plugin.Plugin
+	plugins          []*plugin.Plugin
+	existingCommands []*cobra.Command
+}
+
+func (opts *Opts) findPluginWithArg(arg string) (*plugin.Plugin, error) {
```
jeroenvervaeke, line 32: Could you add a small comment on top of this function describing what it does? "With arg" is a very generic name.

Having some examples of what arg parameter could look like would improve readability by a lot.

internal/cli/plugin/plugin_github_asset_manager.go
```diff
@@ -43,55 +42,63 @@ var (
...
 	errSaveAssetToPluginDir         = errors.New("failed to save asset to plugin directory")
 	errCreateDirToExtractAssetFiles = errors.New("failed to create to plugin directory to extract assets in")
+	errCreatePluginAssetFromPlugin  = errors.New("failed to create plugin asset from plugin")
 )
 
-type GithubRelease struct {
-	owner   string
-	name    string
-	version *semver.Version
+type GithubAssetManager struct {
```
jeroenvervaeke, line 48: See other comment about renaming

internal/cli/plugin/plugin_github_asset_manager.go
```diff
@@ -43,55 +42,63 @@ var (
...
+		name:  p.Github.Name,
+	}, nil
+}
+
+func (g *GithubAssetManager) getPluginDirectoryName() string {
+	return fmt.Sprintf("%s@%s", g.owner, g.name)
 }
 
-func (opts *AssetOpts) getPluginAssetInfo() ([]*github.ReleaseAsset, error) {
+func (g *GithubAssetManager) getPluginAssetInfo() ([]*github.ReleaseAsset, error) {
```
jeroenvervaeke, line 74: Would it make more sense to call this something along the lines of `getReleaseAssets`?
- It returns an array, but your function name is singular
- It returns some struct called `ReleaseAsset`

internal/cli/plugin/plugin_github_asset_manager.go
```diff
@@ -43,55 +42,63 @@ var (
...
-			return nil, fmt.Errorf("could not find the release %s release for %s", opts.githubRelease.name, opts.repository())
+			return nil, fmt.Errorf("could not find the release %s release for %s", g.name, g.repository())
 		}
 	}
 
 	return release.Assets, nil
 }
 
-func (opts *AssetOpts) getAssetID(assets []*github.ReleaseAsset) (int64, error) {
+func (g *GithubAssetManager) getAssetID(assets []*github.ReleaseAsset) (int64, error) {
```
jeroenvervaeke, line 101: Would it make sense to just call this `getID`?
Since the name of this struct will already have the word asset in it? (`GithubAsset`)

internal/cli/plugin/update.go
```diff
@@ -0,0 +1,265 @@
...
+	assetID, err := githubAssetRelease.getAssetID(assets)
+	if err != nil {
+		return err
+	}
+
+	// download plugin asset zip file and save it as ReadCloser
+	rc, err := githubAssetRelease.getPluginAssetAsReadCloser(assetID)
+	if err != nil {
+		return err
+	}
```
jeroenvervaeke, line 107: should we have a `defer rc.Close()` here?

internal/cli/plugin/update.go
```diff
@@ -0,0 +1,265 @@
...
+
+	// rename old plugin directory to <plugin-directory>_old so we can rollback in case something goes wrong
+	oldPluginDirectoryPath := existingPlugin.PluginDirectoryPath + "_old"
+	err = os.Rename(existingPlugin.PluginDirectoryPath, oldPluginDirectoryPath)
+	if err != nil {
+		return err
+	}
+	defer os.RemoveAll(oldPluginDirectoryPath)
```
jeroenvervaeke, line 137: Not sure if this should be here.
- Happy path: we move the old plugin, so this doesn't have to run
- Unhappy path: we move the old plugin back, so this doesn't have to run

internal/cli/plugin/update.go
```diff
@@ -0,0 +1,265 @@
...
+	"github.com/mongodb/mongodb-atlas-cli/atlascli/internal/log"
+	"github.com/mongodb/mongodb-atlas-cli/atlascli/internal/plugin"
+	"github.com/mongodb/mongodb-atlas-cli/atlascli/internal/usage"
+	"github.com/spf13/cobra"
+)
+
+var (
+	errTooManyArguments        = errors.New(`either the "--all" flag or the plugin identifier can be provided, but not both`)
+	errNotEnoughArguments      = errors.New(`either the "--all" flag or the plugin identifier needs to be provided`)
+	errPluginHasNoGithubValues = errors.New(`specified plugin does not have any Github values`)
```
jeroenvervaeke, line 36: This error message is not super clear. Could you add something like "are you sure you installed this plugin using `plugin install`?". (Just a suggestion, I'm sure there's a better way to phrase this)

internal/plugin/plugin.go
```diff
@@ -77,7 +78,11 @@ func GetAllValidPlugins(existingCommands []*cobra.Command) []*Plugin {
 	// Convert manifests to plugins
 	plugins := make([]*Plugin, 0, len(manifests))
 	for _, manifest := range manifests {
-		plugins = append(plugins, createPluginFromManifest(manifest))
+		plugin, err := createPluginFromManifest(manifest)
+		if err != nil {
+			logPluginWarning(err.Error())
```
jeroenvervaeke, line 83: shouldn't this be:
```go
if err != nil {
    logPluginWarning(err.Error())
    continue
}
```

you don't want to add a plugin to the list if `err != nil`

jeroenvervaeke (changes requested):

internal/cli/plugin/install.go
stefan4h: I agree, just changed it

internal/cli/plugin/plugin.go
stefan4h: done

internal/cli/plugin/plugin_github_asset_manager.go
stefan4h: done

internal/cli/plugin/plugin_github_asset_manager.go
stefan4h: done

internal/cli/plugin/plugin_github_asset_manager.go
stefan4h: done

internal/cli/plugin/update.go
stefan4h: done

internal/cli/plugin/update.go
stefan4h: This is my reasoning:

- Happy path: new version of the plugin successfully downloaded, validated and moved to the right place -> we can delete the old plugin files
- Unhappy path: something goes wrong when installing the new version of the plugin -> `oldPluginDirectoryPath` will be used to rollback to the old version of the plugin ([see code here](https://github.com/mongodb/mongodb-atlas-cli/pull/3162/files/e5161f3d7f91822826b61f3c8eb7c763dbfa7798#diff-5148a859ca9e843ca53da026060c474eb50574bc1be614bb63285895f042668cR143)) so this `RemoveAll` call does nothing

internal/cli/plugin/update.go
stefan4h: done

internal/plugin/plugin.go
stefan4h: yeah, oversight ....

jeroenvervaeke (approved):
LGTM!

kyuan-mongodb (approved):
LGTM!

PR #3168, opened 2024-08-08
CLOUDP-266545: Add sample stream to k8s test expected output

roothorp:
## Proposed changes

_Jira ticket:_ CLOUDP-#

Closes #[issue number]

## Checklist

### DIFF
--- a/test/e2e/atlas/kubernetes_config_generate_test.go
+++ b/test/e2e/atlas/kubernetes_config_generate_test.go
@@ -1000,144 +1000,171 @@ func TestProjectWithStreamsProcessing(t *testing.T) {
 		require.NoError(t, err, "should not fail on decode")
 		require.NotEmpty(t, objects)
 
-		expected := []runtime.Object{
-			&akov2.AtlasStreamInstance{
-				TypeMeta: metav1.TypeMeta{
-					Kind:       "AtlasStreamInstance",
-					APIVersion: "atlas.mongodb.com/v1",
-				},
-				ObjectMeta: metav1.ObjectMeta{
-					Name: resources.NormalizeAtlasName(
-						fmt.Sprintf("%s-%s", generator.projectName, generator.streamInstanceName),
-						resources.AtlasNameToKubernetesName(),
-					),
-					Namespace: targetNamespace,
-					Labels: map[string]string{
-						"mongodb.com/atlas-resource-version": "2.4.0",
-					},
-				},
-				Spec: akov2.AtlasStreamInstanceSpec{
-					Name: generator.streamInstanceName,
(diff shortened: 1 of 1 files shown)


test/e2e/atlas/kubernetes_config_generate_test.go
```diff
@@ -1039,6 +1040,14 @@ func TestProjectWithStreamsProcessing(t *testing.T) {
 									),
 									Namespace: targetNamespace,
 								},
+								// sample_stream_solar is a sample connection which is added by default to new stream instances
+								{
+									Name: resources.NormalizeAtlasName(
+										fmt.Sprintf("%s-%s-%s", generator.projectName, generator.streamInstanceName, sampleStreamName),
```
roothorp, line 1046: From what I can tell, doing this doesn't achieve what we want in tests, because the generated project names are so long that the instance and connection names are truncated (so both of these have the same name in k8s). This only affects tests, but I don't know if we want to address it.

igor-karpukhin (dismissed):

filipcirtog (dismissed):

test/e2e/atlas/kubernetes_config_generate_test.go
```diff
@@ -1052,52 +1061,84 @@ func TestProjectWithStreamsProcessing(t *testing.T) {
...
-						},
-						ObjectMeta: metav1.ObjectMeta{
-							Name: resources.NormalizeAtlasName(
-								fmt.Sprintf("%s-%s-%s", generator.projectName, generator.streamInstanceName, generator.streamConnectionName),
-								resources.AtlasNameToKubernetesName(),
-							),
-							Namespace: targetNamespace,
-							Labels: map[string]string{
-								"mongodb.com/atlas-resource-version": "2.4.0",
+				if connection.Spec.Name == sampleStreamName {
```
helderjs, line 1064: Why do you need this conditional check? Don't they always come in the same order?
also, I think `assert` has a `MatchElements()` that check slices independently of order

helderjs (changes requested):

test/e2e/atlas/kubernetes_config_generate_test.go
roothorp: We're receiving more objects from `getK8SEntities` than we're checking for (like the Project), which means using assert's ElementsMatch doesn't work unless we add everything else to this list.

We need to do this check because we're just looping over this list and checking for a stream connection, which previously we assumed would be the only one. Now it could be any one of the two (one which we created, one which is always there).

filipcirtog (approved):

helderjs (approved):

PR #3174, opened 2024-08-12
CLOUDP-260767: First class plugin support

stefan4h:
## Proposed changes

- Added first class plugin support:
  - details of a first class plugin can be added to `first-class-plugin.json` to make the plugin command(s) appear as if they are part of the Atlas CLI. When the user runs the plugin for the first time it will be installed

_Jira ticket:_ CLOUDP-260767


internal/cli/plugin/first_class.go
```diff
@@ -0,0 +1,139 @@
...
+	"github.com/mongodb/mongodb-atlas-cli/atlascli/internal/plugin"
+	"github.com/spf13/cobra"
+)
+
+const (
+	SourceType = "first-class"
+)
+
+//go:embed first-class-plugins.json
+var firstClassPluginsFile []byte
```
jeroenvervaeke, line 32: [q] is there any reason to use this approach over having a static variable in the code?

This approach can fail at runtime when the JSON contains errors.

stefan4h: Changed it to use a static variable instead

internal/cli/plugin/first_class.go
```diff
@@ -0,0 +1,143 @@
...
+import (
+	"fmt"
+
+	"github.com/google/go-github/v61/github"
+	"github.com/mongodb/mongodb-atlas-cli/atlascli/internal/plugin"
+	"github.com/spf13/cobra"
+)
+
+const (
+	SourceType = "first-class"
```
jeroenvervaeke, line 26: Better change this name to `FirstClassSourceType` or something similar, since you've created this in the module `plugin`, not in the module `first_class_plugin` or something similar.

internal/cli/plugin/first_class.go
```diff
@@ -0,0 +1,143 @@
...
+}
+
+type FirstClassPlugin struct {
+	Name     string
+	Github   *Github
+	Commands []*Command
+}
+
+func IsFirstClassPluginCmd(cmd *cobra.Command) bool {
+	if cmdSourceType, ok := cmd.Annotations["sourceType"]; ok && cmdSourceType == SourceType {
```
jeroenvervaeke, line 63: can we make these annotations consts? `sourceType`, `sourcePluginName`, etc

internal/cli/root/builder.go
```diff
@@ -253,6 +253,11 @@ Use the --help flag with any command for more info on that command.`,
 	)
 
 	pluginCmd.RegisterCommands(rootCmd)
+	err := pluginCmd.RegisterFirstClassPluginCommands(rootCmd)
```
jeroenvervaeke, line 256: Can you make `RegisterFirstClassPluginCommands` part of `RegisterCommands` so we can keep our original of having one line to register all plugins? (+ error handling ofc)

internal/cli/plugin/plugin.go
```diff
@@ -69,7 +78,7 @@ func (opts *Opts) findPluginWithName(name string) (*plugin.Plugin, error) {
 }
 
 func RegisterCommands(rootCmd *cobra.Command) {
-	plugins := plugin.GetAllValidPlugins(rootCmd.Commands())
+	plugins := plugin.GetAllValidPlugins(createExistingCommandsMap(rootCmd.Commands()))
```
jeroenvervaeke, line 81: Why don't you register your first class plugins in this function so you don't have to rely on annotations at all?

internal/cli/plugin/first_class.go
stefan4h: done

internal/cli/root/builder.go
stefan4h: done

internal/cli/plugin/plugin.go
stefan4h: done

internal/cli/plugin/first_class.go
stefan4h: done

internal/plugin/plugin.go
```diff
@@ -32,59 +32,63 @@ var (
...
+	// Load manifests from plugin directories
+	manifests := loadManifestsFromPluginDirectories()
+
+	// Convert manifests to plugins
+	plugins := convertManifestsToPlugins(manifests)
+
+	return plugins
+}
+
+func GetAllValidPlugins(existingCommandsMap map[string]bool) []*Plugin {
```
jeroenvervaeke, line 74: Can you just pass a slice of strings here? (commands).

The map you're passing is basically:
```go
	s := map[string]bool{
		"command1": true,
		"command2": true,
		"command3": true,
		"command4": true,
		"command5": true,
	}
```

If you need a `map[string]bool` inside of this function you can always convert it inside of the function.

internal/plugin/plugin_manifest.go
```diff
@@ -210,18 +236,14 @@ func removeManifestsWithDuplicateNames(manifests []*Manifest) ([]*Manifest, []st
 	return uniqueManifests, duplicateManifestNames
 }
 
-func getUniqueManifests(manifests []*Manifest, existingCommands []*cobra.Command) ([]*Manifest, []*Manifest) {
-	existingCommandsMap := make(map[string]bool)
+func getUniqueManifests(manifests []*Manifest, existingCommandsMap map[string]bool) ([]*Manifest, []*Manifest) {
```
jeroenvervaeke, line 239: Same comment as before, a slice would make more sense here than `map[string]bool`

internal/cli/plugin/plugin.go
```diff
@@ -69,12 +82,14 @@ func (opts *Opts) findPluginWithName(name string) (*plugin.Plugin, error) {
...
```
jeroenvervaeke, line 85: It would probably make more sense here to either:
- keep passing cobra commands
- pass a slice with all commands

See other remark for more details.

jeroenvervaeke (changes requested):

jeroenvervaeke (approved):

PR #3192, opened 2024-08-16
CLOUDP-265397: [CLI] Move random port binding to docker instead of doing in the cli

jeroenvervaeke:
## Proposed changes
- Removed logic to allocate a port in the CLI
     - Implemented logic for docker / podman
     - Wrote new tests 
- Improved `updateFields` to use actually published ports and not config (this contains the random port chosen by docker/podman)
- Fixed / updated existing tests

_Jira ticket:_ CLOUDP-265397

### DIFF
--- a/internal/container/docker_test.go
+++ b/internal/container/docker_test.go
@@ -1,0 +1,81 @@
+package container
+
+import "testing"
+
+func TestPortMappingFlag(t *testing.T) {
+	tests := []struct {
+		name     string
+		input    PortMapping
+		expected string
+	}{
+		{
+			name: "All fields populated",
+			input: PortMapping{
+				HostAddress:       "127.0.0.1",
+				HostPort:          8080,
+				ContainerPort:     80,
+				ContainerProtocol: "tcp",
+			},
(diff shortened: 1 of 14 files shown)


internal/container/docker_test.go
```diff
@@ -0,0 +1,95 @@
...
+				HostAddress:   "localhost",
+				ContainerPort: 8080,
+			},
+			expected: "localhost::8080",
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			result := portMappingFlag(tt.input)
```
gssbzn, line 89: can this be run in parallel
```suggestion
		t.Run(tt.name, func(t *testing.T) {
			t.Parallel()
			result := portMappingFlag(tt.input)
```

gssbzn (dismissed):
LGTM

internal/container/docker_test.go
jeroenvervaeke: Yes they could, thanks for the suggestion!

andreaangiolillo (approved):

PR #3198, opened 2024-08-19
CLOUDP-267258: Assign mongocli tickets to correct team

blva:
## Proposed changes

_Jira ticket:_ CLOUDP-#

Closes #[issue number]

## Checklist

### DIFF
--- a/.github/workflows/dependabot.yml
+++ b/.github/workflows/dependabot.yml
@@ -29,7 +29,7 @@
           PR_NUMBER: ${{ github.event.pull_request.number }}
           PR_URL: ${{ github.event.pull_request.html_url }}
           JIRA_API_TOKEN: ${{ secrets.JIRA_API_TOKEN }}
-          JIRA_ASSIGNEE: ${{ secrets.ASSIGNEE_JIRA_TICKET }}
+          JIRA_ASSIGNEE: ${{ secrets.ASSIGNEE_MONGOCLI_JIRA_TICKET }}
         run: |
           json_response=$(curl --request POST \
             --url 'https://jira.mongodb.org/rest/api/2/issue' \


.github/workflows/dependabot.yml
```diff
@@ -29,7 +29,6 @@
           PR_NUMBER: ${{ github.event.pull_request.number }}
           PR_URL: ${{ github.event.pull_request.html_url }}
           JIRA_API_TOKEN: ${{ secrets.JIRA_API_TOKEN }}
-          JIRA_ASSIGNEE: ${{ secrets.ASSIGNEE_JIRA_TICKET }}
```
blva, line 32: for mongocli, i'm removing the assignee since it was being set to apix escalation

build/ci/library_owners_jira.json
```diff
@@ -1,4 +1,4 @@
 {
-	"mongocli": "22223",
+	"mongocli": "27065",
```
blva, line 2: updated team ID to be the new team

blva: keeping mongocli since the team still doesn't have a team ID

.github/workflows/dependabot.yml
andreaangiolillo: this means that the new ticket won't have an assignee. Just flagging this to make sure we are all okay with this.

andreaangiolillo, line 32: Does this action work for the mongocli-master branch? I see these issues:

1. As default, the GH action will use master instead of the mongocli-master branch to run the action, which means it will check the dependencies on master. 
2. the action checks out the master branch instead of mongocli-master (line 17)


I think you need to define a new action in master that run for pull request agains mongocli-master and checkout the mongocli-master branch

blva: yep, ok with that so far until the team has their own escalation, i'll make sure to ask them if they have one

build/ci/library_owners_jira.json
fmenezes: should we split apix-2 and mongocli?

fmenezes: we can't simply say they own all dependencies of the repo, right?

blva: do you mean handover the mongocli team to them? I'm not sure how much logic we have depending on that team so not sure

fmenezes: I think we will need another PR in `master` to replace `mongocli` with `apix-2`

gssbzn: I think what Filipe means is not using this file, we created this logic in master because there were k8s libs we didn't own but on this branch everything is owned by EA

blva: ohh ok, so for mongocli we can just put EA

gssbzn: yes

.github/workflows/dependabot.yml
andreaangiolillo: which action implementation it will use? the one in `master` or the one in `mongocli-master`?  I think it will use the one in `master` 🤔 (I may be wrong, just make sure to double check)

build/ci/library_owners_jira.json
fmenezes: what @gssbzn said plus a review in https://github.com/mongodb/mongodb-atlas-cli/pull/3206

.github/workflows/dependabot.yml
blva: My understanding is that it's using the current branch one, since it is createing the action with "MongoCLI" instead of "AtlasCLI"

fmenezes (approved):

PR #3231, opened 2024-09-03
CLOUDP-269965: [Atlas CLI] Documentation needs to be updated for some commands

andreaangiolillo:
## Proposed changes

_Jira ticket:_ CLOUDP-269965

This PR updates the documentations for the commands listed in CLOUDP-269965

```bash
./bin/atlas accesslog list                                                                                                                            
Error: at least one of the flags in the group [clusterName hostname] is required
```

```bash
./bin/atlas accesslog list --help                                                                                                                         3s  1.22.5  16.18.0
To use this command, you must authenticate with a user account or an API key with the Project Monitoring Admin role.

Usage:
  atlas accessLogs list {--clusterName clusterName | --hostname hostname} [flags]

Aliases:
  list, ls

Examples:
  # Return a JSON-formatted list of all authentication requests made against the cluster named Cluster0 for the project with ID 618d48e05277a606ed2496fe:
  atlas accesslogs list --output json --projectId 618d48e05277a606ed2496fe --clusterName Cluster0

  # Return a JSON-formatted list of all authentication requests made against the hostname named atlas-dqrlj4-shard-00-00.g1nxq.mongodb-dev.net for the project stored in the profile atlas-dev:
  atlas accesslogs list --hostname atlas-dqrlj4-shard-00-00.g1nxq.mongodb-dev.net --profile atlas-dev --output json


Flags:
      --authResult string    Authentication attempts that Atlas should return. When set to success, Atlas filters the log to return only successful authentication attempts. When set to fail, Atlas filters the log to return only failed authentication attempts.
      --clusterName string   Name of the cluster. To learn more, see https://dochub.mongodb.org/core/create-cluster-api.
      --end string           Timestamp in the number of milliseconds that have elapsed, since the UNIX Epoch, for the last entry that Atlas returns from the database access logs.
  -h, --help                 help for list
      --hostname string      The fully qualified domain name (FQDN) of the target node that should receive the authentication attempt.
      --ip string            IP address that attempted to authenticate with the database. Atlas filters the returned logs to include documents with only this IP address.
      --nLog int             Maximum number of log lines to return.
  -o, --output string        Output format. Valid values are json, json-path, go-template, or go-template-file. To see the full output, use the -o json option.
      --projectId string     Hexadecimal string that identifies the project to use. This option overrides the settings in the configuration file or environment variable.
      --start string         Timestamp in the number of milliseconds that have elapsed, since the UNIX Epoch, for the first entry that Atlas returns from the database access logs.

Global Flags:
  -P, --profile string   Name of the profile to use from your configuration file. To learn about profiles for the Atlas CLI, see https://dochub.mongodb.org/core/atlas-cli-save-connection-settings.
```


internal/cli/datalakepipelines/runs/list.go
```diff
@@ -59,17 +59,17 @@ func (opts *ListOpts) Run() error {
 	return opts.Print(r)
 }
 
-// atlas dataLakePipelines runs list [--projectId projectId].
+// ListBuilder represents atlas dataLakePipelines runs list [--projectId projectId].
 func ListBuilder() *cobra.Command {
 	opts := &ListOpts{}
 	cmd := &cobra.Command{
-		Use:     "list",
+		Use:     "list --pipeline pipeline",
```
gssbzn, line 66: we historically have never documented options here

andreaangiolillo: Yes, I understand. This was the general feedback from customers, though. I see value in including it for required options in the list commands, especially since they typically don’t have any required options. WDYT?

gssbzn: are you planning to backfill all commands with the missing required options?

andreaangiolillo: only the list commands if we all agree on this

gssbzn: why the list commands and not the others?

andreaangiolillo: the main reason is that the `list` commands are (usually) the ones that don't require flags so it is easy for the user to forget about them.

davidhou17 (approved):

fmenezes (dismissed):
LGTM

jeroenvervaeke:
@andreaangiolillo, looks like this PR is stale. Do we still need these changes? Or can we close it?

internal/cli/datalakepipelines/runs/list.go
andreaangiolillo: Hi @mmarcon 👋 Do you have any thoughts on using required flags for list commands in the command syntax? Should we apply this new format to all commands or only to list commands? Or should we use it at all?

Customers have provided feedback that required flags for list commands are not displayed in the syntax section. [Example](https://www.mongodb.com/docs/atlas/cli/current/command/atlas-accessLogs-list/#syntax): 
Screenshot 2024-09-25 at 10 22 06


Even though the `atlas accessLogs list` command requires a clusterName or hostname flag, they are not shown in the syntax section of the documentation.

This PR changes that section to be 
```
Usage:
  atlas accessLogs list {--clusterName clusterName | --hostname hostname} [flags]
```

gssbzn: I'm goign to add here that if we decide to document "some" flags we shoudl at least follow http://docopt.org/

meaning it should be 
```
atlas accessLogs list (--clusterName <clusterName> | --hostname <hostname>) [options]
```

mmarcon: @andreaangiolillo not sure you are going to be happy with my answer 😂

My perspective is that we should aim for consistency, i.e. if we believe a good user experience means including the required flags in the usage (which is something I can get behind, given that before working full time on the CLI understanding what flags were necessary without having to check the docs was something I struggled with too), then we should do it everywhere.

Having said that, I don't know if, with auto-generation coming, going through all the commands and adding the required flags manually is a good use of y'all's time. We can make sure that we account for displaying required flags as part of the auto-generation implementation.

@dianchenghu do you feel differently?

dianchenghu: @mmarcon I agree that the entire consistency issue should be reviewed at once (this PR could close one ticket, but we're aware there will be a bigger one still open in the backlog). For now, since the user particularly mentioned the accessLogs list example (and I believe it's not just a single case, list command is often used by human). I believe what @gssbzn  proposed here is not adding more confusion and it's a nice quick fix to have as soon as possible.

gssbzn (approved):

PR #3299, opened 2024-10-04
CLOUDP-276968: improve whoami err message

tibulca:
## Proposed changes
improve whoami err message

_Jira ticket:_ CLOUDP-276968

### DIFF
--- a/internal/cli/auth/whoami.go
+++ b/internal/cli/auth/whoami.go
@@ -25,12 +25,13 @@ import (
 )
 
 type whoOpts struct {
-	OutWriter io.Writer
-	account   string
+	OutWriter   io.Writer
+	authSubject string
+	authType    string
 }
 
 func (opts *whoOpts) Run() error {
-	_, _ = fmt.Fprintf(opts.OutWriter, "Logged in as %s\n", opts.account)
+	_, _ = fmt.Fprintf(opts.OutWriter, "Logged in as %s %s\n", opts.authSubject, opts.authType)
 
 	return nil
 }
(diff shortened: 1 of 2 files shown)


andreaangiolillo (dismissed):

internal/cli/auth/whoami.go
```diff
@@ -35,7 +35,7 @@ func (opts *whoOpts) Run() error {
 	return nil
 }
 
-var ErrUnauthenticated = errors.New("not logged in")
+var ErrUnauthenticated = errors.New("not logged in with an Atlas account")
 
 func AccountWithAccessToken() (string, error) {
```
gssbzn, line 40: what about
```suggestion
func AccountWithAccessToken() (string, error) {
	if config.PublicAPIKey() != "" {
		return config.PublicAPIKey(), nil
	}
	if config.AccessToken() != "" {
		return config.AccessTokenSubject()
	}
	return  "", ErrUnauthenticated
}
```

gssbzn (approved):
LGTM feel free to double check the copy with docs

internal/cli/auth/whoami.go
fmenezes: if we are doing this shouldn't we say `Logged in as email@server.com` and `Logged in as public key XXX` as opposed to `Logged in as XXX`

fmenezes:
```suggestion
func AccountWithAccessToken() (string, string, error) {
	if config.PublicAPIKey() != "" {
		return config.PublicAPIKey(), "key", nil
	}
	if config.AccessToken() != "" {
		return config.AccessTokenSubject(), "account"
	}
	return  "", "", ErrUnauthenticated
}
```

tibulca: makes sense to add a bit more context as I also realized that AccountWithAccessToken() is re-used in the setup cmd so I split the logic too

fmenezes (approved):

corryroot (approved):
LGTM!

PR #3300, opened 2024-10-07
CLOUDP-276465 "atlas clusters upgrade" does not enforce mandatory params

fmenezes:
## Proposed changes
"atlas clusters upgrade" does not enforce mandatory params

_Jira ticket:_ CLOUDP-276465

Closes #[issue number]

## Checklist

### DIFF
--- a/internal/cli/clusters/upgrade.go
+++ b/internal/cli/clusters/upgrade.go
@@ -186,5 +186,7 @@ func UpgradeBuilder() *cobra.Command {
 
 	_ = cmd.MarkFlagFilename(flag.File)
 
+	cmd.MarkFlagsOneRequired(flag.File, flag.Tier, flag.DiskSizeGB, flag.MDBVersion, flag.EnableTerminationProtection, flag.DisableTerminationProtection, flag.Tag)
+
 	return cmd
 }


internal/cli/clusters/upgrade.go
```diff
@@ -186,5 +186,7 @@ func UpgradeBuilder() *cobra.Command {
 
 	_ = cmd.MarkFlagFilename(flag.File)
 
+	cmd.MarkFlagsOneRequired(flag.File, flag.Tier, flag.DiskSizeGB, flag.MDBVersion, flag.EnableTerminationProtection, flag.DisableTerminationProtection, flag.Tag)
```
jeroenvervaeke, line 189: Does this mean that either `file` is required, or all the other ones?

Because if you provide tier, disksizegb is also required.
I also noticed that disksizegb can't be a float but has to be an integer.

fmenezes: so `cmd.MarkFlagsOneRequired` means at least one of the flags should be passed. 

`cmd.MarkFlagsMutuallyExclusive(flag.File, flag.Tier)` means if you're passing --tier you may not pass --file and vice versa and that is why --file is mutually exclusive to all other flags

in essence nothing is mandating that --tier and --disksizegb must be set together

fmenezes: > I also noticed that disksizegb can't be a float but has to be an integer.
no too sure what you mean our code is float64 and the API definition is a number in https://www.mongodb.com/docs/atlas/reference/api-resources-spec/v1/#tag/Clusters/operation/upgradeSharedCluster, float64 is coming from our sdk in https://github.com/mongodb/go-client-mongodb-atlas/blob/master/mongodbatlas/clusters.go#L52C2-L52C9

jeroenvervaeke: The SDK and the default we provide need to be corrected.

```
atlas cluster upgrade upgrade-test --tier M10
Error: POST [redacted]/clusters/tenantUpgrade: 400 (request "CLUSTER_DISK_SIZE_NOT_WHOLE_NUMBER") The cluster's disk size of 0.5 GB is invalid. The disk size must be a positive whole number.
```

When I provide `1` it works (`1GB` only works for `M2`, `M10` needs `10Gb` min)
```
atlas cluster upgrade upgrade-test --tier M10 --diskSizeGB 1
Error: POST [redacted]/clusters/tenantUpgrade: 400 (request "ATLAS_GENERAL_ERROR") Reason: Cluster disk size 1 GB lower than the minimum disk size of 10 GB for instance size M10.
```

fmenezes: you command `atlas cluster upgrade upgrade-test --tier M10` is most likely a `M0` to `M10`, we have no defaults `M0`s are created as 0.5GB if upgrading to `M10` you have to pass disk otherwise it won't work, that is because we have no PATCH API, only POST API, the API design here does not help.

jeroenvervaeke (approved):

PR #3338, opened 2024-10-22
task: update documentation for golang

wtrocki:
## Description

Updating documentation to cover golang version requirement

## Reason
```
 Building atlas binary
go build -ldflags "-s -w -X github.com/mongodb/mongodb-atlas-cli/atlascli/internal/version.GitCommit=436cdc259407d225f622e5a407a43e43f4070d99 -X github.com/mongodb/mongodb-atlas-cli/atlascli/internal/version.Version=1.19.0-374-g436cdc259"  -o ./bin/atlas ./cmd/atlas
go: go.mod requires go >= 1.23.1 (running go 1.21.3; GOTOOLCHAIN=local)
make: *** [build] Error 1
```
## Separate concern

Go.mod file does not use toolchain directive which can automatically request proper golang version to be used. 
Since I moved from Atlas CLI I noticed we no longer use adsf tools definitions.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -50,7 +50,7 @@ To make the contribution process as seamless as possible, we ask for the followi
 #### Prerequisite Tools
 
 - [Git](https://git-scm.com/)
-- [Go (at least Go 1.22)](https://golang.org/dl/)
+- [Go (at least Go 1.23.1)](https://golang.org/dl/)
 
 #### Environment
 


blva (approved):

wtrocki:
Thank you, @blva Created #3340 as follows up to enable toolchain support.

gssbzn:
> Do we have any reason to not define tooling directive?

https://github.com/mongodb/mongodb-atlas-cli/pull/2779/files#r1530657881

wtrocki:
Thank you for reference. I seen that PR and I did also removed adsf support for the repository:
https://github.com/mongodb/mongodb-atlas-cli/pull/2779/files#r1530658440

My understanding is that we want autoinstallation feature to be disabled. 
That would mean this PR was correct (bumped required version that needs to be installed manually).
Closing toolchain [PR](https://github.com/mongodb/mongodb-atlas-cli/pull/3340/files) that adds extra auto option as fallback to local

gssbzn:
At the time I think we discussed offline (sorry we didn't capture it in the PR) given using both asdf and auto toolchain is an issue
- do we disable auto toolchain and prefer asdf
- do we disable asdf and prefer toolchain 

And we decided to disable both and let devs decide what to use, in particular toolchain is disabled for the makefile because installation of the dev tools is wonky with  the autotool detection

wtrocki:
@gssbzn Thank you. All makes sense. I tried to look if we can add any information into contributing guide but we already mention manual installation for specific version and no presence of toolchain/asdf. All good.

PR #3407, opened 2024-11-22
CLOUDP-280745: Build a layer that glues CLI and command framework together

jeroenvervaeke:
## Proposed changes
- convert cobra context to api command request
- resolving flags from profile
- handle file/stdin
- support versions 
- separated strings code

_Jira ticket:_ CLOUDP-280745

Will add E2E tests as a separate ticket after `--format` is implemented.


internal/cli/api/api.go
```diff
@@ -19,13 +19,18 @@ import (
...
 
 	"github.com/iancoleman/strcase"
 	"github.com/mongodb/mongodb-atlas-cli/atlascli/internal/api"
 	"github.com/mongodb/mongodb-atlas-cli/atlascli/internal/log"
 	"github.com/spf13/cobra"
+	"github.com/spf13/pflag"
+)
+
+var (
+	ErrFailedToSetUntouchedFlags         = errors.New("failed to set untouched flags")
```
fmenezes, line 31: what do untouched flags mean in this context?

jeroenvervaeke: The code is setting which flags are available.
For example:
- `--groupId`
- `--orgId`

If the user provided `--groupId`, but not `--orgId`, then `--orgId` is considered "untouched"

PR #3421, opened 2024-11-26
CLOUDP-279515: add code coverage to PR

jeroenvervaeke:
## Proposed changes
add code coverage to PR
_Jira ticket:_ CLOUDP-279515


Makefile
```diff
@@ -1,7 +1,7 @@
 # A Self-Documenting Makefile: http://marmelab.com/blog/2016/02/29/auto-documented-makefile.html
 
 GOLANGCI_VERSION=v1.61.0
-COVERAGE=coverage.out
+COVERAGE?=coverage.out
```
jeroenvervaeke, line 4: Coverage reported in comments on this PR is not working because this change is not in master.

I've tried it locally and changed this line in master and everything should work perfectly

fmenezes:
[q] is `PR 37.5%` the total coverage of the code at that point in time?

gssbzn (changes requested):
Couple of questions, what's the plan with the codecoverage?
Do you plan to enforce a min?
Is there a plan to bring it up?
Will you block PRs if the coverage goes down?

gssbzn (commented):
oh I see this
```
 if (( $(echo "${{ steps.compare.outputs.diff }} < 0" | bc -l) )); then
            echo "Error: Coverage difference (${{ steps.compare.outputs.diff }}%) is negative"
            exit 1
          fi
```
which would answer 
> Will you block PRs if the coverage goes down?

I strongly disagree with this, it has a tendency of inflate the code base with lower quality tests just to make these sort of checks happy 

And for example this PR wouldn't be possible https://github.com/mongodb/mongodb-atlas-cli/pull/3402, if you want to make this change let's revert that PR merge this and try that PR again :)

jeroenvervaeke:
> Couple of questions, what's the plan with the codecoverage?
The goal is to not let it degrade

> Do you plan to enforce a min?
No, for now the goal is to create awareness

> Is there a plan to bring it up?
No, there's only the goal to not make it worse, but no plan.

> Will you block PRs if the coverage goes down?
No

fmenezes:
I think this is a good change to keep up informed on coverage, having a metric is better than having none

fmenezes:
ideally we should calculate coverage with e2e tests as well

fmenezes (dismissed):
LGTM

fmenezes (dismissed):

jeroenvervaeke:
> oh I see this
> 
> ```
>  if (( $(echo "${{ steps.compare.outputs.diff }} < 0" | bc -l) )); then
>             echo "Error: Coverage difference (${{ steps.compare.outputs.diff }}%) is negative"
>             exit 1
>           fi
> ```
> 
> which would answer
> 
> > Will you block PRs if the coverage goes down?
> 
> I strongly disagree with this, it has a tendency of inflate the code base with lower quality tests just to make these sort of checks happy
> 
> And for example this PR wouldn't be possible #3402, if you want to make this change let's revert that PR merge this and try that PR again :)

I've removed the `exit 1`.

Any other things you like to see changed @gssbzn ?

gssbzn:
>Any other things you like to see changed @gssbzn ?

No that's what it, I do think we should consider a proper plan internally if we should bring coverage up otherwise the comments are a bit of noise that people will ignore

.github/workflows/code-health.yml
```diff
@@ -53,6 +55,103 @@ jobs:
...
+            trend="📉"
+          fi
+          
+          echo "Coverage Report $trend" > comment.md
+          echo "| Branch | Commit | Coverage |" >> comment.md
+          echo "|--------|--------|----------|" >> comment.md
+          echo "| ${BASE_REF} | ${{ steps.merge_base.outputs.merge_base }} | ${base_coverage}% |" >> comment.md
+          echo "| ${HEAD_REF} | ${{ github.event.pull_request.head.sha }} | ${pr_coverage}% |" >> comment.md
+          echo "|        | Difference | ${diff}% |" >> comment.md      
+          cat "comment.md" >> "$GITHUB_STEP_SUMMARY"
```
gssbzn, line 136: any reason this can't be fixed instead of using `SHELLCHECK_OPTS: --exclude=SC2129`?
https://www.shellcheck.net/wiki/SC2129

jeroenvervaeke: I disabled it for readability, I re-enabled it

gssbzn (approved):
LGTM thanks for the conversation here

PR #3453, opened 2024-12-06
CLOUDP-280747: Build a layer to handle --format

jeroenvervaeke:
## Proposed changes
- support for `--format`
    - support content type
    - support go templates
- support for `--output`
    - `--output` is required when not piping `stdout` and content type is binary
- fixed bug where all flags were set even though the user did not set them + added more tests

_Jira ticket:_ CLOUDP-280747

### DIFF
--- a/docs/command/atlas-clusters-create.txt
+++ b/docs/command/atlas-clusters-create.txt
@@ -91,7 +91,7 @@ Options
      - false
      - Major MongoDB version of the cluster.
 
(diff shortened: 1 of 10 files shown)


andreaangiolillo (approved):
LGTM

jeroenvervaeke:
> LGTM

Do you think it's fine to skip docs review here, since it's a merge from master?

PR #3481, opened 2024-12-17
feat: include output format in telemetry

gssbzn:
## Proposed changes

Include the output format in the telemetry information

### DIFF
--- a/internal/telemetry/event.go
+++ b/internal/telemetry/event.go
@@ -290,6 +290,40 @@ func withOrgID(cmd CmdFlags, c OrgIDGetter) EventOpt {
 	}
 }
 
+type OutputGetter interface {
+	Output() string
+}
+
+func withOutput(cmd CmdFlags, c OutputGetter) EventOpt {
+	return func(event Event) {
+		if cmd.Flags().Changed(flag.Output) {
+			v, _ := cmd.Flags().GetString(flag.Output)
+			event.Properties["output"] = output(v)
+			return
+		}
+		event.Properties["output"] = output(c.Output())
+	}
+}
+
+const (
+	plaintextFormat = "plaintext"
+	jsonFormat      = "json"
+	jsonPath        = "json-path"
+	goTemplate      = "go-template"
+	goTemplateFile  = "go-template-file"
+)
+
+var templateFormats = []string{jsonFormat, goTemplate, goTemplateFile, jsonPath}
+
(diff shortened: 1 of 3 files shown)


internal/telemetry/tracker.go
```diff
@@ -96,6 +96,7 @@ func (t *tracker) defaultCommandOptions() []EventOpt {
 		withService(config.Default()),
 		withProjectID(t.cmd, config.Default()),
 		withOrgID(t.cmd, config.Default()),
+		withOutput(t.cmd, config.Default()),
```
andreaangiolillo, line 99: Do we need to update the logic in the monolith too? I think the logic there sends all the properties without checking the prop name but it may be worth double-checking

andreaangiolillo (approved):
LGTM

internal/telemetry/tracker.go
gssbzn: yes, this is just a map of values for telemetry so we can always add to it

internal/telemetry/event.go
```diff
@@ -290,6 +290,40 @@ func withOrgID(cmd CmdFlags, c OrgIDGetter) EventOpt {
...
+}
+
+func withOutput(cmd CmdFlags, c OutputGetter) EventOpt {
+	return func(event Event) {
+		if cmd.Flags().Changed(flag.Output) {
+			v, _ := cmd.Flags().GetString(flag.Output)
+			event.Properties["output"] = output(v)
+			return
+		}
+		event.Properties["output"] = output(c.Output())
```
jeroenvervaeke, line 304: Is is safe to send the entire value of --output to our telemetry server?
Templates could contain PII/PHI info.

gssbzn: I'm not sending the entire value, only if you see the implementation of `output` it'll send one of ("plaintext", "json", "json-path","go-template","go-template-file")

jeroenvervaeke: My bad, I've missed that.

Most of the values you're defining below are a copy of what's defined in `output_opts.go`, is that correct?
Why not reuse those values and arrays?

gssbzn: you shouldn't export an slice because it can be reassigned  so I decided to follow 
>[A little copying is better than a little dependency.](https://www.youtube.com/watch?v=PAAkCSZUG1c&t=9m28s)

PR #3496, opened 2024-12-19
CLOUDP-289219: Create CLI commands for manipulating stream processors

auddin431:
## Proposed changes

This work adds the noun `processors` to the top level noun `streams` so that stream processors can be interacted with through the CLI. The `processors` noun has the following verbs:
- create
- delete
- list
- describe
- start
- stop

_Jira ticket:_ CLOUDP-289219

## Checklist

- [X] I have signed the [MongoDB CLA](https://www.mongodb.com/legal/contributor-agreement)
- [X] I have added tests that prove my fix is effective or that my feature works
- [X] I have updated [test/README.md](https://github.com/mongodb/mongodb-atlas-cli/blob/master/test/README.md) (if an e2e test has been added)
- [X] I have run `make fmt` and formatted my code


internal/cli/streams/processor/create_test.go
```diff
@@ -0,0 +1,154 @@
...
+		mockStore.
+			EXPECT().
+			CreateStreamProcessor(gomock.Any()).Return(expected, nil).
+			Times(1)
+
+		if err := createOpts.Run(); err != nil {
+			t.Fatalf("Run() unexpected error: %v", err)
+		}
+		t.Log(buf.String())
+		test.VerifyOutputTemplate(t, createTemplate, expected)
```
tcannon91, line 151: What does this assertion verify? It isn't clear what it actually does, we might want a comment here, especially because I _think_ it might be asserting the same thing as the next line?

tcannon91 (changes requested):
Next time, it might be a little easier to review if we break down the commits into each command or each command into their own PR :) 

Dropping comments because I think I have gone through all of the `create` command. I haven't looked but I will try not to leave the same comments on the other commands if they have the same issues.

test/e2e/atlas/data/create_streams_connection_atlas.json
tcannon91, line 1: [q] I am only seeing this used by changes from 8 months ago. Any ideas why it is being added in your PR/where it is coming from?

test/e2e/atlas/streams_with_clusters_test.go
```diff
@@ -105,6 +106,166 @@ func TestStreamsWithClusters(t *testing.T) {
...
+		req.NoError(streamsErr, string(streamsResp))
+
+		var processor atlasv2.StreamsProcessor
+		req.NoError(json.Unmarshal(streamsResp, &processor))
+
+		a := assert.New(t)
+		a.Equal(processorName, *processor.Name)
+	})
+
+	t.Run("Describe a stream processor with an atlas cluster sink", func(t *testing.T) {
```
tcannon91, line 163: All of the tests below this one concern me because they are relying on data populated/created by previous t.Run() tests. Each test should be able to function in isolation. If you want to rely on previous configuration, then it should be nested within the test.

I would recommend each of the commands `describe`, `list`, `delete`, `start`, `stop` do all of the creation/setup necessary and just don't run assertions on those steps (since the assertions are covered here). 

An alternative idea is to create a larger E2E test that tests everything. It could create, do assertions on the creation; describe, do assertions on the describe; list and assert; start and assert; stop and assert; and delete and assert.

tcannon91 (commented):
E2E tests reviewed

internal/cli/streams/processor/delete.go
```diff
@@ -0,0 +1,74 @@
...
+}
+
+// atlas streams processor delete <processorName>.
+func DeleteBuilder() *cobra.Command {
+	opts := &DeleteOpts{
+		DeleteOpts: cli.NewDeleteOpts("Atlas Stream Processor '%s' deleted\n", "Atlas Stream Processor not deleted"),
+	}
+	cmd := &cobra.Command{
+		Use:   "delete <processorName>",
+		Short: "Delete a specific Atlas Stream Processor in a Stream Processing Instance.",
```
tcannon91, line 42: [nit] 
```suggestion
		Short: "Delete an Atlas Stream Processor in a Stream Processing Instance.",
```

internal/cli/streams/processor/delete.go
```diff
@@ -0,0 +1,74 @@
...
+		Long:  fmt.Sprintf(usage.RequiredRole, "Project Read Only"),
```
tcannon91, line 43: This role doesn't seem correct.... I don't think you should be able to delete with a read only role.

internal/cli/streams/processor/delete_test.go
```diff
@@ -0,0 +1,33 @@
...
+
+import (
+	"testing"
+
+	"github.com/golang/mock/gomock"
+	"github.com/mongodb/mongodb-atlas-cli/atlascli/internal/cli"
+	"github.com/mongodb/mongodb-atlas-cli/atlascli/internal/mocks"
+)
+
+func TestDeleteOpts_Run(t *testing.T) {
```
tcannon91, line 11: There should be a `t.Run()` call in this test describing what is being tested/asserted.

internal/cli/streams/processor/describe.go
```diff
@@ -0,0 +1,93 @@
...
+		return err
+	}
+}
+
+// atlas streams processor describe <processorName>.
+func DescribeBuilder() *cobra.Command {
+	opts := &DescribeOpts{}
+	cmd := &cobra.Command{
+		Use:   "describe <processorName>",
+		Short: "Get details about a specific Atlas Stream Processor in a Stream Processing Instance.",
```
tcannon91, line 61: [nit]
```suggestion
		Short: "Get details about an Atlas Stream Processor in a Stream Processing Instance.",
```

internal/cli/streams/processor/describe.go
```diff
@@ -0,0 +1,93 @@
...
+		opts.store, err = store.New(store.AuthenticatedPreset(config.Default()), store.WithContext(ctx))
+		return err
+	}
+}
+
+// atlas streams processor describe <processorName>.
+func DescribeBuilder() *cobra.Command {
+	opts := &DescribeOpts{}
+	cmd := &cobra.Command{
+		Use:   "describe <processorName>",
```
tcannon91, line 60: fwiw, this command seems like it would make more sense to have a json output file that is the .json configuration file that would work with the create command (and eventually update).

internal/cli/streams/processor/describe.go
```diff
@@ -0,0 +1,93 @@
...
+		return err
+	}
+
+	opts.Output = jsonOutput
+	if opts.includeStats {
+		return opts.Print(r)
+	}
+
+	sp := atlasv2.NewStreamsProcessorWithStats(r.Id, r.Name, r.Pipeline, r.State)
+	return opts.Print(sp)
```
tcannon91, line 45: I think the logic is backwards here... we should be returning early `if !opts.includeStats`

internal/cli/streams/processor/describe.go
```diff
@@ -0,0 +1,93 @@
...
+	if err != nil {
+		return err
+	}
+
+	opts.Output = jsonOutput
+	if opts.includeStats {
+		return opts.Print(r)
+	}
+
+	sp := atlasv2.NewStreamsProcessorWithStats(r.Id, r.Name, r.Pipeline, r.State)
```
tcannon91, line 44: I think this should be created through a store method. I am not familiar with the decision patterns, I have only seen the CLI interact with the api through stores rather than directly like this.

internal/cli/streams/processor/describe_test.go
```diff
@@ -0,0 +1,151 @@
...
+		"dlqMessageCount":   0,
+		"inputMessageCount": 150,
+		"inputMessageSize":  500,
+	}
+
+	mockStore.
+		EXPECT().
+		StreamProcessor(gomock.Any()).
+		Return(expected, nil).
+		Times(2)
```
tcannon91, line 51: I think that separate stores should be pulled into each test to more specifically assert each test's code.

tcannon91 (changes requested):
Delete and Describe reviewed

internal/cli/streams/processor/start.go
```diff
@@ -0,0 +1,82 @@
...
+	}
+}
+
+// atlas streams processor start <processorName>.
+func StartBuilder() *cobra.Command {
+	opts := &StartOpts{}
+	cmd := &cobra.Command{
+		Use:   "start <processorName>",
+		Short: "Start a specific Atlas Stream Processor in a Stream Processing Instance.",
+		Long:  fmt.Sprintf(usage.RequiredRole, "Project Read Only"),
```
tcannon91, line 53: The docs indicate this command will require either `Project Owner` or `Project Stream Processing Owner`

internal/cli/streams/processor/start_test.go
```diff
@@ -0,0 +1,38 @@
...
+		streamsInstance: "ExampleInstance",
+		processorName:   "ExampleProcessor",
+		OutputOpts: cli.OutputOpts{
+			OutWriter: buf,
+		},
+	}
+
+	mockStore.
+		EXPECT().
+		StartStreamProcessor(gomock.Any()).
```
tcannon91, line 29: I think we should have stricter assertions on what we pass to the store calls. Splitting the commands out into separate strings might make this easier to assert on then reconstructing Params objects. But it shouldn't be too hard either way.

tcannon91 (changes requested):
Start and Stop reviewed

internal/cli/streams/processor/create.go
```diff
@@ -0,0 +1,122 @@
...
+
+	cmd.Flags().StringVar(&opts.ProjectID, flag.ProjectID, "", usage.ProjectID)
+	opts.AddOutputOptFlags(cmd)
+	cmd.Flags().StringVarP(&opts.streamsInstance, flag.Instance, flag.InstanceShort, "", usage.StreamsInstance)
+
+	cmd.Flags().StringVarP(&opts.filename, flag.File, flag.FileShort, "", usage.StreamsConnectionFilename)
+	_ = cmd.MarkFlagFilename(flag.File)
+
+	_ = cmd.MarkFlagRequired(flag.Instance)
+	_ = cmd.MarkFlagRequired(flag.File)
```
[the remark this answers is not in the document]
gssbzn: `MarkFlagRequired` returns an error if the flag doesn't exists in the set of available flags, silently ignoring the error so some, including myself prefer to be explicit about ignoring errors

internal/cli/streams/processor/create_test.go
```diff
@@ -0,0 +1,154 @@
...
+	"github.com/stretchr/testify/require"
+	atlasv2 "go.mongodb.org/atlas-sdk/v20241113004/admin"
+)
+
+func TestCreateOpts_Run(t *testing.T) {
+	t.Run("streams processor create should fail if no file is passed", func(t *testing.T) {
+		createOpts := &CreateOpts{fs: afero.NewMemMapFs()}
+		err := createOpts.Run()
+		require.Error(t, err)
+		assert.Contains(t, err.Error(), "missing file")
```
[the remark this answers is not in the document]
gssbzn: err is a go error, assert.Contains is for strings, slices and maps, you may prefer https://pkg.go.dev/github.com/stretchr/testify/assert#ErrorContains

internal/cli/streams/processor/delete_test.go
gssbzn: [nit] t.Run is for test cases, if you don't have test cases please don't add it the name of the Test is what's under test, `TestDeleteOpts_Run` tells me is test the method Run from the DeleteOpts struct, editors like intellij will even enable auto navigation between methods and test by following the naming convention for tests

test/e2e/atlas/streams_with_clusters_test.go
gssbzn: the pattern used here is a conscious decision of the CLI for E2E tests, e2e tests are expensive both in time an resources so we aim to maximize resource re utilization, e2e test are also setup to not run in parallel and rely on evergreen task splitting for isolation

gssbzn:
@auddin431 this is currently mark as stale by our bots, are you still working on this?

test/e2e/atlas/streams_with_clusters_test.go
tcannon91: @gssbzn Would my last suggestion be appropriate then; should it just be one large test? What is advantage of broken up `t.Run()`s?

internal/cli/streams/processor/delete_test.go
tcannon91: I think I agree with most of what you are saying. The information that I am missing that a `t.Run()` (or multiple) could provide is the expected behavior here. When trying to debug why this is failing in the future, I have no context as to what the expected behavior is. I could go through the assertions but it is then not clear if the assertion is no longer valid or the setup.

internal/store/streams.go
```diff
@@ -132,3 +156,39 @@ func (s *Store) DeleteConnection(projectID, tenantName, connectionName string) e
...
+}
+
+// StopStreamProcessor stops running the specified Stream Processor.
+func (s *Store) StopStreamProcessor(request *atlasv2.StopStreamProcessorApiParams) error {
+	_, _, err := s.clientv2.StreamsApi.StopStreamProcessor(s.ctx, request.GroupId, request.TenantName, request.ProcessorName).Execute()
+	return err
+}
+
+// DeleteStreamProcessor deletes the specified Stream Processor.
+func (s *Store) DeleteStreamProcessor(projectID, tenantName, processorName string) error {
```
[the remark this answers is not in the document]
auddin431: yep, completely agree and will stick to this approach. when i first started i went with the approach of passing in the request object because it was the first thing I saw in the older commands, but in retrospect i should've sticked to this in the beginning. thanks for pointing it out!

docs/command/atlas-streams-processors-create.txt
```diff
@@ -0,0 +1,118 @@
...
+     - false
+     - help for create
+   * - -i, --instance
+     - string
+     - true
+     - Name of your Atlas Stream Processing instance.
+   * - -o, --output
+     - string
+     - false
+     - Output format. Valid values are json, json-path, go-template, or go-template-file. To see the full output, use the -o json option.
```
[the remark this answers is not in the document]
auddin431: You can set the output to JSON, i.e. `-o json`. I don't think anyone would dump the `Processor <Name> created.` to a file, but IMO seeing the JSON output could be valuable as a kind of verification that the correct processor was created (i.e proper name and pipeline)

internal/cli/streams/processor/create.go
```diff
@@ -0,0 +1,122 @@
...
+				opts.initStore(cmd.Context()),
+				opts.InitOutput(cmd.OutOrStdout(), createTemplate),
+			)
+		},
+		RunE: func(_ *cobra.Command, _ []string) error {
+			return opts.Run()
+		},
+	}
+
+	cmd.Flags().StringVar(&opts.ProjectID, flag.ProjectID, "", usage.ProjectID)
```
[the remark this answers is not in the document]
auddin431: Nice! Didn't realize that existed, I'll update each command to use that instead

docs/command/atlas-streams-processors-create.txt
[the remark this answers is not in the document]
tcannon91: Ah, I think I was confused because the output section here in the docs does not indicate that it prints the processor configuration. It just says `Processor <Name> created.`

internal/cli/streams/processor/create.go
```diff
@@ -0,0 +1,122 @@
...
+		return nil, err
+	}
+
+	if opts.processorName != "" {
+		processor.Name = &opts.processorName
+	}
+
+	if opts.processorName == "" && processor.Name == nil {
+		return nil, errors.New("streams processor name missing")
+	}
```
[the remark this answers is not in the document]
auddin431: good question, i believe some of the other commands followed that pattern so i went with it. im leaning more toward your idea of forcing the name being passed in the config file

docs/command/atlas-streams-processors-create.txt
```diff
@@ -0,0 +1,118 @@
...
+   :widths: 20 10 10 60
+
+   * - Name
+     - Type
+     - Required
+     - Description
+   * - -f, --file
+     - string
+     - true
+     - Path to a JSON configuration file that defines an Atlas Stream Processing connection.
```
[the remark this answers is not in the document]
auddin431: ahh, great catch. I reused https://github.com/mongodb/mongodb-atlas-cli/blob/77b0248abfd11f1ffc7731fc2bd2a25cbe336178/internal/usage/usage.go#L241
when adding the file flag in `create.go` without realizing it was for the connection specifically 🤦 . I'll fix this and add a new variable for it

internal/cli/streams/processor/create_test.go
```diff
@@ -0,0 +1,154 @@
...
+			fs:              fs,
+			filename:        fileName,
+			streamsInstance: "ExampleInstance",
+			OutputOpts: cli.OutputOpts{
+				Template:  createTemplate,
+				OutWriter: buf,
+			},
+		}
+
+		name := "ExampleSP"
```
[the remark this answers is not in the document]
auddin431: the `atlasv2.StreamsProcessor` struct's `Name` field is a pointer to a string, so I had to define this variable so that I can populate the expected stream processor.

auddin431:
> @auddin431 this is currently mark as stale by our bots, are you still working on this?

@gssbzn yep im still working on this, was out for a while during holidays and got occupied with other tickets. Will push out requested changes in the next day or two

internal/cli/streams/processor/create.go
```diff
@@ -0,0 +1,122 @@
...
+			)
+		},
+		RunE: func(_ *cobra.Command, _ []string) error {
+			return opts.Run()
+		},
+	}
+
+	cmd.Flags().StringVar(&opts.ProjectID, flag.ProjectID, "", usage.ProjectID)
+	opts.AddOutputOptFlags(cmd)
+	cmd.Flags().StringVarP(&opts.streamsInstance, flag.Instance, flag.InstanceShort, "", usage.StreamsInstance)
```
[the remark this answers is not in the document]
auddin431: since different commands have different flags, im in favor of leaving it as explicit as possible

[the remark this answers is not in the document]
tcannon91: Gotcha, I was thinking of only including the very common things like "streamsInstance" and/or "projectId". But I am not oppose to keeping it explicit. Just want to make sure that I am clearly communicating what I was suggesting. Feel free to resolve this if you still want to leave things as explicit

internal/cli/streams/processor/create_test.go
auddin431: i believe it verifies that the output is able to follow the template. In this case, the template is `Processor {{.Name}} created.` so if we pass an object that does not contain the `Name` field, the test fails.

tcannon91: Yeah, I think that is what my tracing showed me when I looked at this last month. That seems like it is asserting the same thing as the assertion below. If this has better error messaging indicating what the problem is, maybe we should keep this in favor of the other assertion. Having both seems redundant.

test/e2e/atlas/data/create_streams_connection_atlas.json
auddin431: good catch! the e2e test generates this file dynamically here: https://github.com/mongodb/mongodb-atlas-cli/blob/77b0248abfd11f1ffc7731fc2bd2a25cbe336178/test/e2e/atlas/streams_with_clusters_test.go#L128-L133
When I ran the test locally, I forgot to clean it up so it was added to the PR as a new file

test/e2e/atlas/streams_with_clusters_test.go
auddin431: IMO the different `t.Run()`s provide a clean separation and clearly define what happens in each block

internal/cli/streams/processor/create_test.go
auddin431: I agree, having both are redundant. I'll get rid of it in favor of the more explicit assertion below it

internal/cli/streams/processor/create.go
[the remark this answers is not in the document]
auddin431: Ahh I see what you're saying now. I mistook as one generic function to combine all streams related flags. I like that suggestion but I think I'll limit specifically to the `streamsInstance` since there's a function already for projectID

internal/cli/streams/processor/delete.go
auddin431: you're right, I will update the commands with the required roles provided in the [API docs](https://www.mongodb.com/docs/atlas/reference/api-resources-spec/v2/#tag/Streams/operation/deleteStreamProcessor)

internal/cli/streams/processor/describe.go
auddin431: yeah so this seems like the case but the API call to fetch the stream processor returns stats, so if we want to include stats, we print out what the API returned. To exclude the stats, I make a new `StreamsProcessorWithStats` which has no stats populated. The reason I went with `NewStreamsProcessorWithStats` instead of `NewStreamsProcessor` is because `NewStreamsProcessor` returns a `StreamsProcessor` struct which does not include the `state` of the stream processor. Regardless of whether or not we include stats, I believe that its important to include that information. Since this is pretty confusing, I can add a comment explaining the decision

internal/cli/streams/processor/describe.go
auddin431: that's a good point, but all I'm trying to do here is create the struct. This doesn't make any API calls, so I'm unsure if it should be a store method. @gssbzn is there a best practice that I should follow here?

internal/cli/streams/processor/describe.go
auddin431: hmm, i think the output of this command would work with create depending on the marshaling (if it ignores all the extra fields it should be fine). only issue is that the API call for creating a stream processor requires a `StreamsProcessor` struct, not the `StreamsProcessorWithStats` struct, but we need the stats for the describe command

internal/cli/streams/processor/describe_test.go
auddin431: hmm, do you mean importing the actual `store.ProcessorDescriber`? im not sure how that would work because we need to mock the output of the API calls

internal/cli/streams/processor/start_test.go
auddin431: im not sure if that's entirely necessary because a lot of times, especially in the case of this test, we won't use those parameters. i might be misunderstanding what you're suggesting though

tcannon91: I am suggesting that we should asset that the projectId, tenantName, and processorName that are passed into this store function match the values passed into the cli command (or the expected values). As it stands, this does not verify the inputs do what is expected so if someone were to accidentally hardcode a tenantName in the cli command to accelerate their testing and it was missed during code review then we would ship this without knowing because the test does not assert the inputs to the cli command are passed all the way through to the store.

internal/cli/streams/processor/describe_test.go
tcannon91: No, I am suggestion that we should set up the mockStore in each `t.Run()` rather than defining it once outside the t.Run. This is an assertion that is not clearly captured in the individual test code and now we are asserting that it is called twice because we have 2 tests.

test/e2e/atlas/streams_with_clusters_test.go
tcannon91: Fair enough. Resolving

internal/cli/streams/processor/create.go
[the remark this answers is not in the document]
tcannon91: Sounds good! I think that gives us a good platform if we have other shared flags as well

internal/cli/streams/processor/delete.go
tcannon91: I haven't opened the editor to look for other examples but is there a way to specify two potential roles? I see that everything has been updated to Project Owner but all the commands seem to support Project Owner OR Project Stream Processing Owner roles

internal/cli/streams/processor/create.go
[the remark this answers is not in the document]
auddin431: i ended up sticking with this pattern because the same behavior is in the add connections command. I figure we should keep our command behavior consistent, but also not entirely opposed to changing this. Let me know what you think

internal/cli/streams/processor/delete.go
auddin431: So here's what the RequiredRole is defined as: https://github.com/mongodb/mongodb-atlas-cli/blob/77b0248abfd11f1ffc7731fc2bd2a25cbe336178/internal/usage/usage.go#L337
I suppose I could add a new variable `RequiredRoles` that would go something like:
```go
RequiredRoles = "To use this command, you must authenticate with a user account or an API key with the %s or %s role."
```
I'm a little uncertain about this because what if we add a command that can use one of three roles? or four? Alternatively, we can write a Streams specific variable for usage and it can live in the `internal/cli/streams` package or something

internal/cli/streams/processor/create.go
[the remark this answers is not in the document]
tcannon91: I think it would be good to get product to weigh in (if you haven't done this already). Or possibly someone from apix/the cli team. If it is intentional behavior we also need to ensure that we always treat the CLI option with higher priority (I am guessing that is the expected behavior at least).

internal/cli/streams/processor/delete.go
tcannon91: We could do something like create a new string format like this:
```
RequiredOneOfRoles = "To use this command, you must authenticate with a user account or an API key with any of the following roles: %s.";
```

then use a `String.join(", ", ...roleList);` and pass that as the string variable.

test/e2e/atlas/data/create_streams_connection_atlas.json
auddin431: Hmm, maybe instead of adding to gitignore we can do a better cleanup within the tests. My guess is that since these E2E tests run in Evergreen, the environment gets wiped at the end of the test so there wasn't much consideration for locally running the tests

auddin431:
evergreen keep-definitions

internal/cli/streams/processor/create.go
```diff
@@ -0,0 +1,124 @@
...
+	createParams := new(atlasv2.CreateStreamProcessorApiParams)
+	createParams.GroupId = opts.ProjectID
+	createParams.TenantName = opts.Instance
+	createParams.StreamsProcessor = processor
+
+	return createParams, nil
+}
+
+// atlas streams processor create [--projectId projectId].
+func CreateBuilder() *cobra.Command {
```
gssbzn, line 86: does this command need a `--watch` as in are the changes applied on create or does it take a while to be effective? see https://github.com/mongodb/mongodb-atlas-cli/blob/97d3e3b7915c514e88be69ce1f1852ef2c503ccf/internal/cli/clusters/create.go#L55 for the use of watch on other commands

internal/cli/streams/processor/delete.go
```diff
@@ -0,0 +1,87 @@
...
+func (opts *DeleteOpts) initStore(ctx context.Context) func() error {
+	return func() error {
+		var err error
+		opts.store, err = store.New(store.AuthenticatedPreset(config.Default()), store.WithContext(ctx))
+		return err
+	}
+}
+
+func (opts *DeleteOpts) Run() error {
+	return opts.Delete(opts.store.DeleteStreamProcessor, opts.ProjectID, opts.Instance)
```
gssbzn, line 46:
```suggestion
	return opts.Delete(opts.store.DeleteStreamProcessor, opts.ConfigProjectID(), opts.Instance)
```
here and everywhere else you do `opts.ProjectID`, project id is layered, it can come from config, env or flag, using th emethod `ConfigProjectID()` will make sure to use the correct one other wise you are looking only to the flag values

internal/cli/streams/processor/delete_test.go
gssbzn: My default assumption when no `t.Run` is the test makes a "happy path" assumption, which most of this `_Run` test are, they mostly just see if you call the correct API or not

internal/cli/streams/processor/describe.go
gssbzn: the store now days is mostly a legacy artifact from days when we needed to deal with cloud manager, atlas and ops manager APIs so it made easy to abstract what API the command was dealing with and as such we used to delegate api object creations to the store, in tthe current context, no strong opinion if you create the object in the comamnd or encapsulate on the store

internal/cli/streams/processor/create.go
auddin431: Hm, I know cluster creation can be a very lengthy process so I see why it's necessary for the `--watch`, but this command takes about 5-10 seconds. Is there a general cutoff as to when `--watch` is necessary?

docs/command/atlas-streams-processors-create.txt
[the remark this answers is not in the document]
tcannon91: I am curious if you have thoughts here. Is there a way to update the output to more accurately reflect what the output is?

tcannon91 (commented):
Looks pretty good

docs/command/atlas-streams-processors-create.txt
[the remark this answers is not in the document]
auddin431: I'm not entirely sure if we can. We could forcefully output JSON like some of the other commands (like `describe`) but there would still be the issue of not having the most informative output statement for JSON like we have here

tcannon91 (approved):
Looks good

internal/cli/streams/processor/create.go
gssbzn: I guess the main rule would be created vs ready for use, customers that script with CLI or other tools want to know when is ready for use and for that reason we provide `--watch`

internal/cli/streams/processor/create_test.go
```diff
@@ -0,0 +1,167 @@
...
+			// gomock.Any() is necessary for the *atlasv2.StreamProcessor argument because newCreateRequest()
+			// creates a new *atlasv2.StreamProcessor with the same data but different memory address, causing
+			// the equality comparison to fail
+			CreateStreamProcessor(gomock.Eq(createOpts.ConfigProjectID()), gomock.Eq(createOpts.Instance), gomock.Any()).
+			Return(expected, nil).
+			Times(1)
+
+		if err := createOpts.Run(); err != nil {
+			t.Fatalf("Run() unexpected error: %v", err)
+		}
```
gssbzn, line 164: since using `testify` to keep it consistent 
```suggestion
		require.NoError(createOpts.Run())
```

internal/cli/streams/processor/describe.go
```diff
@@ -0,0 +1,102 @@
...
+func (opts *DescribeOpts) Run() error {
+	r, err := opts.store.StreamProcessor(opts.ConfigProjectID(), opts.Instance, opts.processorName)
+	if err != nil {
+		return err
+	}
+
+	opts.Output = jsonOutput
+	if opts.includeStats {
+		return opts.Print(r)
+	}
```
gssbzn, line 51: why overloading output options with your own option?

internal/cli/streams/processor/list.go
```diff
@@ -0,0 +1,95 @@
...
+	r, err := opts.store.ListProcessors(opts.ConfigProjectID(), opts.Instance)
+	if err != nil {
+		return err
+	}
+
+	sps := make([]atlasv2.StreamsProcessorWithStats, 0, len(*r.Results))
+	opts.Output = jsonOutput
+	if opts.includeStats {
+		return opts.Print(*r.Results)
+	}
```
gssbzn, line 50: same question? why is this a custom way to output instead of the standard `--out|-o`?

internal/cli/streams/processor/list.go
```diff
@@ -0,0 +1,95 @@
...
+
+	for _, res := range *r.Results {
```
gssbzn, line 52: don't use pointers, use getters https://github.com/mongodb/atlas-sdk-go/blob/main/docs/doc_5_best_practices.md#using-getters-instead-of-direct-field-access
```suggestion
	for _, res := range r.GetResults() {
```

gssbzn, line 52: why are you iterating for the print? why don't you define a template that can handle the envelop object?

gssbzn (changes requested):
I'm finding a bit confusing that some commands don't follow the general output logic of the CLI

auddin431:
Decided to close this PR as we are fine waiting on command autogeneration to complete cc @gssbzn

PR #3529, opened 2025-01-09
CLOUDP-293144: Removes k8s commands

cveticm:
## Proposed changes

Removes kubernetes commands `operator install`, `config generate`, `config apply`.
All methods only used by these commands have also been removed.

Docs and mocks have been regenerated to reflect these removals.

Evergreen tasks remain and will be altered as necessary during [future work](https://jira.mongodb.org/browse/CLOUDP-293169).

_Jira ticket:_ [CLOUDP-293144](https://jira.mongodb.org/browse/CLOUDP-293144)

## Checklist


.github/workflows/code-health.yml
```diff
@@ -157,25 +157,6 @@ jobs:
             echo "Error: Coverage difference (${{ steps.compare.outputs.diff }}%) is negative"
           fi
 
-  fuzz-tests:
```
cveticm, line 160: Removing fuzz test as all fuzz tests are tests for k8s commands

docs/command/atlas.txt
```diff
@@ -68,7 +68,6 @@ Related Commands
 * :ref:`atlas-events` - Manage events for your organization or project.
 * :ref:`atlas-federatedAuthentication` - Manage Atlas Federated Authentication.
 * :ref:`atlas-integrations` - Configure third-party integrations for your Atlas project.
-* :ref:`atlas-kubernetes` - Manage Kubernetes resources.
```
cveticm, line 71: Noting that removal of k8s commands results in the removal of mention of the command in docs.

build/ci/library_owners.json
```diff
@@ -35,23 +35,13 @@
   "go.mongodb.org/atlas": "apix-2",
   "go.mongodb.org/atlas-sdk/v20241113004": "apix-2",
   "go.mongodb.org/atlas-sdk/v20240530005": "apix-2",
-  "go.mongodb.org/atlas-sdk/v20241113001": "apix-2",
```
cveticm, line 38: Removing library owner entries for libraries which have been removed

internal/cli/root/builder.go
```diff
@@ -244,7 +243,6 @@ Use the --help flag with any command for more info on that command.`,
 		logoutCmd,
 		whoCmd,
 		registerCmd,
-		kubernetes.Builder(),
```
cveticm, line 247: Removes k8s from CLI builder.

test/e2e/atlas/atlas_e2e_test_generator_test.go
```diff
@@ -36,23 +36,15 @@ const (
```
cveticm, line 36: Removed methods and variables no longer in use after removal of k8s commands

test/e2e/atlas/helper_test.go
```diff
@@ -169,53 +169,6 @@ func splitOutput(cmd *exec.Cmd) (string, string, error) {
 	return o.String(), e.String(), err
 }
```
cveticm, line 171: Removed methods no longer in use after removal of k8s commands

docs/command/atlas.txt
blva: I think it will be removed from this file and then added back via copybara integration, right?

blva (approved):
LGTM - Removal in the feature branch only.

Notes:
1. no need for docs review at this point
2. would be good to get one more LGTM from APIx
3. would be good to get one LGTM from K8s - K8s, can you confirm this is all the code you expect to live in the plugin, is there any other kubernetes functionality we are missing? (whatever we're not deleting)

RafaelBroseghini:
how will k8s resource(s) importing work going forward?

blva:
> how will k8s resource(s) importing work going forward?
hi @RafaelBroseghini , the code is being moved to a separate code base in https://github.com/mongodb/atlas-cli-plugin-kubernetes and will be used as a 1st class plugin. Users will have the same experience as of today, the only difference is that the plugin will have its own binary and release cadence.

docs/command/atlas.txt
cveticm: I don't think this would be the case since the copybara logic is to replace all doc files in AtlasCLI to DocAtlasCLI. Theres no other modifications applied to files. On k8s plugin side, the only docs it holds are k8s docs and so won't effect atlas.txt.

cveticm: Maybe the copybara workflow in AtlasCLI could modify atlas.txt by including a `core.replace` transformation which replaces 
```
* :ref:`atlas-integrations` - Configure third-party integrations for your Atlas project.
```` 
with 
```
* :ref:`atlas-integrations` - Configure third-party integrations for your Atlas project.
* :ref:`atlas-kubernetes` - Manage Kubernetes resources.
```

Or perhaps the gen-docs logic could install all first class plugins to find and add their root command (in this case `kubernetes`) in the atlas.txt doc.

Comment (blva) [reply docs/command/atlas.txt:71]: 2 good options, I suggest creating a JIRA ticke tin the epic so that we don't miss this update!

File: build/ci/library_owners.json [hunk 7]
```diff
@@ -35,23 +35,13 @@
...
   "go.mongodb.org/mongo-driver": "apix-2",
   "golang.org/x/sys": "apix-2",
   "golang.org/x/tools": "apix-2",
   "google.golang.org/api": "apix-2",
   "google.golang.org/protobuf": "apix-2",
   "golang.org/x/mod": "apix-2",
   "gopkg.in/yaml.v3": "apix-2",
-  "github.com/mongodb/mongodb-atlas-kubernetes/v2": "atlas_kubernetes_team",
```
andreaangiolillo, line 46: We don’t remove commands lightly. Our standard process involves first deprecating the commands and displaying a deprecation warning in stderr for a period determined by the PM. Once the deprecation period ends, we remove the command in follow up release. Could you clarify why this process is not being followed in this case?

igor-karpukhin (approved):

build/ci/library_owners.json
cveticm: Since the release in which these commands are removed from the AtlasCLI repo directly follows the release of the k8s plugin. which contains all of the code that is being removed in this PR, the UX experience will not change at all. The code has just been migrated. Thus, I don't think a deprecation period is required since commands will still be available to customers with same functionality as now.

blva: yep, as per scope we proposed to **replace** them: https://docs.google.com/document/d/1t1uW-O4l_LqdhDcDWIIpLE2qTGUAfVlfPQve0BfPmk4/edit?tab=t.0#heading=h.my8l8kwcujpo

andreaangiolillo: >  the UX experience will not change at all.

I don't agree with this, customers running the CLI in a closed network environment (without access to the GH repo) will start getting an error which is a **breaking change**

PR #3575, opened 2025-01-24
CLOUDP-295015: [Local Atlas Experience] Add support for private Docker registries

jeroenvervaeke:
## Proposed changes
Look for a custom "Atlas local development" image in the following places and order:
1. `ATLAS_LOCAL_DEPLOYMENT_IMAGE` environment variable
2. `local_deployment_image` profile setting
3. Fallback to hardcoded `docker.io/mongodb/mongodb-atlas-local` (can still be overridden by build)

_Jira ticket:_ CLOUDP-295015

### DIFF
--- a/internal/cli/deployments/options/deployment_opts.go
+++ b/internal/cli/deployments/options/deployment_opts.go
@@ -19,6 +19,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"os"
 	"regexp"
 	"sort"
 	"strings"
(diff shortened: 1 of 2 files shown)


fmenezes (approved):

fmenezes:
[q] is the profile setting supposed to be a global setting or per profile?

jeroenvervaeke:
Moved the config to global properties as suggested

fmenezes (approved):

internal/cli/deployments/options/deployment_opts.go
```diff
@@ -130,8 +131,23 @@ func (opts *DeploymentOpts) LocalMongodHostname() string {
 
 var LocalDevImage = "docker.io/mongodb/mongodb-atlas-local"
 
+func getLocalDevImage() string {
+	// First check environment variable
+	if envImage := os.Getenv("ATLAS_LOCAL_DEPLOYMENT_IMAGE"); envImage != "" {
```
gssbzn, line 136: if you use the standard prefix `MONGODB_ATLAS` as in `MONGODB_ATLAS_LOCAL_DEPLOYMENT_IMAGE` then the config can automatically do the checks to read from env or the config file

gssbzn: and I strongly think even if not relating to the features int he config you should follo wthe standard prefix to avoid confusion

tibulca: agree with using the standard prefix https://github.com/mongodb/mongodb-atlas-cli/blob/2224a16135637ec08ddb23f5db6375abe5623270/internal/config/profile.go#L42

jeroenvervaeke: Will address the changes in a future PR

gssbzn: we don't really read variables in the profile because we "cheat" by relying on viper, check https://github.com/spf13/viper?tab=readme-ov-file#working-with-environment-variables, the ones you listed are more the exception because they usually come from external tools and we use them for telemetry

PR #3585, opened 2025-01-28
CLOUDP-292675: Autogeneration: create e2e tests

jeroenvervaeke:
## Proposed changes
Created a few smoke-tests to see if auto-generated commands are working.
Most of the auto-generated code is covered by unit tests.

_Jira ticket:_ CLOUDP-292675


test/e2e/atlas/autogenerated_commands_test.go
```diff
@@ -0,0 +1,80 @@
+//go:build e2e || (atlas && autogeneration)
```
fmenezes, line 1: missing license header

jeroenvervaeke: Added

internal/cli/api/api.go
```diff
@@ -102,10 +102,15 @@ func convertAPIToCobraCommand(command api.Command) (*cobra.Command, error) {
...
 			// - set orgId
 			// - set projectId
+			// - default api version
 			if err := setUnTouchedFlags(NewProfileFlagValueProviderForDefaultProfile(), cmd); err != nil {
 				return errors.Join(ErrFailedToSetUntouchedFlags, err)
 			}
 
+			// Reset version to default if unsupported version was selected
+			// This can happen when the profile contains a default version which is not supported for a specific endpoint
+			ensureVersionIsSupported(command, &version)
```
blva, line 112: [q] should we do this or throw an error for the customer to use right version? 🤔

fmenezes (approved):
LGTM

internal/cli/api/api.go
fmenezes: I think this is preferred 'cause one date can be valid for one resource and invalid for another

jeroenvervaeke: I'm not a big fan of the option to add `api_version` to the profile because it causes the issues @blva described. But this is the best way to deal with it.

If we removed the option to add the `api_version` from your profile, I would 100% return an error.

We also can't just forward the version without checking because we don't know the request's content type. This is part of the version. We also want to avoid putting any versioning logic in the CLI (the logic that's present to select a version on the backend)

blva: ohhhh got it! default here means if someone has set a version in their profile, it **does not mean** you'll find what's the latest API version, right? I misunderstood and thought you were doing the latter option, in that case I think it looks good.

jeroenvervaeke: verifying with product, thanks for bringing it up @blva

PR #3590, opened 2025-01-30
CLOUDP-221450: Automation generation of AtlasCLI commands

jeroenvervaeke:
## Proposed changes
Added pinning reminders

_Jira ticket:_ CLOUDP-221450


internal/cli/api/api.go
```diff
@@ -329,6 +332,22 @@ func defaultAPIVersion(command api.Command) (string, error) {
 	return lastVersion.Version, nil
 }
 
+func remindUserToPinVersion(cmd *cobra.Command) {
+	versionFlag := cmd.Flag(flag.Version)
```
gssbzn, line 336:
```suggestion
	versionFlag := cmd.Flags().Changed(flag.Version)
```

internal/cli/api/api.go
```diff
@@ -329,6 +332,22 @@ func defaultAPIVersion(command api.Command) (string, error) {
...
+	// if we fail to get the version flag (which should never happen), then quit
+	if versionFlag == nil {
```
gssbzn, line 338:
```suggestion
	if !versionFlag {
```

internal/cli/api/api.go
jeroenvervaeke: Thanks for the suggestion, but you missed that I also use the flag to get the value a few lines below `versionFlag.Value.String()`

internal/cli/api/api.go
jeroenvervaeke: See [comment above](https://github.com/mongodb/mongodb-atlas-cli/pull/3590#discussion_r1935503660)

gssbzn (approved):

PR #3616, opened 2025-02-05
CLOUDP-291620: Add streams privateLink create command

tcannon91:
## Proposed changes
Add the ability for users to create PrivateLink endpoints that can be used by atlas stream processors. The configuration for the PrivateLink endpoint will be managed in a .json file, which will be read into the configuration struct.

_Jira ticket:_ CLOUDP-291620

## Checklist

### DIFF
--- a/internal/cli/streams/privatelink/create_test.go
+++ b/internal/cli/streams/privatelink/create_test.go
@@ -57,6 +57,7 @@ func TestCreateOpts_Run(t *testing.T) {
 
 	for _, tc := range testCases {
 		t.Run(tc.name, func(t *testing.T) {
+			t.Parallel()
 			fs := afero.NewMemMapFs()
 
 			if tc.fileContents != "" {


internal/cli/streams/privatelink/create_test.go
```diff
@@ -0,0 +1,136 @@
+// Copyright 2023 MongoDB Inc
```
jasonrydberg, line 1: [q] Were all of the headers autogenerated? A few have 2025 but others 2023.

jasonrydberg (changes requested):

jasonrydberg (dismissed):
LGTM

internal/cli/streams/privatelink/create_test.go
```diff
@@ -0,0 +1,136 @@
...
+	"github.com/spf13/afero"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	atlasv2 "go.mongodb.org/atlas-sdk/v20241113004/admin"
+)
+
+const fileName = "test-privateLink.json"
+
+func TestCreateOpts_Run(t *testing.T) {
+	validPrivateLinkConfigFileContents := `
```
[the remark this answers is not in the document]
tcannon91: I think I can combine 2 out of the 4 to leverage a matrix. The other two have pretty unique assertions. If it is desirable to wrap those assertions in a function that is part of the test matrix setup I can do that, but I don't find that level of abstraction helpful. Let me know how far you want me to take it and I can get it adjusted :)

carriecwk (approved):
LGTM

internal/cli/streams/privatelink/create_test.go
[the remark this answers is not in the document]
gssbzn: >The other two have pretty unique assertion

I would then split those in independent Test as `func TestCreateOpts_Run_specialCase(t *testing.T) {` but this is not a blocker but keeps test isolation better when one case may start failing and may not be related others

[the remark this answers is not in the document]
tcannon91: I polled my team and it seems that the preference is for us to leverage t.Run()s within a test func rather than breaking out into separate functions. I will combine the applicable test cases to a test matrix :)

jasonrydberg (approved):

internal/cli/streams/privatelink/create.go
```diff
@@ -0,0 +1,110 @@
...
+}
+
+func (opts *CreateOpts) Run() error {
+	privateLinkEndpoint := atlasv2.NewStreamsPrivateLinkConnection()
+	if err := file.Load(opts.fs, opts.filename, privateLinkEndpoint); err != nil {
+		return err
+	}
+
+	// Remaining validation will be done by the API
+	if len(privateLinkEndpoint.GetProvider()) == 0 {
```
gssbzn, line 51: this should be safe now that you use getters
```suggestion
	if privateLinkEndpoint.GetProvider() == "" {
```

internal/cli/streams/privatelink/create_test.go
```diff
@@ -0,0 +1,149 @@
...
```
gssbzn, line 33: move closer to where is used which is after the matrix test cases

internal/cli/streams/privatelink/create_test.go
```diff
@@ -0,0 +1,149 @@
...
+				}
+			`,
+			wantErr: func(tt require.TestingT, err error, _ ...any) {
+				require.ErrorContains(tt, err, "provider missing")
+			},
+		},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
```
gssbzn, line 68:
```suggestion
		t.Run(tc.name, func(t *testing.T) {
			t.Parallel()
```

internal/cli/streams/privatelink/create_test.go
```diff
@@ -0,0 +1,149 @@
...
+			fs := afero.NewMemMapFs()
+
+			if len(tc.fileContents) > 0 {
```
gssbzn, line 71: as above, for strings is usually preffered to check for empty string (tho the compiler generates the same code)
```suggestion
			if tc.fileContents != "" {
```

internal/cli/streams/privatelink/create_test.go
```diff
@@ -0,0 +1,149 @@
...
+		expectedProvider := "Azure"
+		expectedRegion := "US_EAST_2"
+		expectedServiceEndpointID := "/subscriptions/fd01adff-b37e-4693-8497-83ecf183a145/resourceGroups/test-rg/providers/Microsoft.EventHub/namespaces/test-namespace"
+		expectedDNSDomain := "test-namespace.servicebus.windows.net"
+
+		expected := atlasv2.NewStreamsPrivateLinkConnection()
+		expected.Provider = &expectedProvider
+		expected.Region = &expectedRegion
+		expected.ServiceEndpointId = &expectedServiceEndpointID
+		expected.DnsDomain = &expectedDNSDomain
```
gssbzn, line 106: you never read the "expected values so just use setters to avoid pointer logic 
```suggestion
	expected := atlasv2.NewStreamsPrivateLinkConnection()
	expected.SetProvider("Azure")
	expected.SetRegion("US_EAST_2")
	expected.SetServiceEndpointId("/subscriptions/fd01adff-b37e-4693-8497-83ecf183a145/resourceGroups/test-rg/providers/Microsoft.EventHub/namespaces/test-namespace")
	expected.SetDnsDomain("test-namespace.servicebus.windows.net")
```

internal/cli/streams/privatelink/create_test.go
```diff
@@ -0,0 +1,149 @@
...
+
+		mockStore.
+			EXPECT().
+			// This test does not assert the parameters passed to the store method
+			CreatePrivateLinkEndpoint(gomock.Any(), gomock.Any()).
+			Return(expected, nil).
+			Times(1)
+
+		require.NoError(t, createOpts.Run())
+		assert.Equal(t, "Atlas Stream Processing PrivateLink endpoint "+expectedInterfaceEndpointID+" created.\n", buf.String())
```
gssbzn, line 147: this can for sure be part of the test matrix

internal/cli/streams/privatelink/privatelink.go
```diff
@@ -0,0 +1,40 @@
...
+	"github.com/spf13/cobra"
+)
+
+var commandRoles = strings.Join([]string{"Project Owner", "Project Stream Processing Owner"}, ", ")
+
+func Builder() *cobra.Command {
+	const use = "privateLink"
+	cmd := &cobra.Command{
+		Use:     use,
+		Aliases: cli.GenerateAliases(use),
```
gssbzn, line 30: assuming you can "list" private links we default to plurals and ` cli.GenerateAliases` can generate the singular 
```suggestion
	const use = "privateLinks"
	cmd := &cobra.Command{
		Use:     use,
		Aliases: cli.GenerateAliases(use),
```

internal/cli/streams/privatelink/create_test.go
tcannon91: These tests are pretty lightweight, is this necessary? From my understanding, using t.Parallel() "by default" isn't the typical pattern.

internal/cli/streams/privatelink/create_test.go
tcannon91: I am running into issues bringing this up into the test matrix function because the mock store does not get called so the assertions (even though they are weak and I don't care about them) fail, but I need them to force a store return value that I can consistently assert against for this string where this comment is. Adding conditionals to my test matrix to have different branches within my test defeats the purpose of a test matrix in my opinion so I think leaving this as a separate test makes the most sense.

internal/cli/streams/privatelink/create_test.go
gssbzn: as a rule of thumb in test matrix I find it to be better to add it than not, it may be the case the matrix will grow and it usually helps catch concurrency issues on the test or function, there's also https://github.com/kunwardeep/paralleltest which has been long on my list of linters to enable and it flags test matrix as needing `t.Parallel `

internal/cli/streams/privatelink/create_test.go
gssbzn: yeah if too hard no worries but was worth flagging that output check could be part of a matrix

internal/cli/streams/privatelink/create_test.go
tcannon91: Cool! I am sold, will add :)

gssbzn (approved):
LGTM

jasonrydberg (approved):
LGTM

PR #3637, opened 2025-02-12
CLOUDP-297411: Implement --watch support

jeroenvervaeke:
## Proposed changes
Added `--watch` support for autogenerated commands.

Introduced two watcher structs:
- `api.Watcher`: Takes care of executing the API call and verifying the output
- `cli.Watcher`: Takes care of calling `api.Watcher` in a loop, cancellation, showing spinner

Changes to Executor:
- I had to move formatting out of the executor because we need to feed the original request's unformatted response to the watcher.
- I only save the response to memory when a watcher is enabled; if not, I pass the HTTP reader to the formatted without storing the response in memory.
- Executing a command now also exposes the HTTPCode so that we can watch the HTTP code

Testing:
- I added new unit tests for most functions, which should cover most cases
- I tested manually by adding the watcher configuration for `DeleteCluster` and running the CLI manually (reverted before committing)

_Jira ticket:_ CLOUDP-297411

### DIFF
--- a/internal/api/watcher.go
+++ b/internal/api/watcher.go
@@ -2,15 +2,23 @@ package api
 
 import (
 	"context"
+	"encoding/json"
 	"errors"
 	"fmt"
+	"io"
+	"reflect"
+	"slices"
 	"time"
+
+	"github.com/PaesslerAG/jsonpath"
 )
 
 var (
-	ErrWatcherFailedToBuildRequest        = errors.New("failed to build watcher request")
-	ErrWatcherFailedToWatch               = errors.New("failed to watch")
-	ErrWatcherFailedToExecuteWatchRequest = errors.New("failed to execute watch request")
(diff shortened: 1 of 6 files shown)


internal/api/watcher.go
```diff
@@ -0,0 +1,235 @@
...
+	return &Watcher{
+		executor: executor,
+		request:  request,
+		expect:   props.Expect,
+	}, nil
+}
+
+// Builds the request we're going to send over and over again until it returns what we expect
+//
+// - `allCommands`, array with all command definitions, realisticly this will always be the static `Commands` variable, but this allows easier unit testing
```
cveticm, line 81: Minor nit! 
```suggestion
// - `allCommands`, array with all command definitions, realistically this will always be the static `Commands` variable, but this allows easier unit testing
```

cveticm (dismissed):
minor spelling nit in comments, otherwise lgtm 👍

cveticm (approved):
LGTM 👍

PR #3646, opened 2025-02-13
task: fix go mod grouping

gssbzn:
## Checklist

Seems like with some recent changes some dependencies that used to be indirect were made direct but we never moved them around for easy finding

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,8 +1,6 @@
 module github.com/mongodb/mongodb-atlas-cli/atlascli
 
-go 1.23.5
-
-toolchain go1.23.6
+go 1.23.6
 
 require (
 	cloud.google.com/go/kms v1.20.5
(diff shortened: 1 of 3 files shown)


go.mod
```diff
@@ -1,8 +1,6 @@
 module github.com/mongodb/mongodb-atlas-cli/atlascli
 
-go 1.23.5
-
-toolchain go1.23.6
```
gssbzn, line 5: if they both define a full version you don't need both, you only need toolchain if you define go like `go 1.23`

internal/cli/plugin/plugin_github_asset.go
```diff
@@ -293,19 +291,15 @@ func (g *GithubAsset) verifyAssetSignature(asset []byte, sigAssetID, pubKeyAsset
 	return nil
 }
 
-func parseGithubReleaseValues(arg string) (*GithubAsset, error) {
-	regexPattern := `^((https?://(www\.)?)?github\.com/)?(?P<owner>[\w.\-]+)/(?P<name>[\w.\-]+)/?(@(?P<version>.+))?$`
-	regex, err := regexp.Compile(regexPattern)
-	if err != nil {
-		return nil, fmt.Errorf("error compiling regex: %w", err)
-	}
+var ghReleaseReg = regexp.MustCompile(`^((https?://(www\.)?)?github\.com/)?(?P<owner>[\w.\-]+)/(?P<name>[\w.\-]+)/?(@(?P<version>.+))?$`)
```
gssbzn, line 294: drive-by, regex should be compiled once if possible because is an expensive operation

blva (approved):

PR #3694, opened 2025-03-03
[tests] remove references to MCLI

gssbzn:
Remove references to MCLI which was the legacy namespace from the monoglci days, and update to the Atlas documented one `MONGODB_ATLAS`


fmenezes (approved):
LGTM, make sure all changed tests are still passing

gssbzn:
All green, also thanks cause I kind of forgot about this one

PR #3696, opened 2025-03-03
[task] update golangci-lint to 1.64.6

gssbzn:
Was having some issues with the old linter locally, tried to update to fix them and it works so let's keep it on latest

### DIFF
--- a/internal/cli/datafederation/logs_test.go
+++ b/internal/cli/datafederation/logs_test.go
@@ -20,7 +20,7 @@ package datafederation
 
 import (
 	"io"
-	"os"
+	"strings"
 	"testing"
 
 	"github.com/golang/mock/gomock"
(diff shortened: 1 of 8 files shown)


internal/cli/datafederation/logs_test.go
```diff
@@ -34,24 +34,12 @@ func TestLogOpts_Run(t *testing.T) {
 	ctrl := gomock.NewController(t)
 	mockStore := mocks.NewMockDataFederationLogDownloader(ctrl)
 
-	const contents = "expected"
-
-	file, err := os.CreateTemp("", "")
-	if err != nil {
-		require.NoError(t, err)
```
gssbzn, line 41: unit tests shouldn't be writing things to disk 🤔

andreaangiolillo (approved):
LGTM

PR #3698, opened 2025-03-03
CLOUDP-303346: Add support for parameter aliases in OpenAPI spec

jeroenvervaeke:
## Proposed changes
- Add support for parameter aliases in OpenAPI spec.
- Added overlay to add `projectId` alias to all `groupId` parameters.
- Flyby, improved `extractExtensionsFromOperation` to return `operationExtensions` instead of a tuple.

_Jira ticket:_ CLOUDP-303346


tools/api-generator/fixtures/06-spec-with-cli-specific-extensions.yaml
```diff
@@ -338,6 +344,10 @@ components:
       description: Flag that indicates whether the response body should be in the <a href="https://en.wikipedia.org/wiki/Prettyprint" target="_blank" rel="noopener noreferrer">prettyprint</a> format.
       in: query
       name: pretty
```
gssbzn, line 346: does pretty have any meaning for the CLI?

tools/api-generator/fixtures/06-spec-with-cli-specific-extensions.yaml
```diff
@@ -338,6 +344,10 @@ components:
...
+      x-xgen-atlascli:
+        flag-short: p
```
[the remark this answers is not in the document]
jeroenvervaeke: This is a fixture in a unit test, I don't think the actual values are relevant here.

tools/api-generator/fixtures/06-spec-with-cli-specific-extensions.yaml
jeroenvervaeke: This is a fixture in a unit test, I don't think the actual values are relevant here.

gssbzn: as in isn't the CLI always pretty?

gssbzn: oh interesting seems like no after trying thi s

tools/api-generator/convert.go
```diff
@@ -273,6 +267,47 @@ func extractParametersNameDescription(parameterRef *openapi3.ParameterRef) (stri
...
+	short   string
+}
+
+func extractParameterExtensions(parameterRef *openapi3.ParameterRef) parameterExtensions {
+	ext := parameterExtensions{
+		aliases: []string{},
+		short:   "",
+	}
+
+	getExtensionValues := func(extensionsMap map[string]any) {
```
gssbzn, line 281: any particular reason for an anonymous function? doesn't seem to be used as a callback or similar pattern of passing as an arg

tools/api-generator/commands.go.tmpl
```diff
@@ -43,6 +43,7 @@ var Commands = GroupedAndSortedCommands{
 								IsArray:	{{ .Type.IsArray }},
 								Type:		`{{ .Type.Type }}`,
 							},
+							Aliases: []string{ {{ range $i, $alias := .Aliases }}{{if $i}},{{end}}`{{ $alias }}`{{end}} },
```
gssbzn, line 46: can this be 
```suggestion
							{{if .Aliases }} Aliases: []string{ {{ range $i, $alias := .Aliases }}{{if $i}},{{end}}`{{ $alias }}`{{end}} }, {{end}}
```

tools/api-generator/fixtures/06-spec-with-cli-specific-extensions.yaml
jeroenvervaeke: But to reply to your question, it does not have a meaning to the CLI; it's just a regular parameter passed to the backend, and the CLI doesn't handle it any differently.

tools/api-generator/commands.go.tmpl
jeroenvervaeke: I was hesitating between what I have and what you suggested.
I've updated the code.

tools/api-generator/convert.go
gssbzn: that's probably a tell we may need better code organization, in lining the function because trying to name it causes confusing with other functions sound like a good time to think if this package needs some cleaning

tools/api-generator/commands.go.tmpl
gssbzn: I still se a couple of empty slices initialization which I was trying to avoid

jeroenvervaeke: That's fixed now. I blindly accepted the suggestion, but I forgot that there's a second piece of code that does the same thing.

tools/api-generator/overlays/groupID_flag.yaml
```diff
@@ -0,0 +1,10 @@
+overlay: 1.0.0
+info:
+  title: Configure groupId flag
+  version: 1.0.0
+actions:
+  - target: $.components.parameters.groupId
+    update:
+      x-xgen-atlascli:
+        aliases: 
+          - projectId
```
fmenezes, line 10: nit, it works the way it is, but currently, we have a small gap we can't provide the short version of the alias (we only provide --projectId but we don't have a way to provide a hypothetical -P)

fmenezes (dismissed):
LGTM

gssbzn (dismissed):
Thanks for the changes

tools/api-generator/overlays/groupID_flag.yaml
jeroenvervaeke: Good suggestion, I'll make the changes!

tools/api-generator/convert.go
jeroenvervaeke: I've reorganized the code to get rid of the anonymous function

tools/api-generator/overlays/groupID_flag.yaml
jeroenvervaeke: Updated

gssbzn (changes requested):
-P is a shorhand for profile

gssbzn (approved):
LGTM

PR #3706, opened 2025-03-04
task: improve autogeneration integration with other tools

gssbzn:
## Proposed changes

Improve the generate code to leverage tools conventions for detecting autogenerated code 

Screenshot 2025-03-04 at 11 18 42

As a side note I find a bit confusing I need to run `make bin/api-generator` if I do changes to the generation code since make will cache the bin and won't try to rebuild, I wonder why we went with a pre-built binary over `go run`


tools/api-generator/commands.go.tmpl
```diff
@@ -12,9 +13,10 @@
 // WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 // See the License for the specific language governing permissions and
 // limitations under the License.
-//
-// This code was autogenerated using `make gen-L1-commands`
+
+// Code generated using `make gen-api-commands`. DO NOT EDIT.
```
gssbzn, line 17: GH will consider any file with `// Code generated ...` as auto generated
Intellij will show a warning when opening files with `DO NOT EDIT.`

tools/api-generator/commands.go.tmpl
```diff
@@ -1,4 +1,5 @@
 {{- /* Template for generating L1 commands */ -}}
+{{- /*gotype: github.com/mongodb/mongodb-atlas-cli/atlascli/internal/api.GroupedAndSortedCommands */ -}}
```
gssbzn, line 2: intellij type hint

tools/api-generator/.snapshots/01-missing-tag.yaml.snapshot
gssbzn, line 1: this file has a year that auto-updates which means tests will start failing on January next year

jeroenvervaeke (approved):

tools/api-generator/.snapshots/01-missing-tag.yaml.snapshot
fmenezes: yeap, we will need to regenerate the snapshots once a year, I think that is ok, better the write `2024` like before

fmenezes (approved):

tools/api-generator/.snapshots/01-missing-tag.yaml.snapshot
gssbzn: copyright years don't need to be updated FYI, we also skip the whole copyright header on some autogenerated files like mocks and we tell the checks to ignore those folders, just things to consider if this becomes a problem

fmenezes: on old files is ok, but generating new files on old dates seems weird.

PR #3711, opened 2025-03-05
CLOUDP-303338: Extracts request body examples and parameter examples for auto-generated commands

cveticm:
## Proposed changes

Adds support to extract examples from the OpenAPI Specification for auto-generated commands. 
Examples extracted are:
* URL parameter examples
* Query parameter examples
* Request body examples

_Jira ticket:_ [CLOUDP-303338](https://jira.mongodb.org/browse/CLOUDP-303338)

## Checklist

- [ x ] I have signed the [MongoDB CLA](https://www.mongodb.com/legal/contributor-agreement)


jeroenvervaeke (changes requested):
The goal of this ticket was to output the examples to `tools/docs/examples.go`, not to `commands.go`, since we want to avoid having all the examples in the binary when we're not using them.

Let's sync offline tomorrow.

fmenezes (approved):

internal/api/api.go
```diff
@@ -115,3 +116,14 @@ type WatcherMatchProperties struct {
 	Path   string
 	Values []string
 }
+
+type Metadata struct {
```
fmenezes, line 120: I don't think this type belongs here, I think this will be used by `tools/docs`, so let's move it there, either inside the generated file or somewhere inside of the tool package

cveticm: Moved to new `tools/docs/metadatatypes` 👍

internal/api/api.go
```diff
@@ -59,6 +59,7 @@ type Parameter struct {
 	Description string
 	Required    bool
 	Type        ParameterType
+	Example     string
```
fmenezes, line 62: why do we need this field?

cveticm: This was an artifact from the older implementation. Thanks for catching, removed now!

fmenezes (approved):

PR #3728, opened 2025-03-11
CLOUDP-303342: Create examples for auto-generated projects CRUD operations

jeroenvervaeke:
## Proposed changes
- Added CRUD examples for projects
    - List projects
    - Create project
    - Get project
    - Update project
    - Delete project

For this I needed to make some changes:
- Replaced `"` with ` in metadata templates
- Added support for empty request bodies so we can add examples for `GET` and `DELETE` requests
- Filter out empty bodies when generating `commands.go`

_Jira ticket:_ CLOUDP-303342

### DIFF
--- a/tools/cmd/api-generator/overlays/examples_projects.yaml
+++ b/tools/cmd/api-generator/overlays/examples_projects.yaml
@@ -1,0 +1,48 @@
+overlay: 1.0.0
+info:
+  title: Configure groupId flag
+  version: 1.0.0
+actions:
+  - target: $.paths['/api/atlas/v2/groups'].post.requestBody.content['application/vnd.atlas.2023-01-01+json']
+    update:
+      examples:
+          create_project:
+              summary: Create a basic project with environment tag
+              value:
+                  name: "MongoTube"
+                  orgId: "67b715468c10250b968dcb84"
+                  tags:
+                    - key: "environment"
+                      value: "e2e"
+              description: Creates a new project named "MongoTube" with an environment tag set to "e2e"
+  - target: $.paths['/api/atlas/v2/groups/{groupId}'].get
+    update:
+      requestBody:
+        content:
+          application/vnd.atlas.2023-01-01+json:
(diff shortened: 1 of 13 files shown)


tools/cmd/api-generator/overlays/examples_projects.yaml
```diff
@@ -0,0 +1,58 @@
...
+          description: Creates a new project named "MongoTube" with an environment tag set to "e2e"
+  - target: $.paths['/api/atlas/v2/groups/{groupId}'].get
+    update:
+      requestBody:
+        content:
+          application/vnd.atlas.2023-01-01+json:
+            examples:
+              get_project:
+                summary: Get a project
+                value:
```
cveticm, line 36: Currently, the example extracting logic will dismiss an operation if `value` is empty. So the getProject example has not been generated into `metadata.go`.

This is a shortcoming of the extracting logic I overlooked since all operations that don't require request body are dismissed. I decided on this to keep metadata.go small since near to all operations include parameter examples. It might be worth undoing that logic.

```
func extractMetadata(operation *openapi3.Operation) (*metadatatypes.Metadata, error) {
	...
	requestBodyExamples, err := extractRequestBodyExamples(operation.RequestBody)
	if err != nil {
		return nil, err
	}
	if len(requestBodyExamples) == 0 { // <-- Remove this
		return nil, nil
	}
```

cveticm: Oh nevermind, totally blind, just missed getProjects in metadata.go

cveticm (approved):

tools/cmd/api-generator/overlays/examples_projects.yaml
jeroenvervaeke: There's an actual requestbody example, it just doesn't have a value.
If it were skipped, the generated `metadata.go` wouldn't be correct. (The examples are there)

PR #3729, opened 2025-03-11
CLOUDP-302671: Ignore preview versions

blva:
## Proposed changes

_Jira ticket:_ CLOUDP-302671

Closes #[issue number]

## Checklist

### DIFF
--- a/tools/cmd/api-generator/convert_commands.go
+++ b/tools/cmd/api-generator/convert_commands.go
@@ -429,8 +429,7 @@ func addContentTypeToVersion(versionedContentType string, versionsMap map[string
 		return fmt.Errorf("unsupported version %q error: %w", versionedContentType, err)
 	}
 
-	// Skip 'preview' versions
-	if strings.EqualFold(version, "preview") {
+	if shouldIgnoreVersion(version) {
 		return nil
 	}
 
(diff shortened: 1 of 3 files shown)


tools/cmd/api-generator/convert_commands.go
```diff
@@ -428,6 +429,11 @@ func addContentTypeToVersion(versionedContentType string, versionsMap map[string
 		return fmt.Errorf("unsupported version %q error: %w", versionedContentType, err)
 	}
 
+	// Skip 'preview' versions
+	if strings.EqualFold(version, "preview") {
+		return nil
+	}
+
```
cveticm, line 436: [suggestion] Not sure if preview versions will contain request body examples but, for completeness, would also be good to add similar for convert_metadata [here](https://github.com/mongodb/mongodb-atlas-cli/blob/master/tools/cmd/api-generator/convert_metadata.go#L96) so requestbody examples aren't extracted for preview versions.

```
if strings.EqualFold(version, "preview") {
		continue
	}
```

tools/cmd/api-generator/convert_commands.go
```diff
@@ -20,6 +20,7 @@ import (
 	"regexp"
```
cveticm, line 20: [question] Is this logic needed now that a new OAS is created which excludes preview versions and hidden endpoints? We could just ensure we are using this new OAS instead of the raw. 
This would also ensure that features/endpoints that only have "preview" versions wouldn't be exposed by auto-gen commands since, as the logic currently stands, a command will still be created for an endpoint if all its versions are preview.

jeroenvervaeke (approved):

tools/cmd/api-generator/convert_commands.go
blva: thanks! updated

tools/cmd/api-generator/convert_commands.go
blva: We are using the new OAS already, however, we don't filter out the public preview from that one, since it's public preview, hence, the only leftover piece is to ignore the 'preview' version via content type

cveticm: Aha, I misunderstood that this was for private preview. Thanks for clarifying!

PR #3735, opened 2025-03-13
CLOUDP-304437: Evaluate the use of pre-built bin for api-generator

jeroenvervaeke:
## Proposed changes
Got rid of rebuild `bin/api-generator`.

Did this by moving `internal/api/api.go` to `tools/shared/api/api.go`.
As a result, the CLI and api-generator tool can be built separately, meaning that a broken CLI build won't prevent us from running the api-generator and the other way around.

_Jira ticket:_ CLOUDP-304437


Makefile
```diff
@@ -111,8 +111,8 @@ bin/api-generator:
 .PHONY: gen-api-commands
 gen-api-commands: bin/api-generator ## Generate api commands
```
cveticm, line 112: The binary doesn't need to be generated anymore, correct? 

```suggestion
gen-api-commands: ## Generate api commands
```
This can also be removed
```
bin/api-generator:
	go build -o ./bin/api-generator ./tools/cmd/api-generator
```

jeroenvervaeke: Thanks, totally missed that!

cveticm (approved):
LGTM! Thanks for this improvement!

PR #3739, opened 2025-03-13
CLOUDP-306367: Add parameters to acceslist example

jeroenvervaeke:
## Proposed changes
Added parameters for project IP access list examples

_Jira ticket:_ CLOUDP-306367


cveticm (approved):
LGTM! Wondering if similar parameter example customisation will be done for the other example overlays?

PR #3763, opened 2025-03-24
task: update to jwt 5

gssbzn:
Unify versions of jwt, right now we include v4 and v5, one is local one a transitive, we can unify to the v5 and simplify


cveticm (approved):
LGTM

PR #3786, opened 2025-03-28
CLOUDP-309137: search e2e tests

andreaangiolillo:
This PR fixes the failing search e2e test.

Issue: the List command changed its output when there are no nodes from `404` to `{}`. I also discovered that there is no need to wait for the deletion of the search node if the next step is deleting the cluster.

### DIFF
--- a/test/e2e/search_nodes_test.go
+++ b/test/e2e/search_nodes_test.go
@@ -58,7 +58,7 @@ func TestSearchNodes(t *testing.T) {
 		respStr := string(resp)
 
 		require.NoError(t, err, respStr)
-		require.Equal(t, "{}", respStr)
+		require.Equal(t, "{}\n", respStr)
 	})
 
 	t.Run("Create search node", func(t *testing.T) {


test/e2e/search_nodes_test.go
```diff
@@ -57,8 +57,8 @@ func TestSearchNodes(t *testing.T) {
 		resp, err := cmd.CombinedOutput()
 		respStr := string(resp)
 
-		require.Error(t, err, respStr)
-		require.Contains(t, respStr, "ATLAS_SEARCH_DEPLOYMENT_DOES_NOT_EXIST", respStr)
+		require.NoError(t, err, respStr)
+		require.Empty(t, respStr)
```
andreaangiolillo, line 61: :hide_the_pain_andrea 😞 
```bash
./bin/atlas cluster search nodes ls -P dev1 --clusterName cluster-924-2d65cdd1524 -o json                                                                     
{}
```

gssbzn (approved):

PR #3799, opened 2025-04-01
CLOUDP-308177: [AtlasCLI] "atlas deployments connect" crashes when used with Atlas deployments in CREATING state

jeroenvervaeke:
## Proposed changes
Added additional null-checks and error handling to 'atlas deployments connect'

_Jira ticket:_ CLOUDP-308177

### DIFF
--- a/internal/cli/deployments/connect.go
+++ b/internal/cli/deployments/connect.go
@@ -41,6 +41,8 @@ var (
 	connectionStringTypeOptions              = []string{ConnectionStringTypeStandard, connectionStringTypePrivate}
 	errConnectionStringTypeNotImplemented    = errors.New("connection string type not implemented")
 	errNetworkPeeringConnectionNotConfigured = errors.New("network peering connection is not configured for this deployment")
+	errCouldNotConnectClusterNotIdle         = errors.New("could not connect: cluster is not in an idle state yet, try again in a few moments")
+	errCouldNotConnectNoConnectionString     = errors.New("could not connect: server did not return connectionstring")
 	promptConnectionStringType               = "What type of connection string type would you like to use?"
 )
 
@@ -287,6 +289,17 @@ func (opts *ConnectOpts) connectToAtlas() error {
 		return err
 	}
 
+	// Connectionstrings are empty when the server is not in IDLE
+	// r.ConnectionStrings.PrivateSrv == nil and r.ConnectionStrings.StandardSrv == nil
+	if r.StateName == nil || *r.StateName != "IDLE" {
+		return errCouldNotConnectClusterNotIdle
+	}
+
+	// This field is optional, if not set, throw an error
+	if r.ConnectionStrings == nil {
+		return errCouldNotConnectNoConnectionString
+	}
+
 	if opts.ConnectionStringType == connectionStringTypePrivate {
 		if r.ConnectionStrings.PrivateSrv == nil {
 			return errNetworkPeeringConnectionNotConfigured
@@ -294,5 +307,9 @@ func (opts *ConnectOpts) connectToAtlas() error {
 		return opts.connectToDeployment(*r.ConnectionStrings.PrivateSrv)
 	}
 
+	// Make sure the string pointer is not nil before dereferencing
+	if r.ConnectionStrings.StandardSrv == nil {
+		return errCouldNotConnectNoConnectionString
+	}
 	return opts.connectToDeployment(*r.ConnectionStrings.StandardSrv)
 }


internal/cli/deployments/connect.go
```diff
@@ -287,12 +289,27 @@ func (opts *ConnectOpts) connectToAtlas() error {
 		return err
 	}
 
+	// Connectionstrings are empty when the server is not in IDLE
+	// r.ConnectionStrings.PrivateSrv == nil and r.ConnectionStrings.StandardSrv == nil
+	if r.StateName == nil || *r.StateName != "IDLE" {
```
gssbzn, line 294:
```suggestion
	if r.GetStateName() != "IDLE" {
```
https://github.com/mongodb/atlas-sdk-go/blob/main/docs/doc_5_best_practices.md#using-getters-instead-of-direct-field-access

internal/cli/deployments/connect.go
```diff
@@ -287,12 +289,27 @@ func (opts *ConnectOpts) connectToAtlas() error {
...
+		return errCouldNotConnectClusterNotIdle
+	}
+
+	// This field is optional, if not set, throw an error
+	if r.ConnectionStrings == nil {
```
gssbzn, line 299: same as above

internal/cli/deployments/connect.go
```diff
@@ -41,6 +41,8 @@ var (
 	connectionStringTypeOptions              = []string{ConnectionStringTypeStandard, connectionStringTypePrivate}
 	errConnectionStringTypeNotImplemented    = errors.New("connection string type not implemented")
 	errNetworkPeeringConnectionNotConfigured = errors.New("network peering connection is not configured for this deployment")
+	errCouldNotConnectClusterNotIdle         = errors.New("could not connect: cluster is not in an idle state yet, try again in a few moments")
+	errCouldNotConnectNoConnectionString     = errors.New("could not connect: server did not return connectionstring")
```
gssbzn, line 45:
```suggestion
	errConnectionError         = errors.New("could not connect")
```

internal/cli/deployments/connect.go
```diff
@@ -287,12 +289,27 @@ func (opts *ConnectOpts) connectToAtlas() error {
 		return err
 	}
 
+	// Connectionstrings are empty when the server is not in IDLE
+	// r.ConnectionStrings.PrivateSrv == nil and r.ConnectionStrings.StandardSrv == nil
+	if r.StateName == nil || *r.StateName != "IDLE" {
+		return errCouldNotConnectClusterNotIdle
```
gssbzn, line 295:
```suggestion
		return fmt.Errorf("%w: cluster is not in an idle state yet, try again in a few moments", errConnectionError)
```

internal/cli/deployments/connect.go
```diff
@@ -287,12 +289,27 @@ func (opts *ConnectOpts) connectToAtlas() error {
...
+	}
+
+	// This field is optional, if not set, throw an error
+	if r.ConnectionStrings == nil {
+		return errCouldNotConnectNoConnectionString
```
gssbzn, line 300:
```suggestion
		return fmt.Errorf("%w: server did not return connectionstring", errConnectionError)
```

internal/cli/deployments/connect.go
jeroenvervaeke: Thanks, I wasn't aware of these getters and setters.

[q] If we recommend using these getters and setters. Why do we expose the underlying fields to the end user?

internal/cli/deployments/connect.go
jeroenvervaeke: I actually want to break the flow in case `ConnectionStrings == nil`. If I use the getter I'll get the default value, which is not what I want.

internal/cli/deployments/connect.go
jeroenvervaeke: updated

internal/cli/deployments/connect.go
jeroenvervaeke: updated

internal/cli/deployments/connect.go
gssbzn: because go, you can't serialize/deserialize private fields

internal/cli/deployments/connect.go
gssbzn: there's `GetConnectionStringsOK()` but in this case I don't think it makes it better

gssbzn (approved):
LGTM thanks for the changes

andreaangiolillo (approved):
LGTM

internal/cli/deployments/connect.go
fmenezes: but we can with `MarshalJSON` / `UnmarshalJSON`, I think we never paid attention to struct best practices in go `Get...` is considered against go's naming conventions https://go.dev/doc/effective_go#Getters

PR #3805, opened 2025-04-02
DEVPROD-16721: dual write release files to new bucket

drichmdb:
This commit adjusts our package_goreleaser task to dual write release
artifacts to the new cdn-origin-mongocli bucket. Eventually we will
remove writes to the old downloads bucket from this task.

### DIFF
--- a/build/ci/release.yml
+++ b/build/ci/release.yml
@@ -416,6 +416,26 @@ tasks:
           permissions: public-read
           content_type: ${content_type|application/x-gzip}
           display_name: downloads-center-
+      - command: s3.put
+        params:
+          role_arn: "arn:aws:iam::119629040606:role/s3-access.cdn-origin-atlascli"
+          local_files_include_filter:
+            - src/github.com/mongodb/mongodb-atlas-cli/dist/*.tar.gz
+            - src/github.com/mongodb/mongodb-atlas-cli/dist/*.zip
+            - src/github.com/mongodb/mongodb-atlas-cli/dist/*.deb
+            - src/github.com/mongodb/mongodb-atlas-cli/dist/*.rpm
+            - src/github.com/mongodb/mongodb-atlas-cli/dist/*.tgz
+            - src/github.com/mongodb/mongodb-atlas-cli/dist/*.json
+            - src/github.com/mongodb/mongodb-atlas-cli/dist/*.msi
+            - src/github.com/mongodb/mongodb-atlas-cli/dist/*.sig
+          remote_file: mongocli/
+          build_variants:
+            - release_mongocli_github
+            - release_atlascli_github
+          bucket: cdn-origin-mongocli
+          permissions: private
+          content_type: ${content_type|application/x-gzip}
+          display_name: downloads-center-new-
       - func: "send slack notification"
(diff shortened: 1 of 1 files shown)


build/ci/release.yml
```diff
@@ -416,6 +416,26 @@ tasks:
...
+            - src/github.com/mongodb/mongodb-atlas-cli/dist/*.rpm
+            - src/github.com/mongodb/mongodb-atlas-cli/dist/*.tgz
+            - src/github.com/mongodb/mongodb-atlas-cli/dist/*.json
+            - src/github.com/mongodb/mongodb-atlas-cli/dist/*.msi
+            - src/github.com/mongodb/mongodb-atlas-cli/dist/*.sig
+          remote_file: mongocli/
+          build_variants:
+            - release_mongocli_github
+            - release_atlascli_github
+          bucket: cdn-origin-mongocli
```
blva, line 435: will we have one for atlascli?

build/ci/release.yml
```diff
@@ -416,6 +416,26 @@ tasks:
...
+            - src/github.com/mongodb/mongodb-atlas-cli/dist/*.zip
+            - src/github.com/mongodb/mongodb-atlas-cli/dist/*.deb
+            - src/github.com/mongodb/mongodb-atlas-cli/dist/*.rpm
+            - src/github.com/mongodb/mongodb-atlas-cli/dist/*.tgz
+            - src/github.com/mongodb/mongodb-atlas-cli/dist/*.json
+            - src/github.com/mongodb/mongodb-atlas-cli/dist/*.msi
+            - src/github.com/mongodb/mongodb-atlas-cli/dist/*.sig
+          remote_file: mongocli/
+          build_variants:
+            - release_mongocli_github
```
blva, line 433: just double-checking if it's intentional to use mongocli, might be due to legacy folders

build/ci/release.yml
```diff
@@ -416,6 +416,26 @@ tasks:
           permissions: public-read
           content_type: ${content_type|application/x-gzip}
           display_name: downloads-center-
+      - command: s3.put
+        params:
+          role_arn: "arn:aws:iam::119629040606:role/s3-access.cdn-origin-atlascli"
```
blva, line 421: nit: let's make this an env var in our evergreen project since this is open source, probably valid for mongocli as well, I'm happy to set it up

build/ci/release.yml
drichmdb: Not as part of this migration -- we can do this in a follow-up where we go through and split the two projects' infrastructure up.

build/ci/release.yml
drichmdb: For the build variant? Honestly I'm not sure, but the other command was using it so I left it in!

build/ci/release.yml
drichmdb: We talked about this a bit offline, but I am happy to keep things as-is here or move it to an env var. Let me know what you decide you'd like to do and if we move to an env var I am happy to update the PR!

drichmdb:
I'm totally fine waiting for a real release to test if that's easier!

PR #3807, opened 2025-04-04
CLOUDP-306589: Extend `--debug` to include Docker logs on `atlas deployments setup --type local`

jeroenvervaeke:
## Proposed changes

`--debug` now includes docker logs when `atlas deployment setup` fails

_Jira ticket:_ CLOUDP-306589

### DIFF
--- a/internal/cli/deployments/setup.go
+++ b/internal/cli/deployments/setup.go
@@ -201,8 +201,11 @@ func (opts *SetupOpts) createLocalDeployment(ctx context.Context) error {
 
 func (opts *SetupOpts) configureContainer(ctx context.Context) error {
 	envVars := map[string]string{
-		"TOOL":            "ATLASCLI",
-		"RUNNER_LOG_FILE": "/dev/stdout",
+		"TOOL": "ATLASCLI",
+	}
+
+	if log.IsDebugLevel() {
+		envVars["RUNNER_LOG_FILE"] = "/dev/stdout"
 	}
 
 	if !config.TelemetryEnabled() {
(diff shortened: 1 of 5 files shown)


internal/cli/deployments/setup.go
```diff
@@ -201,7 +201,8 @@ func (opts *SetupOpts) createLocalDeployment(ctx context.Context) error {
 
 func (opts *SetupOpts) configureContainer(ctx context.Context) error {
 	envVars := map[string]string{
-		"TOOL": "ATLASCLI",
+		"TOOL":            "ATLASCLI",
+		"RUNNER_LOG_FILE": "/dev/stdout",
```
fmenezes, line 205: I think we should only do it if `--debug` was used, people expect this docker container to have mongod logs

internal/cli/deployments/setup.go
```diff
@@ -254,9 +255,22 @@ func (opts *SetupOpts) configureContainer(ctx context.Context) error {
 	}
 
 	// This can be a high number because the container will become unhealthy before the 10 minutes is reached
-	const healthyDeploymentTimeout = 3 * time.Minute
+	const healthyDeploymentTimeout = 10 * time.Minute
 
-	return opts.WaitForHealthyDeployment(ctx, healthyDeploymentTimeout)
+	err = opts.WaitForHealthyDeployment(ctx, healthyDeploymentTimeout)
+	if err != nil {
```
fmenezes, line 261: this actually works but initially I was expecting to tail logs 🤔

jeroenvervaeke: When something fails, don't you want to see the entire log instead of just the runner logs? Or am I misunderstanding your comment?

internal/cli/deployments/setup.go
jeroenvervaeke: Updated, I also only added the debug level check before getting the logs

fmenezes (approved):

PR #3836, opened 2025-04-22
CLOUDP-314780: [Improvement] Introduce logger interface

jeroenvervaeke:
## Proposed changes

- Introduced Logger interface
    - Main changes: `internal/log/log.go` 
- Removed `(int, error)` from logger functions
    - Updated all places where the code is called

_Jira ticket:_ CLOUDP-314780

### DIFF
--- a/internal/cli/api/api.go
+++ b/internal/cli/api/api.go
@@ -55,7 +55,7 @@ func Builder() *cobra.Command {
 			// err should always be nil, errors happening here should be covered by the generate api commands tool
 			// if err != nil there is a bug in the converter tool
 			if err != nil {
-				_, _ = log.Warningf("failed to add command for operationId: %s, err: %s", command.OperationID, err)
+				log.Warningf("failed to add command for operationId: %s, err: %s", command.OperationID, err)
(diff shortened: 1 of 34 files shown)


jeroenvervaeke:
> instead of reinventing the wheel can we use a standard logger like https://pkg.go.dev/log/slog#pkg-overview or https://github.com/uber-go/zap

Thanks for the suggestion! I appreciate you pointing out these standard logging libraries.

This PR is intentionally a small incremental improvement, focusing specifically on making code that uses our logger more testable by introducing an interface and removing tight coupling with IO.Writer return value. The scope is deliberately narrow to address this specific testing pain point.

Adopting a standard logger like slog or zap could be beneficial in the future, but it would require a larger refactoring effort, taking more time for implementation and testing than I currently have available.

jeroenvervaeke:
@gssbzn, I've created a followup ticket for your request [\[CLOUDP-314831\] \[CLI\] \[Improvement\] Move to slogger/zap](https://jira.mongodb.org/browse/CLOUDP-314831)

PR #3845, opened 2025-04-24
DEVPROD-16721: switch role used for writes to new bucket

drichmdb:
This commit adjusts the role we use to write to our new bucket. The previous role will only work when we separate out Atlas CLI from Mongo CLI's infrastructure. Because the two share a common bucket for now, I've swapped over to the appropriate role with access to that bucket.

### DIFF
--- a/build/ci/release.yml
+++ b/build/ci/release.yml
@@ -418,7 +418,7 @@ tasks:
           display_name: downloads-center-
       - command: s3.put
         params:
-          role_arn: "arn:aws:iam::119629040606:role/s3-access.cdn-origin-atlascli"
+          role_arn: "arn:aws:iam::119629040606:role/s3-access.cdn-origin-mongocli"
           local_files_include_filter:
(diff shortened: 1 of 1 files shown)


build/ci/release.yml
```diff
@@ -418,7 +418,7 @@ tasks:
           display_name: downloads-center-
       - command: s3.put
         params:
-          role_arn: "arn:aws:iam::119629040606:role/s3-access.cdn-origin-atlascli"
```
andreaangiolillo, line 421: We migrated mongocli to https://github.com/mongodb/mongodb-cli/tree/main

andreaangiolillo: Also all the post packaging tasks are failing after this PR was merged : https://spruce.mongodb.com/version/mongodb_atlas_cli_master_36f21cd8defe25ad9c89cf4d1d09f424e62cff55/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC

PR #3850, opened 2025-04-25
task: enable checks for unused interfaces

gssbzn:
## Proposed changes

Enable the checker for unused interfaces.

This aims to better follow bets practices where interfaces should be define don consumers, not producers.

This approach of interfaces on consumers was something I started a while ago but without a linter it was hard to enforce everywhere and it required a bit of redoing some old patters.

Now with this linter I have moved almost all interfaces to the consumer side, and also co-located the mocks to help with isolation

See the first commit for the pattern I'm following, the test of commits is applying the same pattern everywhere


internal/store/clusters.go
```diff
@@ -20,99 +20,18 @@ import (
 	atlas "go.mongodb.org/atlas/mongodbatlas"
 )
 
-//go:generate mockgen -destination=../mocks/mock_clusters.go -package=mocks github.com/mongodb/mongodb-atlas-cli/atlascli/internal/store ClusterLister,ClusterDescriber,ClusterCreator,ClusterDeleter,ClusterUpdater,AtlasClusterGetterUpdater,ClusterPauser,ClusterStarter,AtlasClusterQuickStarter,SampleDataAdder,SampleDataStatusDescriber,AtlasClusterConfigurationOptionsDescriber,AtlasSharedClusterDescriber,ClusterUpgrader,AtlasSharedClusterGetterUpgrader,AtlasClusterConfigurationOptionsUpdater,ClusterTester,ClusterDescriberStarter
+//go:generate mockgen -destination=../mocks/mock_clusters.go -package=mocks github.com/mongodb/mongodb-atlas-cli/atlascli/internal/store ClusterLister,ClusterDescriber
 
-type ClusterLister interface {
+type ClusterLister interface { //nolint:iface // right now requires some refactor to deployment commands
```
gssbzn, line 25: these are the only interfaces I could not move as right now they are inter-tangled in the deployments commands

andreaangiolillo (approved):
LGTM

PR #3859, opened 2025-04-29
task: delete unused store methods and mocks

gssbzn:
As part of #3850 I identified a couple of methods that are unused and we should delete


cveticm (approved):
LGTM

PR #3860, opened 2025-04-29
task: improve flag validation to use naive check for mutually exclusive

gssbzn:
## Proposed changes

Found a couple of places that can do with the cobra way to check for flag requirements


blva (approved):

erabil-mdb (commented):
LGTM!

PR #3903, opened 2025-05-22
CLOUDP-318638: Failed Test: mongodb-atlas-cli-master-github.com_mongodb_mongodb-atlas-cli_atlascli_test_e2e_backupflex.TestFlexBackup/atlas_backups_flex_e2e/e2e_atlas_backups

andreaangiolillo:
## Proposed changes
CLOUDP-318638

### DIFF
--- a/internal/cli/backup/restores/start.go
+++ b/internal/cli/backup/restores/start.go
@@ -174,7 +174,7 @@ func (opts *StartOpts) checkIsFlexCluster() error {
 		return nil
 	}
 
-	if !atlasv2.IsErrorCode(err, cannotUseFlexWithClusterApisErrorCode) {
+	if !atlasClustersPinned.IsErrorCode(err, cannotUseFlexWithClusterApisErrorCode) {
 		return err
 	}
 


jeroenvervaeke (approved):

internal/cli/backup/restores/start.go
```diff
@@ -174,7 +174,7 @@ func (opts *StartOpts) checkIsFlexCluster() error {
 		return nil
 	}
 
-	if !atlasv2.IsErrorCode(err, cannotUseFlexWithClusterApisErrorCode) {
+	if !atlasClustersPinned.IsErrorCode(err, cannotUseFlexWithClusterApisErrorCode) {
```
blva, line 177: [q] do we also need to check `	if !atlasv2.IsErrorCode(err, cannotUseFlexWithClusterApisErrorCode) {` in an && statement?

blva (approved):
🧹🧼

internal/cli/backup/restores/start.go
andreaangiolillo: no, the error is coming only from `atlasClustersPinned`. The issue was that the error was coming from SDK `atlasClustersPinned` but we were checking the error with `atlasv2` with is a different sdk version

PR #3932, opened 2025-06-03
fix snapshots for api-generation

jeroenvervaeke:
## Proposed changes

Made `specToCommands` function by receiving `now` as an argument.


andreaangiolillo (approved):
LGTM

jeroenvervaeke:
It's safe to ignore `Code Health / code-coverage`, it's failing because the unit tests on the previous commit were failing (which is what I fixed)

PR #3945, opened 2025-06-09
CLOUDP-322808: [AtlasCLI] Security violation - leaked secret

jeroenvervaeke:
## Proposed changes
Moved terraform password to Evergreen Secrets.

The password isn't used anywhere else in the codebase, we use certificates instead.

_Jira ticket:_ CLOUDP-322808


andreaangiolillo (approved):
LGTM

PR #3949, opened 2025-06-09
Cloudp 322615 2

blva:
## Proposed changes

_Jira ticket:_ CLOUDP-#

Closes #[issue number]

## Checklist

### DIFF
--- a/internal/cli/setup/cluster_config.go
+++ b/internal/cli/setup/cluster_config.go
@@ -1,0 +1,193 @@
+// Copyright 2025 MongoDB Inc
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package setup
+
+import (
+	"strings"
+
+	"github.com/mongodb/mongodb-atlas-cli/atlascli/internal/cli"
+	"github.com/mongodb/mongodb-atlas-cli/atlascli/internal/pointer"
+	atlasClustersPinned "go.mongodb.org/atlas-sdk/v20240530005/admin"
+	atlasv2 "go.mongodb.org/atlas-sdk/v20250312003/admin"
+	atlas "go.mongodb.org/atlas/mongodbatlas"
+)
+
+const (
+	defaultNodeCount = 3
+	defaultPriority  = 7
+)
+
+func defaultDiskSizeGB(provider, tier string) float64 {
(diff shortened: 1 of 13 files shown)


internal/cli/setup/cluster_config.go
```diff
@@ -0,0 +1,193 @@
...
+	defaultNodeCount = 3
+	defaultPriority  = 7
+)
+
+func defaultDiskSizeGB(provider, tier string) float64 {
+	return atlas.DefaultDiskSizeGB[strings.ToUpper(provider)][tier]
+}
+
+// newCluster creates a new cluster for the pinned API version.
+func (opts *Opts) newCluster() *atlasClustersPinned.AdvancedClusterDescription {
```
blva, line 37: this is from `cluster_setup.go`, I updated it a bit to simplify but added unit tests to make sure it's behaving the same

jvincent-mongodb (approved):

internal/cli/setup/cluster_config.go
```diff
@@ -0,0 +1,193 @@
...
+	if providerName == tenant {
+		regionConfig.BackingProviderName = &opts.Provider
+	} else {
+		regionConfig.ElectableSpecs.NodeCount = pointer.Get(defaultNodeCount)
+	}
+
+	return regionConfig
+}
+
+// getDiskSizeOverride returns the disk size override, which is only applicable for non-tenant clusters.
```
jeroenvervaeke: `DefaultDiskSizeGB` contains values for `TENANT`, so this comment is not 100% correct.

```go
var DefaultDiskSizeGB = map[string]map[string]float64{
	"TENANT": {
		"M2": 2,
		"M5": 5,
	},
	"AWS": {
		"M10":       10,
```

internal/cli/setup/cluster_config.go
```diff
@@ -0,0 +1,193 @@
...
+func (opts *Opts) getDiskSizeOverride() *float64 {
+	if opts.providerName() != tenant {
+		diskSizeGB := defaultDiskSizeGB(opts.providerName(), opts.Tier)
```
jeroenvervaeke: When you pass an incorrect combination to `defaultDiskSizeGB` the value returned is `0`, shouldn't we check that the value `!=0` before returning?

https://go.dev/play/p/mFYKv99R1VO

jeroenvervaeke (changes requested):

internal/cli/setup/cluster_config.go
blva: gotcha, it's because based on the previous implementation, it only sets the disk size override if the provider is not tenant based on `if opts.providerName() != tenant {` , let me see how I can rephrase it!

internal/cli/setup/cluster_config.go
blva: yes! good idea, and if it's zero we should also return nil, thank you 🙏

blva: updated and added unit tests, thanks!

internal/cli/setup/cluster_config.go
blva: updated the comment, let me know if it's better, if not we can remove it or you can make a suggestion! thanks

PR #3960, opened 2025-06-11
CLOUDP-300024: Bug: "plugin warning: could not load plugin"

jeroenvervaeke:
## Proposed changes
Updated list and uninstall command to support invalid plugins (duplicate manifest names/commands), so they can be uninstalled.

_Jira ticket:_ CLOUDP-300024

## Testing
Verified by:
1. Creating and installing a plugin with a duplicate command (I picked `api`)
2. Verifying that the error shows that the plugin is not loaded
3. Use `atlas plugin list` and verify that the plugin is visible
4. Uninstall the plugin using `atlas plugin uninstall`
5. Verify using `atlas plugin list` and the directory structure that the plugin is uninstalled

### DIFF
--- a/internal/cli/plugin/first_class.go
+++ b/internal/cli/plugin/first_class.go
@@ -90,8 +90,12 @@ func (fcp *FirstClassPlugin) isAlreadyInstalled(plugins *plugin.ValidatedPlugins
 	return false
 }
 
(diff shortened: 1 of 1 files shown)


andreaangiolillo (approved):
LGTM

andreaangiolillo (approved):
LGTM

PR #3964, opened 2025-06-13
CLOUDP-317733: Code coverage report fails for forks

jeroenvervaeke:
## Proposed changes
Split the code coverage and commenting workflow into 2 workflows:
1. The original workflow still runs code coverage and compares the code coverage like before, with a read-only token
    - Instead of commenting we save the comment as an artifact 
2. `code-heatlth-comment-coverage.yml` runs on the base of the branch (for security) with a write token and comments.

_Jira ticket:_ CLOUDP-317733


.github/workflows/code-health-comment-coverage.yml
```diff
@@ -1,20 +1,50 @@
 ---
 name: Code Health - Code coverage comment
 on:
-  workflow_run:
-    workflows: ["Code Health"]
-    types:
-      - completed
+  pull_request_target:
```
fmenezes: [q] do you need both `pull_request_target` and `pull_request` ?

fmenezes: I tihnk `pull_request_target` only triggers on forks not normal PRs https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#pull_request_target

jeroenvervaeke: No, but it limits the permissions:

> For workflows that are triggered by the pull_request_target event, the GITHUB_TOKEN is granted read/write repository permission unless the permissions key is specified and the workflow can access secre

fmenezes (dismissed):

PR #3968, opened 2025-06-16
updating copybara config

jwilliams-mongo:
## Proposed changes

we're consolidating our documentation repositories. Tomorrow we will move the source for the Atlas CLI documentation into the core mongodb docs repository. This updates the copybara config that creates a pull request to copy the generated docs files to the Atlas CLI docs.

We're moving the Atlas docs later so that section of the config will not change yet.


andreaangiolillo (approved):
LGTM

PR #3989, opened 2025-06-19
CLOUDP-326299: List independentShardScaling flag should list only ISS…

blva:
… clusters

## Proposed changes

_Jira ticket:_ CLOUDP-#

Closes #[issue number]

## Checklist

### DIFF
--- a/internal/cli/clusters/list.go
+++ b/internal/cli/clusters/list.go
@@ -36,6 +36,7 @@ type ClusterLister interface {
 	ProjectClusters(string, *store.ListOptions) (*atlasClustersPinned.PaginatedAdvancedClusterDescription, error)
 	ListFlexClusters(*atlasv2.ListFlexClustersApiParams) (*atlasv2.PaginatedFlexClusters20241113, error)
 	LatestProjectClusters(string, *store.ListOptions) (*atlasv2.PaginatedClusterDescription20240805, error)
+	GetClusterAutoScalingConfig(string, string) (*atlasv2.ClusterDescriptionAutoScalingModeConfiguration, error)
 }
 
 type ListOpts struct {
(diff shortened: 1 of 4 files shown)


internal/cli/clusters/list.go
```diff
@@ -85,6 +90,22 @@ func (opts *ListOpts) RunDedicatedCluster() error {
 	return opts.Print(r)
 }
 
+func (opts *ListOpts) filterClustersByAutoScalingMode(clusters *atlasv2.PaginatedClusterDescription20240805) (*atlasv2.PaginatedClusterDescription20240805, error) {
```
cveticm, line 93: Question: Does LatestProjectClusters lists clusters of all autoScaling modes? If so, can the old cluster list method be replaced by this with filtering for clusterWide clusters? 

Also, is there any worry that getting a cluster in clusterWideScaling mode might trigger the cluster to transition to ISS? I recall having to be cautious about this when I touched ISS logic in TF but don't remember if this extending to get methods.

blva: > Question: Does LatestProjectClusters lists clusters of all autoScaling modes? If so, can the old cluster list method be replaced by this with filtering for clusterWide clusters?

Yes. It does, but we cannot replace it completely since it has a different json output. To avoid breaking changes, we can only use the latest listing when a customer explicitly uses --autoScalingMode independentShardScaling

Question: Does LatestProjectClusters lists clusters of all autoScaling modes? If so, can the old cluster list method be replaced by this with filtering for clusterWide clusters?

> Also, is there any worry that getting a cluster in clusterWideScaling mode might trigger the cluster to transition to ISS? I recall having to be cautious about this when I touched ISS logic in TF but don't remember if this extending to get methods.
Yes, but this would only happen for UPDATE commands, if the customer calls update with the independentShardScaling, their cluster will have it enabled, but that is expected.

cveticm: Awesome, thanks for the explanation!

cveticm (approved):
LGTM!

PR #4034, opened 2025-07-10
CLOUDP-329791: Refactor auth check

cveticm:
## Proposed changes

This change migrates the auth check logic to run during transport instead of in the prerun of the root command.

### Main Changes
**1. Remove validation from `atlas` prerun**
Now the default transport client has a wrapper which calls the default RoundTrip and, on response status 401: Unauthorized, the following error is bubbled up:
```
Error: Get "https://cloud.mongodb.com/api/atlas/v2/groups?includeCount=true&itemsPerPage=100&pageNum=1": this action requires authentication

```

The same behaviour is preserved as, if neither API Keys or Access Token are present, the default transport will be used and, if used to call a cmd that requires auth, it will error with similar messaging.

**Note:**
The unauthorized error now takes a slightly different form, with the response wrapped in the error. (E.g. `Get "https://cloud.mongodb.com/api/atlas/v2/groups?includeCount=true&itemsPerPage=100&pageNum=1":`) To my understanding, this is the standard behaviour of Go http clients & cobra. Removing this would require additional code to catch and format the error at the command level. As I believe this is not a breaking change, I have left this addition to the error messaging.

**2. Remove token refresh from `atlas` prerun**
The access token Round Tripper already verifies if the token is expired and refreshes if it is not.

_Jira ticket:_ [CLOUDP-329791](https://jira.mongodb.org/browse/CLOUDP-329791)

## Checklist

### DIFF
--- a/internal/cli/root/builder.go
+++ b/internal/cli/root/builder.go
@@ -67,13 +67,11 @@ import (
 	"github.com/mongodb/mongodb-atlas-cli/atlascli/internal/homebrew"
 	"github.com/mongodb/mongodb-atlas-cli/atlascli/internal/latestrelease"
 	"github.com/mongodb/mongodb-atlas-cli/atlascli/internal/log"
-	"github.com/mongodb/mongodb-atlas-cli/atlascli/internal/plugin"
 	"github.com/mongodb/mongodb-atlas-cli/atlascli/internal/prerun"
 	"github.com/mongodb/mongodb-atlas-cli/atlascli/internal/sighandle"
 	"github.com/mongodb/mongodb-atlas-cli/atlascli/internal/telemetry"
 	"github.com/mongodb/mongodb-atlas-cli/atlascli/internal/terminal"
 	"github.com/mongodb/mongodb-atlas-cli/atlascli/internal/usage"
-	"github.com/mongodb/mongodb-atlas-cli/atlascli/internal/validate"
 	"github.com/mongodb/mongodb-atlas-cli/atlascli/internal/version"
 	"github.com/spf13/afero"
 	"github.com/spf13/cobra"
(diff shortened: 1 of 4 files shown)


internal/transport/transport.go
```diff
@@ -110,3 +112,24 @@ func (tr *tokenTransport) RoundTrip(req *http.Request) (*http.Response, error) {
...
+func (tr *AuthRequiredRoundTripper) RoundTrip(req *http.Request) (*http.Response, error) {
+	resp, err := tr.Base.RoundTrip(req)
+	if resp != nil && resp.StatusCode == http.StatusUnauthorized {
+		return nil, fmt.Errorf(
+			`%w
+
+To log in using your Atlas username and password, run: atlas auth login
+To set credentials using API keys, run: atlas config init`,
+			validate.ErrMissingCredentials,
+		)
```
cveticm, line 129: Example in practice:
```
Error: Get "https://cloud.mongodb.com/api/atlas/v2/groups?includeCount=true&itemsPerPage=100&pageNum=1": this action requires authentication

```
Note the addition of the response error `Get "https://cloud.mongodb.com/api/atlas/v2/groups?includeCount=true&itemsPerPage=100&pageNum=1":`

Please flag if you ahve any concerns with this addition to the error messaging.

jeroenvervaeke (approved):

internal/cli/root/builder.go
```diff
@@ -289,43 +253,6 @@ func shouldSetService(cmd *cobra.Command) bool {
...
-	}
-	for p, r := range customRequirements {
-		if strings.HasPrefix(cmd.CommandPath(), p) {
-			return r
-		}
-	}
-
-	if plugin.IsPluginCmd(cmd) || pluginCmd.IsFirstClassPluginCmd(cmd) {
-		return OptionalAuth
-	}
```
blva, line 324: [qq] how will this work for plugins? now that we moved it under transport, maybe it won't show up for K8s anymore - but since it's optional auth it's probably all right

cveticm: That's a good point, the K8 plugin uses its own transport code which means it does not have an auth check. 
I think the best solution is to make the fix on the atlas k8s plugin code. I will open a ticket to do this.

Interestingly, the plugins are marked as OptionalAuth and no other action is taken with them. ie no auth check is done for plugins! And the current behaviour of k8s plugin is to fail with the raw 401:Unauth message.

PR #4038, opened 2025-07-14
CLOUDP-330675: Reworks `atlas auth login` flow

cveticm:
## Proposed changes

1. Adds prompt flow to ask user which form of authentication they would like to log in with.
```
atlas auth login
? Select authentication type:  [Use arrows to move, type to filter]
> UserAccount - (best for getting started)
  APIKeys - (for existing automations)
```
APIKeys logic is taken directly from `atlas config init`. This duplicated code will be removed as a part of [CLOUDP-329797](https://jira.mongodb.org/browse/CLOUDP-329797)

2. Adds hidden flag `--force`. When used, browser flow (User Account) is taken and user will not be prompted on authentication method.

_Jira ticket:_ [CLOUDP-330675](https://jira.mongodb.org/browse/CLOUDP-330675)

## Checklist


jeroenvervaeke (approved):

internal/cli/auth/login.go
```diff
@@ -316,8 +421,10 @@ func LoginBuilder() *cobra.Command {
 	}
 
 	cmd.Flags().BoolVar(&opts.IsGov, "gov", false, "Log in to Atlas for Government.")
-	cmd.Flags().BoolVar(&opts.NoBrowser, "noBrowser", false, "Don't try to open a browser session.")
+	cmd.Flags().BoolVar(&opts.NoBrowser, "noBrowser", false, "Don't try to open a browser session to authenticate your User Account.")
```
jwilliams-mongo: if this is true, will the CLI use only config files / env variables for auth? Or perhaps a session that is already authenticated? Wondering if we need to be clear about that behavior here.

I think we can remove the word "try" as it isn't clear to me if this is true if it just doesn't open a browser window or if it opens it if specific conditions are met or not met.
```suggestion
	cmd.Flags().BoolVar(&opts.NoBrowser, "noBrowser", false, "Don't open a browser session to authenticate your User Account.")
```

cveticm: When `--noBrowser=false`, a browser session is opened for users to authenticate through.

When it is `true`, the behaviour is that a browser session will not be opened automatically and users will have to access the link manually.

Would this make that behaviour more clear?
```suggestion
	cmd.Flags().BoolVar(&opts.NoBrowser, "noBrowser", false, "Don't automatically open a browser session.")
```

internal/cli/auth/login.go
```diff
@@ -238,6 +344,9 @@ func (opts *LoginOpts) oauthFlow(ctx context.Context) error {
 
 		opts.printAuthInstructions(code)
 		if !askedToOpenBrowser {
+			fmt.Printf("\nBrowser will be automatically opened in %d seconds...", browserWait)
```
fmenezes: make sure to run by Jonny, he designed these changes

cveticm: Noted, will check

cveticm: modified this logic to more closely follow what was outlined in Scope (prompt user to press a key to open browser) and checked w Jonny, he gives his thumbs up on new implementation

internal/cli/auth/login.go
```diff
@@ -49,20 +51,119 @@ type LoginConfig interface {
 	ProjectID() string
 }
 
+type TrackAsker interface {
```
fmenezes, line 54: while I think this is fine, why have you added this? don't we just disable telemetry on our tests?

cveticm: I tried and was unable to get this to work for testing runApiKeysLogin.
I believe the limitation is that, even if telemetry is disabled, the prompts are still asked and wait for input, resulting in the test failing with err EOF. 

This structure also allows us to mock prompt responses (as seen for Test_shouldRetryAuthenticate [here](https://github.com/mongodb/mongodb-atlas-cli/pull/4038/files#diff-03fa31381919ba1ca6511afca2b2ff17292ba7266c0f5adc6491e96c75ec38acR218-R225)) which would allow us to cover more specific cases in future if needed.

fmenezes (approved):

PR #4043, opened 2025-07-16
CLOUDP-330561: Moves unauth error check to common errors

cveticm:
#### Issues
Initial work to move unauth check from builder.go to transport.go (#4034) neglected digest and access token transport clients. This meant that if either of these transport methods received a 401 response, the error would not be wrapped with any additional messaging. For example, when a user's access token expires.

#### Solution
To fix this, I have moved the unauth check to atlas.go.
The flow now is:
* a command finishes executing
* if err
  * check if 401 
    * if 401, print custom error message to stdErr

Note that the raw transport error message will be printed to terminal before the command is fully complete. In practice, on terminal this looks much the same as the previous implementation:
```
Error: https://cloud-dev.mongodb.com/api/atlas/v2/groups/670cd17af33cea212ea1ed61/clusters GET: HTTP 401 Unauthorized (Error code: "") Detail: You are not authorized for this resource. Reason: Unauthorized. Params: []
this action requires authentication

```

#### Testing
Unit tests have been added.

As oauth authentication (Access Tokens) requires a browser-based flow, e2e testing could not be created to verify that access tokens are refreshed on expiry.
Manual testing was done instead. Access and Refresh Tokens are successfully refreshed when they expire. 
**Future work:** E2E testing will be created once Service Accounts are implemented ([CLOUDP-329808](https://jira.mongodb.org/browse/CLOUDP-329808)). An e2e test will be created to verify tokens are refreshed on expiry.

_Jira ticket:_ [CLOUDP-330561](https://jira.mongodb.org/browse/CLOUDP-330561)

## Checklist

### DIFF
--- a/cmd/atlas/atlas.go
+++ b/cmd/atlas/atlas.go
@@ -37,10 +37,8 @@ func execute(rootCmd *cobra.Command) {
 
 To learn more, see our documentation: https://www.mongodb.com/docs/atlas/cli/stable/connect-atlas-cli/`
 	if cmd, err := rootCmd.ExecuteContextC(ctx); err != nil {
-		errMsg := commonerrors.GetHumanFriendlyErrorMessage(err)
-		if errMsg != nil {
-			fmt.Fprintln(os.Stderr, errMsg)
-		}
+		err := commonerrors.Check(err)
+		rootCmd.PrintErrln(rootCmd.ErrPrefix(), err)
 		if !telemetry.StartedTrackingCommand() {
 			telemetry.StartTrackingCommand(cmd, os.Args[1:])
 		}
--- a/internal/cli/commonerrors/errors.go
+++ b/internal/cli/commonerrors/errors.go
@@ -31,63 +31,93 @@ var (
 	
 To log in using your Atlas username and password, run: atlas auth login
 To set credentials using API keys, run: atlas config init`)
+	ErrInvalidRefreshToken = errors.New(`session expired
+
+Please note that your session expires periodically. 
+If you use Atlas CLI for automation, see https://www.mongodb.com/docs/atlas/cli/stable/atlas-cli-automate/ for best practices.
+To login, run: atlas auth login`)
 )
 
 const (
-	asymmetricShardUnsupportedErrorCode = "ASYMMETRIC_SHARD_UNSUPPORTED"
+	unknownErrorCode                        = "UNKNOWN_ERROR"
+	asymmetricShardUnsupportedErrorCode     = "ASYMMETRIC_SHARD_UNSUPPORTED"
+	tenantClusterUpdateUnsupportedErrorCode = "TENANT_CLUSTER_UPDATE_UNSUPPORTED"
+	globalUserOutsideSubnetErrorCode        = "GLOBAL_USER_OUTSIDE_SUBNET"
+	unauthorizedErrorCode                   = "UNAUTHORIZED"
+	invalidRefreshTokenErrorCode            = "INVALID_REFRESH_TOKEN"
 )
 
+// Check checks the error and returns a more user-friendly error message if applicable.
 func Check(err error) error {
 	if err == nil {
 		return nil
 	}
 
-	apiError, ok := atlasv2.AsError(err)
-	if ok {
-		switch apiError.GetErrorCode() {
-		case "TENANT_CLUSTER_UPDATE_UNSUPPORTED":
-			return errClusterUnsupported
-		case "GLOBAL_USER_OUTSIDE_SUBNET":
-			return errOutsideVPN
-		case asymmetricShardUnsupportedErrorCode:
-			return errAsymmetricShardUnsupported
-		}
+	apiErrorCode := getErrorCode(err)
+
+	switch apiErrorCode {
+	case unauthorizedErrorCode:
+		return ErrUnauthorized
+	case invalidRefreshTokenErrorCode:
+		return ErrInvalidRefreshToken
+	case tenantClusterUpdateUnsupportedErrorCode:
+		return errClusterUnsupported
+	case globalUserOutsideSubnetErrorCode:
+		return errOutsideVPN
+	case asymmetricShardUnsupportedErrorCode:
+		return errAsymmetricShardUnsupported
 	}
+
+	apiError := getError(err) // some `Unauthorized` errors do not have an error code, so we check the HTTP status code
+
+	if apiError == http.StatusUnauthorized {
+		return ErrUnauthorized
+	}
+
 	return err
 }
 
-// GetHumanFriendErrorMessage returns a human-friendly error message if error is status code 401.
-func GetHumanFriendlyErrorMessage(err error) error {
+// getErrorCode extracts the error code from the error if it is an Atlas error.
+// This function checks for v2 SDK, the pinned clusters SDK and the old SDK errors.
+// If the error is not any of these Atlas errors, it returns "UNKNOWN_ERROR".
+func getErrorCode(err error) string {
 	if err == nil {
-		return nil
+		return unknownErrorCode
 	}
 
-	statusCode := GetErrorStatusCode(err)
-	if statusCode == http.StatusUnauthorized {
-		return ErrUnauthorized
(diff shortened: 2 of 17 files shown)


internal/transport/transport.go
```diff
@@ -112,24 +110,3 @@ func (tr *tokenTransport) RoundTrip(req *http.Request) (*http.Response, error) {
 
 	return tr.base.RoundTrip(req)
 }
-
-type AuthRequiredRoundTripper struct {
-	Base http.RoundTripper
-}
-
```
cveticm, line 119: transport wrapper no longer needed since its primary purpose, to check for unauth, is now done after the command executes

internal/cli/commonerrors/errors.go
```diff
@@ -49,6 +54,13 @@ func Check(err error) error {
 	return err
 }
 
+func CheckHTTPErrors(err error) error {
+	if strings.Contains(err.Error(), "401") && strings.Contains(err.Error(), "Unauthorized") {
```
jeroenvervaeke: I don't think this is the proper way to check the error code coming from the API.

I think you'll have to use:
```go
admin.AsError(err)
```

I was first going to recommend using the errorcode: `UNAUTHORIZED`, but it looks like we have so many [variations](https://github.com/10gen/mms/blob/7d8d9184723e598a46d8d930f65d41fe993f8683/server/src/main/com/xgen/svc/mms/api/res/common/ApiErrorCode.java#L2097).

It's better to check the HTTP status code instead (through the error).

Unfortunately I don't fully remember how to get the status code from the SDK, let me know if you can't find it and want to look into it together

cmd/atlas/atlas.go
```diff
@@ -36,6 +37,10 @@ func execute(rootCmd *cobra.Command) {
 
 To learn more, see our documentation: https://www.mongodb.com/docs/atlas/cli/stable/connect-atlas-cli/`
 	if cmd, err := rootCmd.ExecuteContextC(ctx); err != nil {
+		errMsg := commonerrors.CheckHTTPErrors(err)
+		if errMsg != nil {
+			fmt.Println(errMsg)
+		}
```
andreaangiolillo: shouldn't we print this to stderr?

internal/cli/commonerrors/errors.go
andreaangiolillo: > Unfortunately I don't fully remember how to get the status code from the SDK, let me know if you can't find it and want to look into it together

`AsError` returns the `ApiError` struct which contains the error code as param
https://github.com/mongodb/atlas-sdk-go/blob/main/admin/model_api_error.go#L6

cveticm: Yes i tried this initially, but these unauth errors aren't GenericOpenAPIError types so admin.AsError(err) returns !ok.

fmenezes: I think we need a custom error, perhaps should we introduce error handling on transport and create an ErrUnauthorized? at least a generic http error that has code as a property

cmd/atlas/atlas.go
fmenezes: I think this prints it twice because cobra prints it

fmenezes: If we are moving the error check here we should remove each individual usage of it and move it here

internal/cli/commonerrors/errors.go
cveticm: Introduced custom error on transport which stores response details on response status code being an error. Let me know if this is a better approach

internal/validate/validate_test.go
```diff
@@ -169,7 +169,7 @@ func TestCredentials(t *testing.T) {
 	})
 }
 
-func TestNoAPIKeyss(t *testing.T) {
+func TestNoAPIKeys(t *testing.T) {
```
cveticm, line 172: Drive-by fix

cmd/atlas/atlas.go
```diff
@@ -36,6 +37,10 @@ func execute(rootCmd *cobra.Command) {
 
 To learn more, see our documentation: https://www.mongodb.com/docs/atlas/cli/stable/connect-atlas-cli/`
 	if cmd, err := rootCmd.ExecuteContextC(ctx); err != nil {
+		errMsg := commonerrors.GetHumanFriendlyErrorMessage(err)
+		if errMsg != nil {
+			fmt.Fprintln(os.Stderr, errMsg)
```
fmenezes: we did not have this print before, how was cobra printing it?

fmenezes: I think errors might be printing twice, there is a SilenceErrors property in cobra, double check if you need it

cveticm: cobra prints to stderr but only prints response error. errMsg only prints custom error (ie. "this action requires authentication ...").

cveticm: Since errMsg has different content than err printed by cobra, we don't get duplication.

fmenezes: running this code this is what I get after my browser login is expired
```
Error: Get "https://cloud-dev.mongodb.com/api/atlas/v2/groups/686f96db2f08ee49a6da7ca8/clusters?includeCount=true&includeDeletedWithRetainedBackups=false&itemsPerPage=100&pageNum=1": POST https://cloud-dev.mongodb.com/api/private/unauth/account/device/token: 400 (request "INVALID_REFRESH_TOKEN") The refresh token is invalid or expired.
exit status 1
```

on master I get
```
Error: session expired

Please note that your session expires periodically. 
If you use Atlas CLI for automation, see https://www.mongodb.com/docs/atlas/cli/stable/atlas-cli-automate/ for best practices.
To login, run: atlas auth login
exit status 1
```

I prefer master's experience, can we keep it?

tested by running `atlas clusters list`

fmenezes (changes requested):
check my comment on the thread unfortunately the UX here needs some more work, good news is I've managed the find the way I think we should move forward, take a look at my comments and let me know.

cmd/atlas/atlas.go
fmenezes: I know why that is, the error of not being able to refresh token is a 400 not a 401. I think this is a sort of chicken and egg problem.

I've managed to go back a to a good UX by:

- updated `internal/cli/commonerrors/errors.go`
```diff
func Check(err error) error {
	if err == nil {
		return nil
	}
+
+	apiErrorCode := "UNKNOWN"
+
+	var sdkError *atlas.ErrorResponse
+	if ok := errors.As(err, &sdkError); ok {
+		apiErrorCode = sdkError.ErrorCode
+	}

	apiError, ok := atlasv2.AsError(err)
	if ok {
-		switch apiError.GetErrorCode() {
+		apiErrorCode = apiError.GetErrorCode()
+	}
+
+	apiPinnedError, ok := atlasClustersPinned.AsError(err)
+	if ok {
+		apiErrorCode = apiPinnedError.GetErrorCode()
+	}
+
+	switch apiErrorCode {
	case "TENANT_CLUSTER_UPDATE_UNSUPPORTED":
		return errClusterUnsupported
	case "GLOBAL_USER_OUTSIDE_SUBNET":
		return errOutsideVPN
	case asymmetricShardUnsupportedErrorCode:
		return errAsymmetricShardUnsupported
+	case "INVALID_REFRESH_TOKEN":
+		return ErrUnauthorized
-		}
	}
+
	return err
}
```
- updated `cmd/atlas/atlas.go`
```diff
	if cmd, err := rootCmd.ExecuteContextC(ctx); err != nil {
-		errMsg := commonerrors.GetHumanFriendlyErrorMessage(err)
+		errMsg := commonerrors.Check(err)
		if errMsg != nil {
```
- updated `internal/cli/root/builder.go`
```diff
		Example: `  # Display the help menu for the config command:
  atlas config --help
`,
		SilenceUsage:  true,
+		SilenceErrors: true,
		Annotations: map[string]string{
			"toc": "true",
		},

```

cveticm: Oh I see! I agree, keeping UX from master is better. Thanks for looking into this and giving the suggestion

fmenezes (approved):
Thanks for the changes 🎉

PR #4044, opened 2025-07-16
CLOUDP-329797: Make `atlas config init` alias of `atlas auth login`

cveticm:
## Proposed changes

As these commands are under different parent commands, this is achieved by calling the Login Builder in the Init Builder and setting different Use and Example values.

This results in `atlas config init` running exactly the same as `atlas auth login` while still referring to the command correctly as `atlas config init` in help text.

_Jira ticket:_ [CLOUDP-329797](https://jira.mongodb.org/browse/CLOUDP-329797)

## Checklist

- [x] I have signed the [MongoDB CLA](https://www.mongodb.com/legal/contributor-agreement)
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] I have added any necessary documentation in document requirements section listed in [CONTRIBUTING.md](https://github.com/mongodb/mongodb-atlas-cli/blob/master/CONTRIBUTING.md) (if appropriate)
- [ ] I have addressed the @mongodb/docs-cloud-team comments (if appropriate)
- [ ] I have updated [test/README.md](https://github.com/mongodb/mongodb-atlas-cli/blob/master/test/README.md) (if an e2e test has been added)
- [x] I have run `make fmt` and formatted my code

PR #4050, opened 2025-07-23
CLOUDP-332913: [AtlasCLI] Decouple profile from viper

jeroenvervaeke:
## Proposed changes

- Decoupled profile from viper by introducing `config.Store` (originally called `ConfigStore`, but the linter didn't like it)
    - All viper logic now lives in `ViperConfigStore`, I tried to not touch the logic, mainly copy paste where possible
- Introduced `WithProfile`/`ProfileFromContext` and attached the profile to the context, so in commands you can get the `Profile` from the `context` instead of using static variables.
- Split `profile.go` into `profile.go`/`identifiers.go` (code was unrelated to profile)
- I had to disable tests in `internal/validate/validate_test.go` because they directly invoke viper 👎 , will fix them in a followup PR

_Jira ticket:_ CLOUDP-332913


cmd/atlas/atlas.go
```diff
@@ -48,12 +49,17 @@ To learn more, see our documentation: https://www.mongodb.com/docs/atlas/cli/sta
...
 
 // loadConfig reads in config file and ENV variables if set.
-func loadConfig() error {
-	if err := config.LoadAtlasCLIConfig(); err != nil {
-		return fmt.Errorf("error loading config: %w. Please run `atlas config init` to reconfigure your profile", err)
+func loadConfig() (*config.Profile, error) {
+	configStore, initErr := config.LoadViperStore(afero.NewOsFs())
+
+	if initErr != nil {
+		return nil, fmt.Errorf("error loading config: %w. Please run `atlas config init` to reconfigure your profile", initErr)
```
cveticm: With #4044, 'atlas config init' is deprecated. we should avoid pointing users to it
```suggestion
		return nil, fmt.Errorf("error loading config: %w. Please run `atlas auth login` to reconfigure your profile", initErr)
```

internal/validate/validate_test.go
```diff
@@ -141,6 +141,7 @@ func TestObjectID(t *testing.T) {
 	}
 }
 
+/*
```
fmenezes: instead of commenting the test you can add `t.Skip("Will reenable on ticket CLOUDP-XXX")`

internal/validate/validate_test.go
```diff
@@ -168,23 +169,26 @@ func TestCredentials(t *testing.T) {
...
-		// this function depends on the global config (globals are bad I know)
-		// the easiest way we have to test it is via ENV vars
-		viper.AutomaticEnv()
-		t.Setenv("PUBLIC_API_KEY", "test")
-		t.Setenv("PRIVATE_API_KEY", "test")
-		if err := NoAPIKeys(); err == nil {
-			t.Fatalf("NoAPIKeys() expected error\n")
-		}
-	})
+	/*
```
fmenezes: same as above `t.Skip`

cmd/atlas/atlas.go
jeroenvervaeke: Sure, I didn't write this. I moved the line from somewhere else :)

internal/config/viper_store.go
```diff
@@ -0,0 +1,221 @@
...
+	"github.com/spf13/viper"
+)
+
+type ViperConfigStore struct {
+	viper     viper.Viper
+	configDir string
+	fs        afero.Fs
+}
+
+// / ViperConfigStore specific methods
```
andreaangiolillo: [nit] feel free to address in a follow up
```suggestion
// ViperConfigStore specific methods
```

internal/config/viper_store.go
```diff
@@ -0,0 +1,221 @@
...
+func LoadViperStore(fs afero.Fs) (*ViperConfigStore, error) {
```
andreaangiolillo: (maybe I am missing something) I would have expected `viperStore` to implement the `store` interface, with `config` defining a `LoadStore` function that returns a generic `Store` interface. I’m wondering if you considered this design approach. The pros of this approach is that we will be able to easily remove viper with something else if needed int he future.

internal/config/viper_store.go
```diff
@@ -0,0 +1,221 @@
...
+
+func ViperConfigStoreFilename(configDir string) string {
+	return filepath.Join(configDir, "config.toml")
+}
+
+func (s *ViperConfigStore) Filename() string {
+	return ViperConfigStoreFilename(s.configDir)
+}
+
+/// ConfigStore implementation
```
andreaangiolillo: is this needed?

internal/config/viper_store.go
andreaangiolillo: Never mind my previous comment, I downloaded your PR and verified that your code is indeed doing what I was suggesting above. Great! 💯

andreaangiolillo: [nit]
```suggestion
func NewViperStore(fs afero.Fs) (*ViperConfigStore, error) {
```

internal/config/viper_store.go
```diff
@@ -0,0 +1,221 @@
...
+	"github.com/pelletier/go-toml"
+	"github.com/spf13/afero"
+	"github.com/spf13/viper"
+)
+
+type ViperConfigStore struct {
+	viper     viper.Viper
+	configDir string
+	fs        afero.Fs
+}
```
andreaangiolillo, line 37:
```suggestion
// ViperConfigStore implements the config.Store interface
type ViperConfigStore struct {
	viper     viper.Viper
	configDir string
	fs        afero.Fs
}
```

internal/config/viper_store.go
```diff
@@ -0,0 +1,221 @@
...
+	return profileNames
+}
+
+func (s *ViperConfigStore) RenameProfile(oldProfileName string, newProfileName string) error {
+	if err := validateName(newProfileName); err != nil {
+		return err
+	}
+
+	// Configuration needs to be deleted from toml, as viper doesn't support this yet.
+	// FIXME :: change when https://github.com/spf13/viper/pull/519 is merged.
```
andreaangiolillo, line 135: [nit] can we add a cloudp ticket in backlog so we don't forget about this?

internal/config/viper_store.go
```diff
@@ -0,0 +1,221 @@
...
+	if _, err := f.WriteString(tomlString); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func (s *ViperConfigStore) DeleteProfile(profileName string) error {
+	// Configuration needs to be deleted from toml, as viper doesn't support this yet.
+	// FIXME :: change when https://github.com/spf13/viper/pull/519 is merged.
```
andreaangiolillo, line 167: 👀

internal/config/store.go
```diff
@@ -0,0 +1,108 @@
...
+			profileNames = append(profileNames, key)
+		}
+	}
+	// keys in maps are non-deterministic, trying to give users a consistent output
+	sort.Strings(profileNames)
+	return profileNames
+}
+
+func (*InMemoryStore) RenameProfile(_, _ string) error {
+	panic("not implemented")
```
andreaangiolillo, line 73: I am wondering what is the benefit here of using panic instead of returning the error as the function signature suggests to do 🤔

internal/config/viper_store.go
```diff
@@ -0,0 +1,221 @@
...
+	if err != nil {
+		return err
+	}
+	if !exists {
+		if err := s.fs.MkdirAll(s.configDir, defaultPermissions); err != nil {
+			return err
+		}
+	}
+
+	return s.viper.WriteConfigAs(s.Filename())
```
andreaangiolillo, line 112: I think is more readable but feel free to ignore 
```suggestion
	if exists {
               return s.viper.WriteConfigAs(s.Filename())
	}

	   return s.fs.MkdirAll(s.configDir, defaultPermissions)
```

andreaangiolillo (approved):
LGTM. All my comments are not blocking. Great improvement! 💯

internal/config/viper_store.go
jeroenvervaeke: The comment?

internal/config/viper_store.go
jeroenvervaeke: Updated

internal/config/viper_store.go
jeroenvervaeke: Updated

internal/config/viper_store.go
jeroenvervaeke: This comment was there since 2018, I didn't want to change the code to make the PR easier to review.

internal/config/store.go
jeroenvervaeke: `InMemoryStore` is not meant to be used, it's just there to keep tests working. The ultimate goal is to get rid of it. So if someone creates a new test they will realize they shouldn't use the method.

internal/config/viper_store.go
andreaangiolillo: yes, I was wondering if you forgot to remove it or if it was needed

internal/config/viper_store.go
jeroenvervaeke: Will create a ticket to address the actual logic, right now it's just a copy pasted

internal/config/viper_store.go
andreaangiolillo: ah wow, I am wondering if it is still valid 😅

internal/validate/validate_test.go
jeroenvervaeke: Done, thanks for the suggestion!

internal/validate/validate_test.go
jeroenvervaeke: Done

internal/config/viper_store.go
jeroenvervaeke: Fixed

andreaangiolillo (approved):

fmenezes (approved):

PR #4054, opened 2025-07-24
Removed defaultMongoDBVersion api call in builders. MongoDB version i…

jeroenvervaeke:
## Proposed changes

- Got rid of api call in the `internal/cli/clusters/create.go` builder.
- Had to introduce `usage.MDBVersionCreate`, because I didn't want the `update` commands to be affected by the docs change.

Made every single api call command 100% faster 🚀 

_Jira ticket:_ CLOUDP-333404


jeroenvervaeke:
> @jeroenvervaeke did you test this against prod? I vaguely remember we had to add a number instead of blank because our tests / prod / dev all behave differently.

I've tested against QA.
Sending a blank is not working indeed, but `null` should work (it's a `*string`, not `string`, hence my update).

According to the prod api docs:
> MongoDB major version of the cluster. Set to the binary major version.
>
> On creation: Choose from the available versions of MongoDB, or leave unspecified for the current recommended default in the MongoDB Cloud platform. The recommended version is a recent Long Term Support version. The default is not guaranteed to be the most recently released version throughout the entire release cycle. For versions available in a specific project, see the linked documentation or use the API endpoint for [project LTS versions endpoint](https://www.mongodb.com/docs/api/doc/atlas-admin-api-v2/operation/operation-createcluster#tag/Projects/operation/getProjectLtsVersions).

jeroenvervaeke:
Note to future self: This PR could cause an issue when releasing MongoDB 9.0, but this problem should be resolved by merging https://github.com/mongodb/mongodb-atlas-cli/pull/4050

fmenezes:
if you want you can update the snaphots by adding `update-snaphots` label, I think we will need to do it on master at least

xargom (approved):

fmenezes (approved):

PR #4061, opened 2025-07-29
CLOUDP-330235: Update config to support service accounts

cveticm:
## Proposed changes

* Adds support for **new profile attributes**:
  * `clientID`: Service Account client ID
  * `clientSecret`: Service Account client secret
  * `authType`: string representing the type of authentication used by the profile
* Adds support for **new global config attribute** `version`.
  * This version is not CLI version but version of the config. This is to support migrations when config logic is changed.
* Added **config migration** to loadConfig. After the config and profile are loaded in:
  * Checks if config is already version == 2
    * Returns
  * Calculates and sets `authType` for every profile
    * Saves config
  * This migration will be expanded in future to migrate credentials to secure storage. There is no concern introducing the migration in two steps as the migration will not be released until fully complete.
* Sets `authType` on `atlas auth login`
* Replaces all instances of checking authentication type by seeing what credentials are available with checking `authType`
  * To avoid breaking users who cannot successfully migrate and users using env vars, on the prerun of the `atlas` cmd,  the authType is set according the credentials available if it is nil.

_Jira ticket:_ [CLOUDP-330235](https://jira.mongodb.org/browse/CLOUDP-330235)

## Checklist

### DIFF
--- a/internal/cli/profile.go
+++ b/internal/cli/profile.go
@@ -25,6 +25,7 @@ import (
 var errUnsupportedService = errors.New("unsupported service")
 
 func InitProfile(profile string) error {
+	initAuthType()
 	if profile != "" {
 		return config.SetName(profile)
 	} else if profile = config.GetString(flag.Profile); profile != "" {
@@ -38,4 +39,22 @@ func InitProfile(profile string) error {
 	}
 
 	return nil
+}
+
+// initAuthType initializes the authentication type based on the current configuration.
+// If the user has set credentials via environment variables and has not set
+// 'MONGODB_ATLAS_AUTH_TYPE', it will set the auth type accordingly.
+func initAuthType() {
+	// If the auth type is already set, we don't need to do anything.
+	authType := config.AuthType()
+	if authType != "" {
+		return
+	}
+	// If the auth type is not set, we try to determine it based on the available credentials.
+	if config.PrivateAPIKey() != "" && config.PublicAPIKey() != "" {
+		config.SetAuthType(config.APIKeys)
+	}
+	if t, _ := config.Token(); t != nil {
+		config.SetAuthType(config.UserAccount)
+	}
 }
--- a/internal/cli/profile.go
+++ b/internal/cli/profile.go
@@ -54,7 +54,10 @@ func initAuthType() {
 	if config.PrivateAPIKey() != "" && config.PublicAPIKey() != "" {
 		config.SetAuthType(config.APIKeys)
 	}
-	if t, _ := config.Token(); t != nil {
+	if config.AccessToken() != "" && config.RefreshToken() != "" {
 		config.SetAuthType(config.UserAccount)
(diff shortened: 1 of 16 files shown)


internal/cli/setup/setup_cmd.go
```diff
@@ -589,7 +589,7 @@ func (opts *Opts) PreRun(ctx context.Context) error {
 
 	// if profile has access token and refresh token is valid, we can skip login
 	if _, err := auth.AccountWithAccessToken(); err == nil {
-		if err := opts.login.RefreshAccessToken(ctx); err != nil && !commonerrors.IsInvalidRefreshToken(err) {
+		if err := opts.login.RefreshAccessToken(ctx); !commonerrors.IsInvalidRefreshToken(err) {
```
cveticm, line 592: drive-by fix: should only skip this block if there is an error and the error is InvalidRefreshToken.

internal/config/profile.go
```diff
@@ -559,3 +589,15 @@ func SetLocalDeploymentImage(v string) { Default().SetLocalDeploymentImage(v) }
...
+
+// GetVersion returns the configuration version.
+func GetVersion() int64 { return Default().GetVersion() }
+func (p *Profile) GetVersion() int64 {
+	return p.GetInt64(Version)
+}
+
+// SetPublicAPIKey sets the configuration version.
+func SetVersion(v int64) { Default().SetVersion(v) }
+func (*Profile) SetVersion(v int64) {
```
fmenezes, line 601: we should have some protection against setting version at least reject it on `atlas config set` and validate on startup (maybe leave a warning `version not valid`?)

internal/cli/profile.go
```diff
@@ -25,6 +25,7 @@ import (
 var errUnsupportedService = errors.New("unsupported service")
 
 func InitProfile(profile string) error {
+	initAuthType()
```
fmenezes: clever, but I think this should happen after setting the name

cveticm: Good point, done 👍

internal/config/profile.go
cveticm: Added [protection against setting version](https://github.com/mongodb/mongodb-atlas-cli/pull/4061/files#diff-8d7e1225954a1c1b5df6495ab88dee69f678e30f0b5ed2615dd0ee7eee8a19fa). Also added [validation check](https://github.com/mongodb/mongodb-atlas-cli/pull/4061/files#diff-9b1ff1e58184b413e882ec267393f4cce4d67046ebc31dff4d8a5167cf1fa069) after [config is loaded](https://github.com/mongodb/mongodb-atlas-cli/pull/4061/files#diff-1ac01942db13bed1049ed82006abf3981b1f5b18e8e9576d3c7c7f3135756172). This will return an error if authType is present in an old config version or authType is not present for version 2 config.

PR #4096, opened 2025-08-06
CLOUDP-329793: Replace L1 transport

cveticm:
## Proposed changes
* decouples httpClient method from Store
* L1 call to httpClient instead of create their own client
* Removes api/transport

**Note on unauth error:**
If SA credentials are incorrect for example, an error is encountered while trying to set up token. Unlike with API Keys and User Accounts, SA will fail before an API call is made.
This results in different outputs
API Keys stdout:
```
{
  "error" : 401,
  "reason" : "Unauthorized",
  "detail" : "You are not authorized for this resource."
}Error: server returned an error response code
```
 Service Accounts stderr:
```
Error: this action requires authentication

To log in using your Atlas username and password or to set credentials using API keys, run: atlas auth login
```

This is not a concern as both scenarios return exit status 1. Thus the error is programatically visible to users either way.

_Jira ticket:_ [CLOUDP-329793](https://jira.mongodb.org/browse/CLOUDP-329793)

## Checklist


internal/transport/transport.go
```diff
@@ -114,17 +113,13 @@ func (tr *tokenTransport) RoundTrip(req *http.Request) (*http.Response, error) {
...
 		cfg.TokenURL = config.OpsManagerURL() + "api/oauth/token"
 	}
-
-	ctx := context.Background()
-
-	return &oauth2.Transport{
-		Base:   base,
-		Source: cfg.TokenSource(ctx),
-	}, nil
+	return cfg.Client(context.Background())
```
cveticm, line 124: drive-by: Since we already use atlas-sdk for `clientcredentials.NewConfig()`, it is clearer if we continue to use it to create the client. This method `Client()` returns a client with Transport same as before. Behaviour is the same.

internal/store/store.go
```diff
@@ -57,28 +57,25 @@ type Store struct {
 	ctx            context.Context
 }
 
-func (s *Store) httpClient(httpTransport http.RoundTripper) (*http.Client, error) {
-	switch s.authType {
+func HTTPClient(authType config.AuthMechanism, publicKey, privateKey string,
```
jeroenvervaeke: Can we accept an interface here instead of a long list of arguments?

PR #4112, opened 2025-08-08
 CLOUDP-329800: reuse logout and deprecate config delete 

blva:
## Proposed changes

_Jira ticket:_ CLOUDP-#

Closes #[issue number]

## Checklist


internal/cli/config/delete.go
```diff
@@ -29,31 +31,38 @@ type DeleteOpts struct {
...
 	cmd := &cobra.Command{
-		Use:     "delete <name>",
-		Aliases: []string{"rm"},
-		Short:   "Delete a profile.",
-		Args:    require.ExactArgs(1),
+		Use:        "delete <name>",
+		Aliases:    []string{"rm"},
+		Short:      "Delete a profile.",
+		Args:       require.ExactArgs(1),
+		Deprecated: "Please use the 'atlas auth logout' command instead.",
```
jwilliams-mongo, line 65: Q: if the page is deleted, will users actually see this deprecation message? perhaps it is displayed in help output? Usually when we deprecate features we keep the page around for a while but add a banner so users get advance guidance on how they need to adjust their workflows accordingly.

My recommendation would be to keep this page and we can add a banner on our end to give users a heads-up. What are your thoughts?

jwilliams-mongo (commented):
left a suggestion for keeping the page around for a while for the deprecated command, but LGTM otherwise

internal/cli/config/delete.go
blva: Hey @jwilliams-mongo , yes the deprecation warning would show up on call and on help.

Unfortunately, the removal is inherited from the cobra framework, I'll check if they have any way to override it but if they don't it might not be worth the effort to keep the page.

internal/cli/config/delete.go
```diff
@@ -29,31 +31,38 @@ type DeleteOpts struct {
...
+	executable, err := os.Executable()
+	if err != nil {
+		return fmt.Errorf("failed to get executable path: %w", err)
 	}
 
-	if err := config.SetName(opts.Entry); err != nil {
-		return err
-	}
+	// Execute: atlas auth logout --profile <profile-name> --force
+	cmd := exec.Command(executable, "auth", "logout", "--profile", opts.Entry, "--force")
```
cveticm, line 42: Why `--force`? 
Both commands prompt for confirmation. I think `--force` should only be passed if the user has set the flag

blva: I'm keeping the config prompt, so to avoid two prompts I'm using force in the logout one, but I could remove the prompt from config and use logout, no strong opinion

cveticm: Aha makes sense. No preference either on which command prompts. Thanks for explanation!

fmenezes: let's remove prompt and blindly use the logout one, in case we want to update prompts we just update one command

internal/cli/config/delete.go
jwilliams-mongo: yeah if it's too much effort I'm fine with moving forward with the removal

blva: I checked it and it seems that the removal happens at cobra level. I'll chat w/ the team to see if we can find a workaround, but meanwhile I think we'll keep it this way

internal/config/viper_store.go
```diff
@@ -165,7 +165,7 @@ func (s *ViperConfigStore) RenameProfile(oldProfileName string, newProfileName s
 func (s *ViperConfigStore) DeleteProfile(profileName string) error {
 	// Configuration needs to be deleted from toml, as viper doesn't support this yet.
 	// FIXME :: change when https://github.com/spf13/viper/pull/519 is merged.
-	settings := viper.AllSettings()
+	settings := s.viper.AllSettings()
```
blva, line 168: bugfix - without this, we were deleting the whole profile on logout

internal/cli/config/delete.go
blva: I'll address this in a follow up PR

cveticm (approved):
LGTM! 🚀

PR #4178, opened 2025-09-02
Chore: removes config files covered in core library

cveticm:
## Proposed changes
These files have already been removed and all uses replaced by core. \
They have likely come back during some merge to or from master.


blva (approved):
thanks for catching!

PR #4195, opened 2025-09-08
fix: config describe prints secure properties with value redacted

cveticm:
## Proposed changes

Config describe should print secure properties (`public_api_key`,`private_api_key`, etc.) with value redacted.

### How it works
* Create a new Store (configStore) without reading in env vars
* Get a map of all the insecurely stored profile properties using configStore
* For all secure properties, get the property value using configStore and include it in the map with value redacted
  * No additional logic is needed if the secure property is held insecurely or securely. configStore will call to the appropriate place. And, as configStore does not take into account env vars, these will not cause any trouble either.

### Design decisions
1. Checking for all secure properties, despite the set `auth_type`
If a user has managed to configure their profile in a way that it has multiple authentication type credentials, this should be reflected in `config describe` to easier help the user diagnose any auth issues this might cause

2. Redacted text
To make clear to the user where a secure property is being held, the redacted text includes its source:
`[redacted - source: secure storage]`
`[redacted - source: config file]`

example output:
```
// before - secure properties not shown at all
bin/atlas config describe sa   
SETTING           VALUE
auth_type         service_account
org_id            670cd17af33cea212ea1ed61
output            plaintext
project_id        670e34d35a4f587387db2102
service           cloud

//after - secure properties show but redacted
bin/atlas config describe sa   
SETTING           VALUE
auth_type         service_account
client_id         [redacted - source: secure storage]
client_secret     [redacted - source: secure storage]
org_id            670cd17af33cea212ea1ed61
output            plaintext
project_id        670e34d35a4f587387db2102
service           cloud

### DIFF
--- a/build/package/purls.txt
+++ b/build/package/purls.txt
@@ -75,7 +75,7 @@ pkg:golang/github.com/mholt/archives@v0.1.3
 pkg:golang/github.com/mikelolasagasti/xz@v1.0.1
 pkg:golang/github.com/minio/minlz@v1.0.0
 pkg:golang/github.com/mongodb-forks/digest@v1.1.0
-pkg:golang/github.com/mongodb/atlas-cli-core@v0.0.0-20250909090003-0d467f64c631
+pkg:golang/github.com/mongodb/atlas-cli-core@v0.0.0-20250909113945-ffb322e05f59
 pkg:golang/github.com/montanaflynn/stats@v0.7.1
 pkg:golang/github.com/nwaples/rardecode/v2@v2.1.0
 pkg:golang/github.com/pelletier/go-toml/v2@v2.2.3
--- a/go.mod
+++ b/go.mod
@@ -29,7 +29,7 @@ require (
 	github.com/mattn/go-isatty v0.0.20
 	github.com/mholt/archives v0.1.3
 	github.com/mongodb-labs/cobra2snooty v1.19.1
-	github.com/mongodb/atlas-cli-core v0.0.0-20250909090003-0d467f64c631
+	github.com/mongodb/atlas-cli-core v0.0.0-20250909113945-ffb322e05f59
 	github.com/pelletier/go-toml v1.9.5
 	github.com/pkg/browser v0.0.0-20240102092130-5ac0b6a4141c
 	github.com/shirou/gopsutil/v4 v4.25.8
--- a/go.sum
+++ b/go.sum
@@ -320,8 +320,8 @@ github.com/mongodb-forks/digest v1.1.0 h1:7eUdsR1BtqLv0mdNm4OXs6ddWvR4X2/OsLwdKk
 github.com/mongodb-forks/digest v1.1.0/go.mod h1:rb+EX8zotClD5Dj4NdgxnJXG9nwrlx3NWKJ8xttz1Dg=
 github.com/mongodb-labs/cobra2snooty v1.19.1 h1:GDEQZWy8f/DeJlImNgVvStu6sgNi8nuSOR1Oskcw8BI=
 github.com/mongodb-labs/cobra2snooty v1.19.1/go.mod h1:Hyq4YadN8dwdOiz56MXwTuVN63p0WlkQwxdLxOSGdX8=
-github.com/mongodb/atlas-cli-core v0.0.0-20250909090003-0d467f64c631 h1:193LDba8bvkFfGiGWqFyNyHLazN/U5pV7x/xgEGx3RE=
-github.com/mongodb/atlas-cli-core v0.0.0-20250909090003-0d467f64c631/go.mod h1:QDfVGpdfxXM1httLNXCKsfWTKv6slzCqBZxkkPIktlQ=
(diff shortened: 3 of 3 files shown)

### CODE REVIEW

File: internal/cli/config/describe.go [hunk 1]
```diff
@@ -32,6 +33,47 @@ var descTemplate = `SETTING	VALUE{{ range $key, $value := . }}
 {{$key}}	{{$value}}{{end}}
 `
 
+const redacted = "redacted"
+
+// AddSecureProperties adds secure properties to the map with "redacted" value
+// if they are available in the config.
+func (opts *describeOpts) AddSecureProperties(m map[string]string) (map[string]string, error) {
```
blva: [q] is there a way can we expand this on `atlas config edit` since that one opens the config file? cc @jeroenvervaeke

cveticm: Looking at the code, `atlas config edit` opens the actual config file (with vi or similar). We'd have to entirely refactor how the command works to support this.

blva: I think this is the tradeoff of adding it at the config level, having to re-add the `redacted` values. I found myself confused before when trying to find out if my credentials were correct or not and saw that it was not in the file at `atlas config edit`. 

I'm good with this change anyways since it fixes config describe but I think we should probably discuss if we're happy to keep atlas config edit as is or not in the team channel

blva (approved):

internal/cli/config/describe.go
cveticm: Agreed on discussing if followup work is needed for config edit 👍 

I think we could change the command to 
1. Open a copy of the config file that we populate with redacted secure properties. 
2. When a user is finished, we'd run the migration logic on the file and replace the actual config with the copy. 

This is just the first solution that came to mind, happy to discuss more with the team tomorrow during sync

blva: yep if we have a list of options that would be awesome to chat w/ the team async, my stamp is 🚢 anyways

internal/cli/config/describe.go
```diff
@@ -32,6 +33,35 @@ var descTemplate = `SETTING	VALUE{{ range $key, $value := . }}
...
+	// Check if secure storage is available
+	configStore, err := config.NewDefaultStore()
+	if err != nil {
+		return nil, err
+	}
+	if !configStore.IsSecure() {
```
jeroenvervaeke: [q] When we use the insecure store, we still want to redact the secrets. No?

internal/cli/config/describe.go
```diff
@@ -32,6 +33,35 @@ var descTemplate = `SETTING	VALUE{{ range $key, $value := . }}
...
+		return m, nil
+	}
+	serviceName := servicePrefix + opts.name
+
+	// We are using a keyring client directly here to avoid printing env vars
+	secureKeyring := secure.NewDefaultKeyringClient()
+	for _, key := range config.SecureProperties {
```
jeroenvervaeke, line 52: Thanks for this change!

internal/cli/config/describe.go
jeroenvervaeke: I don't think we should alter `atlas config edit`, since the secrets are not saved in the config I don't think we should re-add them and say `[redacted]`.

I think that approach will make things even more confusing and give customers the impression that our credentials are stored in the config, which is not true.

internal/cli/config/describe.go
```diff
@@ -32,6 +33,35 @@ var descTemplate = `SETTING	VALUE{{ range $key, $value := . }}
...
 `
 
+const (
+	redacted      = "redacted"
+	servicePrefix = "atlascli_"
+)
+
+// AddSecureProperties adds secure properties to the map with "redacted" value
+// if they are available in the config.
+func (opts *describeOpts) AddSecureProperties(m map[string]string) (map[string]string, error) {
```
jeroenvervaeke: Can't this entire function be written as follows?

```go
func (opts *describeOpts) AddSecureProperties(m map[string]string, ctx context.Context) (map[string]string, error) {
	for _, key := range config.SecureProperties {
		if v, found := m[key]; found && v != "" {
			m[key] = redacted
		}
	}

	return m, nil
}
```

jeroenvervaeke (changes requested):

internal/cli/config/describe.go
cveticm: This is already done by the function config.Map(), [see here](https://github.com/mongodb/atlas-cli-core/blob/a78a6e497eaf3235e2293aef47693d514bccc102/config/profile.go#L579)

internal/cli/config/describe.go
cveticm: Map `m` is only the values in `config.toml`. Thus we have to populate the map with the additional properties held in secure storage.

internal/cli/config/describe_test.go
jeroenvervaeke, line 1: Thanks for increasing the test coverage 🙏

jeroenvervaeke (approved):

PR #4233, opened 2025-09-24
CLOUDP-345026: Update the L1 generation tool to support x-xgen-operation-id-override

Luke-Sanderson:
## Proposed changes

_Jira ticket:_ [CLOUDP-345026](https://jira.mongodb.org/browse/CLOUDP-345026)

- Adds support for x-xgen-operation-id-override
- Atlascli specific overrides take priority over x-xgen-operation-id-override
- Flyby: Add instructions for making changes to the API Generator to `contributing.md`
## Checklist


tools/cmd/api-generator/convert_commands.go
```diff
@@ -101,37 +101,67 @@ func extractSunsetDate(extensions map[string]any) *time.Time {
...
+	// Only add original operation ID to aliases if there's no Atlas CLI operation ID override
+	overrides := extractOverrides(operation.Extensions)
+	if overrides == nil {
+		ext.operationAliases = append(ext.operationAliases, strcase.ToLowerCamel(operation.OperationID))
+		return
+	}
+
+	if _, hasOverride := overrides["operationId"].(string); !hasOverride {
+		ext.operationAliases = append(ext.operationAliases, strcase.ToLowerCamel(operation.OperationID))
+	}
```
cveticm: Nit: code readability
```suggestion
// if overrides exist and there is an operationid override, skip
	if overrides := extractOverrides(operation.Extensions); overrides != nil {
	  if _, hasOverride := overrides["operationId"].(string); hasOverride {
	    return
	  }
	}
		ext.operationAliases = append(ext.operationAliases, strcase.ToLowerCamel(operation.OperationID))
```

tools/cmd/api-generator/convert_commands.go
```diff
@@ -170,6 +201,7 @@ func operationToCommand(now time.Time, path, verb string, operation *openapi3.Op
 	if overrides := extractOverrides(operation.Extensions); overrides != nil {
 		if overriddenOperationID, ok := overrides["operationId"].(string); ok && overriddenOperationID != "" {
 			operationID = overriddenOperationID
+			shortOperationID = ""
```
cveticm: nit: Wondering could we just not set `extensions.shortOperationID` in `func processShortOperationIDOverride` instead of here since we are already doing some logic for if an atlasCLI opId override exists ([referring to here](https://github.com/mongodb/mongodb-atlas-cli/pull/4233/files#diff-a9e532d41c33e1fd55f7260388868185093e49d6caea16e6b58bd5e1d6c0ff10R156))

tools/cmd/api-generator/convert_commands.go
cveticm: Re on [this nit](https://github.com/mongodb/mongodb-atlas-cli/pull/4233/files#r2379646085), you could set `ext.shortOperationID` here, not setting a value if an atlasCLI opId override exists.

internal/api/commands.go
```diff
@@ -12350,8 +12506,9 @@ For more information and examples, see: https://www.mongodb.com/docs/atlas/cli/c
 				},
 			},
 			{
-				OperationID: `listIdentityProviders`,
-				Aliases:     nil,
+				OperationID:      `listIdentityProviders`,
+				ShortOperationID: ``,
+				Aliases:          nil,
```
cveticm, line 12511: I think the [spec](https://github.com/mongodb/mongodb-atlas-cli/blob/master/tools/internal/specs/spec.yaml) needs to be updated. Looking at the latest [openapi](https://raw.githubusercontent.com/mongodb/openapi/refs/heads/main/openapi/v2.json), `"x-xgen-operation-id-override": "listIdentityProviders"`
I expect here to see:
```
OperationID:      `listFederationSettingIdentityProviders `,
ShortOperationID: `listIdentityProviders`,
Aliases:          []string{`listFederationSettingIdentityProviders`},
```
(see this [make command](https://github.com/mongodb/mongodb-atlas-cli/blob/master/Makefile#L192). This [github workflow](https://github.com/mongodb/mongodb-atlas-cli/blob/master/.github/workflows/autoupdate-spec.yaml) may also help)

internal/cli/api/api.go
```diff
@@ -99,6 +99,14 @@ func createAPICommandGroupToCobraCommand(group shared_api.Group) *cobra.Command
 func convertAPIToCobraCommand(command shared_api.Command) (*cobra.Command, error) {
 	// command properties
 	commandName := strcase.ToLowerCamel(command.OperationID)
+	commandAliases := command.Aliases
+
+	// Prefer to use shortOperationID if present
+	if command.ShortOperationID != "" {
+		commandName = strcase.ToLowerCamel(command.ShortOperationID)
+		commandAliases = append(commandAliases, strcase.ToLowerCamel(command.OperationID))
```
jeroenvervaeke: Small nit, but why not re-use the variable you created in the previous line, makes it easier to follow:
```suggestion
		commandAliases = append(commandAliases, commandName)
```

tools/cmd/api-generator/convert_commands.go
```diff
@@ -101,37 +101,67 @@ func extractSunsetDate(extensions map[string]any) *time.Time {
...
+	if !ok || shortOperationID == "" {
+		return
+	}
+
+	ext.shortOperationID = shortOperationID
+
+	// Only add original operation ID to aliases if there's no Atlas CLI operation ID override
+	overrides := extractOverrides(operation.Extensions)
+	if overrides == nil {
+		ext.operationAliases = append(ext.operationAliases, strcase.ToLowerCamel(operation.OperationID))
```
jeroenvervaeke: Why do we add the original OperationID to the extensions?

AFAIK, the goal of these methods is to extract the extensions only and apply the business logic later.
Running business logic here (adding `operation.OperationID` to aliases) should happen outside of the `extractExtensionsFromOperation` method family.

internal/api/commands.go
Luke-Sanderson: I think this will be covered by [CLOUDP-347637](https://jira.mongodb.org/browse/CLOUDP-347637) ([Failed github workflow run](https://github.com/mongodb/mongodb-atlas-cli/actions/runs/17940433093/job/51015517453) for context). 

We might need to merge the other ticket in before this one?

internal/cli/api/api.go
Luke-Sanderson: That wouldn't work because commandName was updated to be the shortOperationID and we want to add the original OperationID to the alias but i refactored to hopefully make that more clear

tools/cmd/api-generator/convert_commands.go
Luke-Sanderson: Refactored to keep `extractExtensionsFromOperation` only for extracting the extension information

tools/cmd/api-generator/convert_commands.go
Luke-Sanderson: No longer applicable due to other refactor

tools/cmd/api-generator/convert_commands.go
Luke-Sanderson: Removed this check due to it being a duplicate

tools/cmd/api-generator/convert_commands.go
```diff
@@ -198,11 +175,8 @@ func operationToCommand(now time.Time, path, verb string, operation *openapi3.Op
 		return nil, fmt.Errorf("failed to clean description: %w", err)
 	}
 
-	if overrides := extractOverrides(operation.Extensions); overrides != nil {
-		if overriddenOperationID, ok := overrides["operationId"].(string); ok && overriddenOperationID != "" {
-			operationID = overriddenOperationID
-			shortOperationID = ""
-		}
+	if shortOperationID != "" {
```
jeroenvervaeke: Do we need this logic? Since we already to the same thing in `cli/api`?

Luke-Sanderson: We do not and I have now removed it

tools/cmd/api-generator/convert_commands.go
```diff
@@ -167,21 +175,16 @@ func operationToCommand(now time.Time, path, verb string, operation *openapi3.Op
 		return nil, fmt.Errorf("failed to clean description: %w", err)
 	}
 
-	if overrides := extractOverrides(operation.Extensions); overrides != nil {
```
blva, line 170: [q] for my own understanding, we have removed `operationId` overrides officially in benefit of `x-xgen-operation-id-override`? I'm trying to understand if we haven't missed any overrides here that should've been translated to `x-xgen-operation-id-override`

Luke-Sanderson: There is no mention of replacing the `operationId` override officially by `x-xgen-operation-id-override` in the TD so I think I removed this my mistake. Will add it back in

blva: but does it make sense to keep two types of overrides? I think my main concern is just making sure we're handling both if they exist, but we could look into how the overrides are being added at the moment and if we should rename them, or maybe, we don't even use them

Luke-Sanderson: Since both target the same field (operationID) one will override the other. From what I can see in the latest openapi spec, I don't think `operationID` override is used. It was appear as `x-operationid` right?

blva: interesting! I think this is the AtlasCLI framework for overrides, e.g. search for the openapi spec tests with `x-xgen-atlascli`. There is a different format there for different overrides. Perharps by keeping this here, the idea is that if we need to override via AtlasCLI we can, let's drop a comment here maybe capturing this behaviour in case we see this in the future?

Luke-Sanderson: Oh yes this is the `x-xgen-atlascli` one. I think I might have originally removed it because I thought it was duplicating the logic from [here](https://github.com/mongodb/mongodb-atlas-cli/pull/4233/files#diff-a9e532d41c33e1fd55f7260388868185093e49d6caea16e6b58bd5e1d6c0ff10R133). Yes I agree there is a use case for this override for overriding the operationId only for atlascli and I think that this override should take priority. 

I will add a comment to reflect this

cveticm (approved):

PR #4340, opened 2025-11-26
chore: Minor fix to update SDK

cveticm:
## Proposed changes

- Issue 1: 
  - If no SDK update, action always fails on task `auto merge`
  Solution:
    - Add caveat that auto-merge is only set if there is an sdk update
- Issue 2:
  - If sdk update results in changes that cause build failures, task `make gen-purls` will fail, preventing the PR from being opened
  Solution:
    - Move this task to after PR is opened. This provides greater visual that an SDK update is available but has changes that cause failure.


blva (approved):
LGTM, thanks for the fix!

PR #4383, opened 2026-01-07
VULN-1034: Fix nil pointer dereference in deployments start

blva:
## Description

Fixes a nil pointer dereference bug in `internal/cli/deployments/start.go`.

The condition `if err != nil` was incorrectly accessing `clusterAutoScalingConfig.GetAutoScalingMode()` when the API call **failed**, which would cause a panic since the config would be nil on error.

Changed to `if err == nil` to only access the config when the API call succeeds.

## JIRA
[VULN-1034](https://jira.mongodb.org/browse/VULN-1034)


wtrocki:
evergreen refresh

PR #4388, opened 2026-01-12
chore: Sunset messages should take precident over deprecation messages

cveticm:
## Proposed changes

Issue: All sunset versions are marked as deprecated. Thus the deprecated message would be printed and return before ever checking if sunset message needs to be printed.

Solution: 
1. Handle sunset message before deprecated
2. Fix unit test to set `Deprecated` to `true` when `Sunset` is set

Before running sunset version of a cmd:
```
➜ bin/atlas api alerts acknowledgeGroupAlert --version 2023-01-01
warning: version '2023-01-01' is deprecated. Consider upgrading to a newer version: 2024-05-30.
Error: required flag(s) "alertId" not set
```

After:
```
➜ bin/atlas api alerts acknowledgeGroupAlert --version 2023-01-01
warning: version '2023-01-01' is deprecated for this command and will be sunset on 2026-11-30. Consider upgrading to a newer version if available.
Error: required flag(s) "alertId" not set
```

### DIFF
--- a/internal/cli/api/api.go
+++ b/internal/cli/api/api.go
@@ -545,13 +545,25 @@ func printDeprecatedVersionWarning(apiCommand shared_api.Command, versionString 
 		return
 	}
 
+	// Print a warning or error if the version has a sunset date
+	if commandVersion.Sunset != nil {
+		sunsetDate := commandVersion.Sunset.Format("2006-01-02")
+		// if date is in the past, warn the user that it will not work
+		if commandVersion.Sunset.Before(time.Now()) {
+			fmt.Fprintf(os.Stderr, "error: version '%s' is deprecated for this command and has already been sunset since %s. Consider upgrading to a newer version if available.\n", *versionString, sunsetDate)
+			return
+		}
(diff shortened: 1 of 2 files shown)


internal/cli/api/api.go
```diff
@@ -545,13 +545,25 @@ func printDeprecatedVersionWarning(apiCommand shared_api.Command, versionString
...
+		if commandVersion.Sunset.Before(time.Now()) {
+			fmt.Fprintf(os.Stderr, "error: version '%s' is deprecated for this command and has already been sunset since %s. Consider upgrading to a newer version if available.\n", *versionString, sunsetDate)
+			return
+		}
+		fmt.Fprintf(os.Stderr, "warning: version '%s' is deprecated for this command and will be sunset on %s. Consider upgrading to a newer version if available.\n", *versionString, sunsetDate)
+		return
+	}
+
 	// Find the latest command version
 	latestCommandVersion, err := defaultAPIVersion(apiCommand)
```
blva, line 561: nit: we should probably also share the latest version in sunset cases then - but OK to do that as a follow-up after release - the warnings are already good enough for this release

PR #4393, opened 2026-01-13
CLOUDP-372584: Update serverless e2e tests to expect error from watch…

cveticm:
… and describe and empty from list commands

## Proposed changes

The Serverless Instance GET endpoints have been modified: https://jira.mongodb.org/browse/CLOUDP-367742
GET all returns empty list
GET one returns 404 NOT_FOUND

This resulted in our snapshots failing: https://jira.mongodb.org/browse/CLOUDP-372159

This PR resolves this issue by removing the entire Serverless Instance E2E test.
Why? This endpoint will be sunset on 2026-01-22 and thus removing testing coverage now is not concerning.
Follow-up ticket to remove Serverless L2 commands has been created: [CLOUDP-372984](https://jira.mongodb.org/browse/CLOUDP-372984) 

_Jira ticket:_ [CLOUDP-372584](https://jira.mongodb.org/browse/CLOUDP-372584)


wtrocki (dismissed):

test/e2e/atlas/serverless/instance/serverless/serverless_test.go
```diff
@@ -81,10 +81,9 @@ func TestServerless(t *testing.T) {
 		)
```
blva: [q] are all the endpoints being sunset? if yes, should we delete the tests? if not, I think it should be fine to delete the tests that throw 404 and only keep the ones that still work (and expect an empty result)

cveticm: All endpoints are being sunset next week BAR GET endpoints (singular and plural) which will be kept online and return 404s (and empty lists in the case of plural GET). 
Context: AKO users would otherwise be broken and forced to upgrade due to the nature of how the operator checks serverless instances.

I would be happy to remove the entire test if you are comfortable with the commands remaining untested for the final week of its life.

cveticm: Discussed offline, proceeding with removal of entire e2e test and opening follow-up ticket to remove and cleanup serverless isntance after sunset next week.

PR #4402, opened 2026-01-19
VULN-1042: replace golang.org/x/text v0.3.0 with v0.33.0

cveticm:
## Proposed changes

Eliminates vulnerable `golang.org/x/text@v0.3.0` dependency by adding a replace directive to redirect it to the secure `v0.33.0` version already in use.

### Changes
- Added `replace golang.org/x/text v0.3.0 => golang.org/x/text v0.33.0` directive
- Reorganized replace block for better maintainability
- Updated go.sum with additional module metadata

### Background
The vulnerable `golang.org/x/text@v0.3.0` was present in go.sum due to phantom dependencies from older transitive dependency versions, even though the project uses the secure `v0.33.0`. This follows the same proven approach used for VULN-1046 (#4395).

_Jira ticket:_ [VULN-1042](https://jira.mongodb.org/browse/VULN-1042)


blva (approved):

go.mod
```diff
@@ -2,8 +2,12 @@ module github.com/mongodb/mongodb-atlas-cli/atlascli
 
 go 1.24.2
 
-// Security fix for VULN-1046: Update yaml.v3 v3.0.0-20200313102051-9f266ea9e77c to v3.0.1
-replace gopkg.in/yaml.v3 v3.0.0-20200313102051-9f266ea9e77c => gopkg.in/yaml.v3 v3.0.1
+replace (
+	// Security fix for VULN-1042: Update golang.org/x/text v0.3.0 to v0.33.0
+	golang.org/x/text v0.3.0 => golang.org/x/text v0.33.0
```
blva, line 7: [q] who was the culprit for that dependency? just checking if we could upgrade it completely

cveticm: Numerous deps that are up-to-date
For example, `github.com/google/go-licenses/v2` used in the `tool` section of `go.mod` and its latest version also uses this vuln version of `x/text`.

Here is the results of asking Claude to create a list of direct deps that cause x/text@v0.3.0 appearing in the go.sum:
* github.com/google/go-licenses/v2 - up-to-date v still calls for x/text@v0.3.0
* google.golang.org/api@v0.259.0 - up-to-date
* cloud.google.com/go/kms@v1.23.2 - up-to-date
* github.com/mholt/archives@v0.1.5 - up-to-date

blva: got it! thanks for the details

PR #4446, opened 2026-02-11
feat(plugin): add MinVersion field to first-class plugins for auto-update

jeroenvervaeke:
## Proposed changes

First-class plugins now support a MinVersion field that triggers automatic updates when the installed version is below the specified minimum.

Changes:
- Add MinVersion field to FirstClassPlugin struct
- Implement needsUpdate() to compare installed version against MinVersion
- Auto-update plugins before execution when version is below minimum
- Skip installed plugin commands in favor of first-class handlers when update is needed
- Add getFirstClassPluginsNeedingUpdate() helper for command registration

Set initial MinVersion values:
- atlas-cli-plugin-kubernetes: 1.2.4
- atlas-local-plugin: 0.0.10

_Jira ticket:_ CLOUDP-379821

### DIFF
--- a/internal/cli/plugin/first_class.go
+++ b/internal/cli/plugin/first_class.go
@@ -17,6 +17,7 @@ package plugin
 import (
 	"fmt"
 
+	"github.com/Masterminds/semver/v3"
 	"github.com/mongodb/mongodb-atlas-cli/atlascli/internal/plugin"
 	"github.com/spf13/cobra"
 )
@@ -42,7 +43,8 @@ var FirstClassPlugins = []*FirstClassPlugin{
 	// 	},
 	// },
 	{
-		Name: "atlas-cli-plugin-kubernetes",
+		Name:       "atlas-cli-plugin-kubernetes",
+		MinVersion: "1.2.4",
 		Github: &Github{
 			Owner: "mongodb",
 			Name:  "atlas-cli-plugin-kubernetes",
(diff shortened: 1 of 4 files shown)


internal/cli/plugin/first_class.go
```diff
@@ -92,17 +96,30 @@ func IsFirstClassPluginCmd(cmd *cobra.Command) bool {
...
 
-func (fcp *FirstClassPlugin) runFirstClassPluginCommand(cmd *cobra.Command, args []string, plugins *plugin.ValidatedPlugins) error {
+func (fcp *FirstClassPlugin) needsUpdate(installedPlugin *plugin.Plugin) bool {
+	if fcp.MinVersion == "" {
+		return false
+	}
+
+	minVersion, err := semver.NewVersion(fcp.MinVersion)
+	if err != nil {
+		return false
```
andreaangiolillo, line 116: [nit] This means that if the minimum version is in the wrong format, you ignore the error and simply return false. Should we instead log a warning so users are aware of the formatting issue and can fix it?

jeroenvervaeke: This would only happen if we type an incorrect version number in our hardcoded config.

I think it's ok to silently fail here. But I can print a warning if you like.

internal/cli/plugin/first_class.go
```diff
@@ -42,7 +43,8 @@ var FirstClassPlugins = []*FirstClassPlugin{
 	// 	},
 	// },
 	{
-		Name: "atlas-cli-plugin-kubernetes",
+		Name:       "atlas-cli-plugin-kubernetes",
+		MinVersion: "1.2.4",
```
cveticm, line 47: q: Could we capture when we bump the min version in some documentation, CONTRIBUTING maybe?

PR #4512, opened 2026-03-25
CLOUDP-392089: fixes cluster not found in generic e2e tests

cveticm:
_Jira ticket:_ [CLOUDP-392089](https://jira.mongodb.org/browse/CLOUDP-392089)

## Context

The [following validation](https://github.com/10gen/mms/pull/160517) was added to the API on 2026-03-19. Since 2026-03-20, our generic E2E tests (both regular and gov) have been consistently failing with: `ATLAS_SCOPE_CLUSTER_NOT_FOUND`  `The scope cluster name "Cluster0" does not exist in this project.`
Evergreen waterfall for the [first failures](https://spruce.corp.mongodb.com/project/mongodb-atlas-cli-master/waterfall?minOrder=5726).

This validation has been [reverted](https://github.com/10gen/mms/pull/161337) on 2026-03-30 to address failing e2e tests but will be returned.

## Change
Since the validation will require the clusters defined in the scope to be actual running clusters, we will utilise the long living clusters available on both our test orgs (gov and non-gov).

This PR makes the generic E2E tests take dynamic cluster names for dbuser tests by:
* Introducing two env vars: `E2E_CLUSTER_1_NAME` and `E2E_CLUSTER_2_NAME`, used by the tests instead of hard-coded `Cluster0` / `Cluster1`.
* Wiring those vars into Evergreen so they are set to the actual long-lived cluster names in the dev E2E project.

As a part of this work, two long-living clusters have been created in the govdev testing org.

See the following [evg patch](https://spruce.corp.mongodb.com/version/69c6832844b0500007c62361/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC) which was ran before the validation was reverted for successful run.

#### Why not create clusters per run?
An alternative would be to provision the required clusters as part of the test setup. I've rejected that because:
* Cluster provisioning adds several minutes of latency per run 
  * Currently ~3 minutes just to get a healthy cluster. For comparison, the `atlas_deployments_atlas_clusters_e2e` task takes ~24 minutes.
  * Creating clusters per run would significantly slow down CI.
* Generic tests run on every PR, so per-run provisioning would increase resource usage in both dev and govdev.

By reusing two stable clusters and parameterizing their names, we keep tests fast, reduce load on dev/govdev, and make the tests tolerant of future renames.

## How to review
Due to the change of cluster names, I've updated snapshot tests. 

Please only review the following:
https://github.com/mongodb/mongodb-atlas-cli/compare/0d1b5f3...82a2b90


test/internal/env.go
```diff
@@ -35,6 +35,8 @@ const (
 	snapshotOrgID              = "a0123456789abcdef012345a"
 	snapshotProjectID          = "b0123456789abcdef012345b"
 	snapshotOpsManagerURL      = "http://localhost:8080/"
+	snapshotCluster1Name       = "Cluster0"
```
jeroenvervaeke: Wouldn't it be less confusing to call these `snapshotCluster0Name` and `snapshotCluster1Name`?

cveticm: Fair point, done!

test/internal/env.go
```diff
@@ -30,11 +30,13 @@ const (
 	cloudgov                   = "cloudgov"
 	snapshotCloudRoleID        = "c0123456789abcdef012345c"
 	snapshotTestBucket         = "test-bucket"
-	snapshotFlexInstanceName   = "test-flex"
+	snapshotFlexInstanceName   = "snapshotCluster0"
```
cveticm, line 33: `snapshotFlexInstanceName` and `snapshotCluster1Name` use the same real cluster during recording, some cluster `x`. Due to how we overwrite cluster names after recording snapshot tests, this results in all instances of `x` being renamed `test-flex` since that is the overwrite that is applied first. Then, snapshot tests using `snapshotCluster1Name` will fail since they expect a cluster name `snapshotCluster0` but get `test-flex` instead.
Unified the names to avoid this problem. Kept both consts for code readability.

PR #4560, opened 2026-04-29
CLOUDP-400180: treat Atlas sample data "continuing through error" as success

blva:
## Proposed changes

_Jira ticket:_ CLOUDP-400180

The Atlas sample data loading API can mark a job as `FAILED` with an error message prefixed `"continuing through error:"` even on a fresh cluster. This happens when Atlas's internal loading service encounters duplicate key errors during its own retry logic — the job completes and `completeDate` is set, but the state is `FAILED`. The CLI was surfacing this as a hard error, causing the `atlas_deployments_atlas_clusters_e2e` task to fail.

The fix detects this prefix in `sampleDataWatcher` and treats it as a successful completion instead of an error.

## Checklist

## Further comments

The four test cases in `TestOpts_SampleDataWatcher` cover: normal completion, still-in-progress, a genuine hard failure, and the `"continuing through error:"` case that was causing the flake.

### DIFF
--- a/internal/cli/setup/setup_cmd.go
+++ b/internal/cli/setup/setup_cmd.go
@@ -277,7 +277,11 @@ func (opts *Opts) sampleDataWatcher() (any, bool, error) {
 		return nil, false, err
 	}
 	if result.GetState() == "FAILED" {
-		return nil, false, fmt.Errorf("failed to load data: %s", result.GetErrorMessage())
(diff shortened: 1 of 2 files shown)


blva:
Closing this. The root cause was the education team adding a `config.settings` collection to the sample dataset, which broke mongorestore. Fixed on the Atlas side in CLOUDP-400349 — no CLI change needed.

PR #3966, opened 2025-06-16
CLOUDP-302696: Merge L3 commands and L2 create command

blva:
## Proposed changes

_Jira ticket:_ CLOUDP-302696

- Adds all reviewed PRs

Closes #[issue number]

## Checklist

- [x] I have signed the [MongoDB CLA](https://www.mongodb.com/legal/contributor-agreement)
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] I have added any necessary documentation in document requirements section listed in [CONTRIBUTING.md](https://github.com/mongodb/mongodb-atlas-cli/blob/master/CONTRIBUTING.md) (if appropriate)
- [ ] I have addressed the @mongodb/docs-cloud-team comments (if appropriate)
- [ ] I have updated [test/README.md](https://github.com/mongodb/mongodb-atlas-cli/blob/master/test/README.md) (if an e2e test has been added)
- [ ] I have run `make fmt` and formatted my code

PR #4583, opened 2026-05-19
fix: rename _iss.go files to _iss_test.go so go test picks them up

fmenezes:
## Summary

- Renames `deployments_atlas_test_iss.go` → `deployments_atlas_iss_test.go`
- Renames `setup_force_test_iss.go` → `setup_force_iss_test.go`

These files contain `TestXxx` functions but were not named with the `_test.go` suffix, so `go test` never registered or ran them — they were silently dead.

Both files were introduced in #3966 with a `//go:build e2e || ...` tag that prevented the issue from surfacing. The build tags were removed in #4089, which exposed the problem: without the tag, the regular (non-test) package variant tries to compile these files without the symbols defined in the accompanying `_test.go` files, causing `undefined` type errors.

Renaming to `_iss_test.go` fixes both issues: the tests are now actually run by `go test`, and the compiler sees them only in the test variant where their dependencies are available.

## Test plan

- [ ] `go build ./...` passes (files no longer compiled in regular build)
- [ ] `go test -tags e2e ./test/e2e/atlas/deployments/atlasclusters/deploymentsatlas/...` runs `TestDeploymentsAtlasISS`
- [ ] `go test -tags e2e ./test/e2e/atlas/interactive/setupforce/...` runs `TestSetupISS`


cveticm (approved):
lgtm

PR #4600, opened 2026-06-02
CLOUDP-405482: redact PEM private keys from e2e test snapshots

jeroenvervaeke:
## Proposed changes

_Jira ticket:_ [CLOUDP-405482](https://jira.mongodb.org/browse/CLOUDP-405482)

### Issue

GitHub secret scanning flagged a real RSA private key committed to the repo inside a test snapshot file. The `dbusers certs create` endpoint returns `text/plain` PEM data (certificate + private key). The snapshot recorder in `prepareSnapshot` only masked JSON responses, so the raw PEM — private key included — was written verbatim to disk and committed.

### Fix

`prepareSnapshot` now reads every response body (not just JSON), applies PEM private key redaction via regex before writing the snapshot, then applies the existing `maskString` and header masking. The regex covers PKCS8, RSA, and EC key types:

```
(?s)-----BEGIN [A-Z ]* PRIVATE KEY-----.*?-----END [A-Z ]* PRIVATE KEY-----
```

The existing snapshot file with the leaked key will be overwritten when snapshot recording is re-run for `TestDBUserCerts`.

### Fly-by

The pre-existing `resp.Header["Content-Length"]` direct map write was replaced with the idiomatic `resp.Header.Set(...)`. Body masking (`maskString`) and header masking are now applied to all response content types, not just JSON — this is an improvement since non-JSON responses (e.g. certs) can also contain project IDs or other identifiable values.

## Checklist

### DIFF
--- a/test/internal/atlas_e2e_test_generator.go
+++ b/test/internal/atlas_e2e_test_generator.go
@@ -45,6 +45,12 @@ import (
 const redactedToken = "redactedToken"
 const ipMax = 255
 
+var pemPrivateKeyRe = regexp.MustCompile(`(?s)-----BEGIN [A-Z ]*PRIVATE KEY-----.*?-----END [A-Z ]*PRIVATE KEY-----`)
+
+func redactPEMPrivateKeys(body []byte) []byte {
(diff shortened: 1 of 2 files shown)


blva (approved):
🧹💥

PR #4601, opened 2026-06-02
CLOUDP-410005: support paths with different variable names in gen-api-commands

jeroenvervaeke:
## Proposed changes

_Jira ticket:_ [CLOUDP-410005](https://jira.mongodb.org/browse/CLOUDP-410005)

The MongoDB Atlas OpenAPI spec defines two paths that are structurally identical — same URL pattern, only the path parameter names differ:

- `/api/atlas/v2/groups/{groupId}/sampleDatasetLoad/{name}` (POST)
- `/api/atlas/v2/groups/{groupId}/sampleDatasetLoad/{sampleDatasetId}` (GET)

These are genuinely conflicting from an OpenAPI perspective: any request to e.g. `/groups/abc/sampleDatasetLoad/xyz` matches both templates, so a router cannot distinguish them. `kin-openapi` correctly rejects the spec with:

```
Error: spec validation failed, error: invalid paths: conflicting paths
"/api/atlas/v2/groups/{groupId}/sampleDatasetLoad/{sampleDatasetId}" and
"/api/atlas/v2/groups/{groupId}/sampleDatasetLoad/{name}"
```

`kin-openapi` does not expose a validation option to disable this check.

**Fix:** before calling `spec.Validate()`, load a second copy of the spec and deduplicate structurally identical paths in that copy by merging their operations. Validation runs on the deduplicated copy; command generation runs on the original spec so parameter names are preserved.

## Checklist

- [x] I have added tests that prove my fix is effective or that my feature works (`TestDeduplicateConflictingPaths` verifies that two paths differing only in parameter name are merged into one path with all operations intact, preventing regression)

## Further comments

The `deduplicateConflictingPaths` function keeps the first path key and merges HTTP operations from any duplicate into it. If the spec ever gains more such pairs, the function handles them generically without further changes.

### DIFF
--- a/tools/cmd/api-generator/main.go
+++ b/tools/cmd/api-generator/main.go
@@ -23,6 +23,7 @@ import (
 	"io"
 	"os"
 	"reflect"
+	"regexp"
 	"slices"
 	"strings"
 	"text/template"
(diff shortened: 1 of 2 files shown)


Luke-Sanderson (approved):
LGTM

PR #4602, opened 2026-06-03
CLOUDP-392159: handle rate-limited MCP telemetry

jeroenvervaeke:
## Proposed changes

_Jira ticket:_ [CLOUDP-392159](https://jira.mongodb.org/browse/CLOUDP-392159)

The MMS backend enforces a rate limit on telemetry events (CLOUDP-387377), returning HTTP 429 when clients exceed the per-request threshold of 32 events. The CLI was sending all cached events in a single unbatched request, so a command run in a tight loop would accumulate events, trigger 429, and then retry the full set on every subsequent invocation — continuously spamming the backend.

**Batch cap:** `trackCommand` now sends at most 32 events per invocation. Any remainder is written back to the cache for the next run.

**Backoff:** On any send error, a sentinel file (`telemetry_backoff`) is created in the cache directory. Its mtime marks when the backoff window started. Subsequent invocations skip the HTTP send entirely for one minute and just append their event to the cache. On a successful send the file is removed. No content is stored in the file — existence + mtime is sufficient.

## Tests added

- `TestTrackCommandBatchLimit` — exactly 32 events sent when cache has 31 + 1 new; remaining 1 written back
- `TestTrackCommandRemainingEventsPersistedAfterSuccess` — overflow events correctly written back after a successful partial send
- `TestTrackCommandRateLimitedPreservesEvents` — all events (cached + new) survive a send failure
- `TestSaveAndIsBackedOff` — backoff file created and detected
- `TestRemoveBackoff` — backoff file removed on explicit clear
- `TestBackoffExpired` — expired backoff (past mtime) is not active
- `TestTrackCommandSkipsSendWhenBackedOff` — no HTTP call made during active backoff; event cached
- `TestTrackCommandSetsBackoffOnSendError` — backoff file created for both 429 and generic errors
- `TestTrackCommandClearsBackoffOnSuccess` — backoff file removed after successful send

## Checklist

### DIFF
--- a/internal/cli/output_opts.go
+++ b/internal/cli/output_opts.go
@@ -117,7 +117,7 @@ func isNil(o any) bool {
 	}
 	ot := reflect.TypeOf(o)
 	otk := ot.Kind()
-	switch otk { //nolint:exhaustive // clearer cod
+	switch otk { //nolint:exhaustive // only nilable kinds matter here
 	case reflect.Array, reflect.Slice, reflect.Map, reflect.Chan, reflect.Pointer, reflect.UnsafePointer, reflect.Interface:
 		return reflect.ValueOf(o).IsNil()
 	default:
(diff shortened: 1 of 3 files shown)


cveticm (approved):
LGTM. Cool solution with creating the backoff file to track time since last send!

PR #4628, opened 2026-06-26
CLOUDP-417280: Match TRAE_AI_SHELL_ID agent telemetry on presence

blva:
## Proposed changes

_Jira ticket:_ CLOUDP-417280

This PR is the first in the ticket to expend tracing support for AI agents.

`TRAE_AI_SHELL_ID` is now detected whenever the variable is set, rather than only when it holds a non-empty value. Its value is a session id, so presence alone is the agent signal.

This is implemented with a `presenceOnly` flag on the agent env var table, set only for `TRAE_AI_SHELL_ID`. The existing value-matching logic for every other agent is left unchanged.

## Checklist

## Further comments

Follow-up to CLOUDP-400902. The broader Bun/Vercel env var additions tracked in this ticket are not included here.

### DIFF
--- a/internal/telemetry/event.go
+++ b/internal/telemetry/event.go
@@ -152,26 +152,34 @@ func withCI() EventOpt {
 }
 
 var agentEnvVars = []struct {
-	envVar string
-	value  string // empty means any non-empty value matches
-	name   string
+	envVar       string
+	value        string // expected value; empty matches any non-empty value
+	name         string
+	presenceOnly bool // match whenever the variable is set, regardless of value
 }{
-	{"CLAUDECODE", "1", "claude_code"},
-	{"CURSOR_AGENT", "1", "cursor"},
-	{"GEMINI_CLI", "1", "gemini_cli"},
-	{"CODEX_SANDBOX", "seatbelt", "codex_cli"},
-	{"AUGMENT_AGENT", "1", "auggie_cli"},
-	{"CLINE_ACTIVE", "true", "cline"},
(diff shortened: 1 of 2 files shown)


internal/telemetry/event.go
```diff
@@ -152,26 +152,34 @@ func withCI() EventOpt {
...
-	{"AMP_AGENT", "1", "amp"},
-	{"GOOSE_AGENT", "1", "goose"},
+	{"CLAUDECODE", "1", "claude_code", false},
+	{"CURSOR_AGENT", "1", "cursor", false},
+	{"GEMINI_CLI", "1", "gemini_cli", false},
+	{"CODEX_SANDBOX", "seatbelt", "codex_cli", false},
+	{"AUGMENT_AGENT", "1", "auggie_cli", false},
+	{"CLINE_ACTIVE", "true", "cline", false},
+	{"OPENCODE_CLIENT", "1", "opencode_client", false},
+	{"TRAE_AI_SHELL_ID", "", "trae_ai", true},
```
cveticm, line 167: why are we only doing this for trae ai? Couldn't this behaviour be done for any of these other agents?

cveticm: That is, would just the presence of any of these env vars be an indicator of the agent, regardless of if its empty or not?

blva: It could, I think I'm being conservative in tracing the value of the environment variable to the expected value wherever it is possible. And TRAE will set an ID which is always different

cveticm (approved):

PR #4634, opened 2026-07-02
Add UserAccountConnect login option (OAuth authorization code flow)

jricher:
## Proposed changes

Adds authorization-code-based login to Atlas as a new `UserAccountConnect` option in the
`atlas login` menu. Authorization server metadata is discovered automatically (RFC 8414)
and cached; a new `--discover` flag forces re-discovery when needed.

Behavior changes worth noting beyond the diff:
- Device-flow token refresh is now gated to `UserAccount` profiles, which also fixes the
  `clusters watch` retry path for non-device-flow profiles.
- logout for the new auth type is wired in `PreRunE`, fixing a nil-func panic that
  previously aborted logout's cleanup.
- `NoAccessToken` works for opaque tokens (tokens for the new method are not parsed)

_Jira ticket:_ CLOUDP-418563

Depends on: mongodb/atlas-cli-core#70, mongodb/go-client-mongodb-atlas#562

## Checklist
- [x] I have signed the MongoDB CLA
- [x] I have added tests that prove my feature works
- [x] I have added any necessary documentation (docs regenerated for `--discover`)


saisundar:
@jricher Might it be possible for to you break this down into a stack of self-contained changes? The size of this diff might slow down (manual) reviews.

jricher:
@saisundar I'm honestly not sure what smaller set of changes would be meaningful here, as this is an entirely new authentication and api connection path being added, and that cascades throughout the code. What functions do you think would be separable?

internal/cli/auth/user_delegation.go
```diff
@@ -0,0 +1,197 @@
...
+	}
+
+	if metadata, ok := cached["metadata"].(map[string]any); ok {
+		return metadata, nil
+	}
+
+	return nil, errors.New("discovery response missing metadata")
+}
+
+func (opts *UserDelegationFlow) Run(ctx context.Context) error {
```
saisundar, line 90: This part, a good chunk of the PKCE and persistence piece, seems to be not be covered by tests. Could we please add some to gain confidence especially on some of the side effects?

internal/cli/auth/whoami.go
```diff
@@ -18,21 +18,52 @@ import (
...
+	}
 	_, _ = fmt.Fprintf(opts.OutWriter, "Logged in as %s %s\n", opts.authSubject, opts.authType)
+	return nil
+}
+
+func (opts *whoOpts) runUserDelegation() error {
+	msg := "Connected to MongoDB Atlas"
+
+	if opts.tokenExpiry != "" {
+		if t, err := time.Parse(time.RFC3339, opts.tokenExpiry); err == nil {
```
saisundar, line 48: Should these have the same else fallbacks as the empty token expiry?

internal/cli/auth/login.go
```diff
@@ -75,9 +79,10 @@ const (
 var (
 	ErrProjectIDNotFound = errors.New("project is inaccessible. You either don't have access to this project or the project doesn't exist")
 	ErrOrgIDNotFound     = errors.New("organization is inaccessible. You don't have access to this organization or the organization doesn't exist")
-	authTypeOptions      = []string{userAccountAuth, prompt.ServiceAccountAuth, prompt.APIKeysAuth}
+	authTypeOptions      = []string{userAccountAuth, prompt.UserDelegationAuth, prompt.ServiceAccountAuth, prompt.APIKeysAuth}
 	authTypeDescription  = map[string]string{
-		userAccountAuth:           "(best for getting started)",
+		userAccountAuth:           "(legacy account connection)",
```
saisundar: I realize this is what we want to call out when things have stabilized, and verified as working cleanly - does this wording change (legacy vs “best for user accounts”) need to ship as part of this release?

saisundar (commented):
Please add a note on the failing checks and why they are a no-op when fixed.

saisundar:
> @saisundar I'm honestly not sure what smaller set of changes would be meaningful here, as this is an entirely new authentication and api connection path being added, and that cascades throughout the code. What functions do you think would be separable?

a) the refresher_opts.go file can be independent.
b) user_delegation.go and its associated tests can be an independent PR. 
c) wiring + integration tests can be its own PR. 

Figuring out the DAG and then shipping the innermost independent PRs ( kinda like reverse toplogical sort) is usually the way to go. 

you can always do b) c) a) too - but this reverse topological sorting helps make reviewing faster.

internal/cli/auth/whoami.go
jricher: No, as they don't have the same auto-refresh capabilities wired in to the transport layer that this path does.

internal/cli/auth/login.go
jricher: This label needs feedback from the CLI team (as does the rest of the user-facing components, like the callback page and paste-helper page) before release.

jricher:
I disagree with the proposed split. The changes to `refresher_opts` are a meaningless no-op outside of the changes that the new code path brings. Additionally, I do not believe there is value in splitting tests from the code changes into separate PRs.

internal/cli/auth/login.go
saisundar: This major change would ideally be accompanied by a 
-  rollback strategy and a rollout strategy. Could you link one ( can't find it in the linked tickets or TD/Scope) ? 

There is the technical capability, and the product phrasing.
Phrasing this yet-to-be-validated-by-prod-traffic path as 

"	prompt.UserDelegationAuth: "(best for user accounts)",

 during its v0 commit is heavy-handed.

PR #4726, opened 2026-08-18
CLOUDP-435365: Bump ubi-minimal base image from 9.5 to 9.8

Luke-Sanderson:
## Proposed changes

_Jira ticket:_ CLOUDP-435365

Update Atlas CLI Docker image base from UBI 9.5 to UBI 9.8.

## Checklist

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,5 @@
 # syntax=docker/dockerfile:1.3-labs
-FROM registry.access.redhat.com/ubi9/ubi-minimal:9.5
+FROM registry.access.redhat.com/ubi9/ubi-minimal:9.8
 ENV MONGODB_ATLAS_IS_CONTAINERIZED=true
 
 COPY <<EOF /etc/yum.repos.d/mongodb-org-x86_64-8.0.repo
--- a/Dockerfile
+++ b/Dockerfile
@@ -20,12 +20,11 @@ enabled=1
 gpgkey=https://www.mongodb.org/static/pgp/server-8.0.asc
 EOF
 
-RUN microdnf -y install jq yum &&\
-    yum -y update &&\
-    yum install -y mongodb-atlas &&\
-    yum clean all &&\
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -1,5 +1,5 @@
 # syntax=docker/dockerfile:1.3-labs
```
cveticm, line 1: Scanning with Claude, it looks like we don't need the two remaining critical vuln libraries, curl-minimal and libcurl-minimal. I think we can remove them here too.
This will make yum non-fucntional after removal but it looks like yum is only needed for building the image and so it can be removed after as well

Luke-Sanderson: Hmm I believe that is probably true, however it looks suspiciously the same as this closed PR https://github.com/mongodb/mongodb-atlas-cli/pull/4545 so i'll DM Bianca and see if she remembers why it was closed

Dockerfile
```diff
@@ -1,5 +1,5 @@
...
-FROM registry.access.redhat.com/ubi9/ubi-minimal:9.5
+FROM registry.access.redhat.com/ubi9/ubi-minimal:9.8
```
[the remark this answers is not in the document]
Luke-Sanderson: /ar The image is intended to support `atlas local` commands, which require access to the mounted Docker socket. Running as a non-root user causes permission errors unless users add extra UID/GID flags, breaking the expected UX. The image is an ephemeral CLI utility rather than a long-running service, so root is intentional here to preserve Docker-based local deployment functionality.

Dockerfile
Luke-Sanderson: Done, I have made it uninstall yum and the curl dependencies. I was misunderstanding the task, the closed PR is a much more extensive change.

jeroenvervaeke (dismissed):

cveticm (approved):
lgtm!! 🚀

PR #4738, opened 2026-08-28
CLOUDP-439953: Stage unsigned release images on a host-local registry

fmenezes:
## Proposed changes

_Jira ticket:_ CLOUDP-439953

Unsigned release images were pushed to `docker.io/mongodb/apix_test`, a **public** repository, before being signed and promoted — so every daily release image was publicly pullable pre-signature and pre-verification, with nothing pruning the accumulated tags.

## Further comments

**Not yet validated by a real run.** The build-and-sign path can't be tested easily, I will test this via `workflow_dispatch` after merge.


Luke-Sanderson (approved):
LGTM

PR #4750, opened 2026-09-02
CLOUDP-432503: Migrate AI agent detection to the SDK helper and set the User-Agent on atlas api requests

fmenezes:
## Proposed changes

_Jira ticket:_ CLOUDP-432503

Two related changes for identifying AI agent traffic, plus a fix for `atlas api` requests being unattributable.

### 1. Telemetry migrated to the SDK helper

The hand-maintained `agentEnvVars` table and matching loop in `internal/telemetry/event.go` are replaced by `detectaiagent.Detect()` from the SDK (CLOUDP-432500), removing ~40 lines. 

### 2. User-Agent on `atlas api` requests

`atlas api` builds its own client and never set a User-Agent. The clients from `atlas-cli-core` only inject one on the service-account and OAuth-token paths, so requests authenticated with **API keys, a user account, or no auth were sending `Go-http-client/1.1`** — not attributable to the Atlas CLI at all, let alone to an agent.

## Checklist

## Further comments

https://github.com/mongodb/atlas-cli-core/pull/76 is required


Luke-Sanderson (dismissed):
LGTM

internal/telemetry/event_test.go
```diff
@@ -404,66 +404,13 @@ func TestWithHelpCommand_NotFound(t *testing.T) {
...
-	for _, a := range agentEnvVars {
-		// t.Setenv registers restoration of the original value; os.Unsetenv
-		// then actually clears it so presence-based checks see it as unset.
-		t.Setenv(a.envVar, "")
-		require.NoError(t, os.Unsetenv(a.envVar))
-	}
-}
-
 func TestWithAgent(t *testing.T) {
-	tests := []struct {
```
blva, line 418: nit: could you please keep the unit tests so that we can validate we're not adding any regressions to the fields?

blva (changes requested):

internal/store/store.go
```diff
@@ -137,7 +138,7 @@ func WithContext(ctx context.Context) Option {
 
 // setAtlasClient sets the internal client to use an Atlas client and methods.
 func (s *Store) setAtlasClient() error {
-	opts := []atlas.ClientOpt{atlas.SetUserAgent(config.UserAgent(version.Version))}
+	opts := []atlas.ClientOpt{atlas.SetUserAgent(useragent.UserAgent(version.Version))}
```
blva: is this covering api commends as well?

internal/telemetry/event_test.go
fmenezes: will SDK be source of truth of cli?

internal/store/store.go
fmenezes: I'm reworking this given https://github.com/mongodb/atlas-cli-core/pull/76

internal/telemetry/event_test.go
blva: It will, but we should know if there is a regression added in the fields, right?

fmenezes: ideally the regression should be caught in the SDK, but I can revert the test

internal/telemetry/event.go
```diff
@@ -151,47 +152,13 @@ func withCI() EventOpt {
...
-				continue
-			}
-			if !ok || v == "" {
-				continue
-			}
-			if a.value != "" && v != a.value {
-				continue
-			}
-			event.Properties["agent_env_var"] = a.name
+		agent, ok := detectaiagent.Detect()
```
manupedrozo, line 157: Q: I don't see the `agent.UserAgentIdentifier` being appended the UserAgent. 
My understanding is that the CLI was only reporting AI agent usage in analytics so maybe the UserAgent update is expected to happen elsewhere, how is the UserAgent wired in the CLI?

manupedrozo: Nvm, just saw: https://github.com/mongodb/atlas-cli-core/pull/76

Does this mean that `detectaiagent.Detect()` is potentially called twice within a single CLI operation?

internal/telemetry/event_test.go
blva: fair point, since we're here now we can at least try to figure out if we will need to backfill anything on the analytics side for consistency

internal/telemetry/event.go
fmenezes: potentially, one for telemetry and another to setup the http client for the command, any concerns?

manupedrozo: @blva called this out during design as something to keep an eye on.

Detection is not too expensive so no big concern but good to keep in mind since it runs for every CLI command. Its easier to amortize in tools like Terraform where we execute multiple operations in a single execution.

fmenezes: It is possible to write code to prevent it, but sdk only checks env vars and checks if 1 file exists, it seemed excessive to counter for it

internal/store/store.go
blva: thanks!

internal/telemetry/event.go
blva: my concern has been around the risk of having commands taking longer than expected over time, perharps that is something we could capture as a follow up? what would you recommend @fmenezes ?

fmenezes: reading env vars does not take long, same for fs.Stat, commands are taking long because of network requests, in the case of CLI we do at least 2 requests per command their main API + telemetry.

fmenezes: if anything we should revise telemetry and where possible switch our backend to async APIs

blva: cool, that sounds reasonable, do you want to cut a ticket for us to triage this approach in the future?

blva (approved):

PR #4760, opened 2026-09-08
CLOUDP-432111: Add hidden parameter "includeGlobal" to Atlas CLI 

aniket2907:
## Proposed changes

Users with a global Atlas role currently see every organization in the deployment when running atlas auth login, atlas auth register, or atlas setup, which makes the org picker unusable.

MMS supports a hidden includeGlobal query parameter on `listOrgs`. The default is `true`; setting it to `false` returns only organizations the caller is a member of.

Because the parameter is marked `@Parameter(hidden = true)`, it isn't in the OpenAPI spec or `atlas-sdk-go`, so it can't be sent through the typed SDK path.

This adds internal/api/orgsapi, which copies the generated listOrgs command, adds the hidden parameter, and executes it through L1. The parameter reaches the wire without appearing in the spec, --help, or docs/.

`auth login`, `auth register`, and `setup` always send `includeGlobal=false`

`organizations list` is unchanged by default, with hidden `--includeGlobal=false`available to opt out

When `includeGlobal=true`, the parameter is omitted, so default requests remain byte-identical to before. There is no customer impact; the parameter only controls the global-role check.

_Jira ticket:_ CLOUDP-432111

## Checklist

## Further comments

This is the first L3 command to call the L1 executor directly (other non-test callers: internal/cli/api/api.go:187,208 and internal/api/watcher.go:170).


internal/api/orgsapi/orgsapi.go
```diff
@@ -0,0 +1,192 @@
...
+
+	return command
+}
+
+// listOrgsCommand returns a copy of the generated listOrgs command with the includeGlobal
+// parameter appended.
+func listOrgsCommand() (shared_api.Command, error) {
+	command, err := findCommand(listOrgsOperationID)
+	if err != nil {
+		return shared_api.Command{}, err
```
cveticm, line 100: Is there any use to returning `shared_api.Command{}` here?  Following the flow to the end when err!=nil, we dismiss the `shared_api.Command{}` result [here](https://github.com/mongodb/mongodb-atlas-cli/pull/4760/changes#diff-35ecff1953e6f3b53381475a6e6cf6ef0c8c988dbdb96cd065a29308f4eed16dR166).

internal/api/orgsapi/orgsapi.go
```diff
@@ -0,0 +1,192 @@
...
+
+// withIncludeGlobal returns a copy of a command with the includeGlobal parameter appended.
+// Copying a Command only copies the slice header, so the parameters are cloned before
+// appending: without that, appending to a slice with spare capacity would write into the
+// backing array of api.Commands, which is global state backing the cobra tree.
+func withIncludeGlobal(command shared_api.Command) shared_api.Command {
+	queryParameters := slices.Clone(command.RequestParameters.QueryParameters)
+	queryParameters = append(queryParameters, shared_api.Parameter{
+		Name:        includeGlobalParam,
+		Description: "Flag that indicates whether organizations visible through a global role are returned.",
```
cveticm, line 87: nit: could we use the new `usage.OrgIncludeGlobal` and `flag.IncludeGlobal`?

internal/api/orgsapi/orgsapi.go
aniket2907: Hey, you're right that it's essentially a discarded/redudant return, but it's intentional (Go requires a value on every return path for a (T, error) function). Since Command is a struct, not a pointer, nil isn't valid; so shared_api.Command{} is the zero value and the minimal thing that compiles.

The returned command is intentionally discarded since callers check err first. The same pattern can be found in tools/cmd/api-generator/convert_commands.go:339-374 and internal/cli/deployments/options/deployment_opts_pre_run.go:75-90.

internal/api/orgsapi/orgsapi.go
aniket2907: Good catch. I was thinking - since this description field is never read (read only when registering cobra flags for the 'atlas api ..' commands, and since this is a copy of the command sent through an L1 executor, it never gets there and is thus essentially a dead string), it could be fully removed without consequenc

jeroenvervaeke (approved):

